PR #19, opened 2019-03-26
Support for go modules

### LINKED ISSUE #7: ko does not work with Go modules

paivagustavo:
I took some inspiration based on the @jonjohnsonjr PoC google/go-containerregistry/pull/329

And take it futher using the new https://godoc.org/golang.org/x/tools/go/packages package to work with both modules and normal go projects.

I'm not yet satisfied how the codes look but from my first tests it seems to being working. Will try to implement tests to this, don't know how to actually test this.

Fixes #7


paivagustavo:
@jonjohnsonjr could you please do an initial review?

paivagustavo:
I may have broken the `kodata` test. Guess I still have some work to do.  Having some trouble to discover the absolute path for a given package.

pkg/build/gobuild.go
```diff
@@ -270,18 +288,39 @@ func tarKoData(importpath string) (*bytes.Buffer, error) {
 	return buf, nil
 }
 
+// getAbsolutePath returns the absolute path which is calculated by joining the current directory and
+// the relative path for this importPath to the base package.
+//
+// This is necessary because when the project is a module project since "$GOPATH/importPath" is not the
+// actual absolute path of the file.
+func (g *gobuild) getAbsolutePath(importPath string) (string, error) {
+	wd, err := os.Getwd()
```
jonjohnsonjr: This seems like it would break any cases where you're building something not rooted in your current directory. For go modules, I guess that's unavoidable, but I'd like to avoid breaking anyone not using go modules yet.

paivagustavo: Yes! I ended up using the `go list` command to find out the absolute path, that worked just fine.

paivagustavo:
@jonjohnsonjr I've cleaned up the code and it's working for both modules and non modules projects. It still missing some test, I can't figure out how to do it.

Maybe create another "ko test" command but with modules and force the `export GO111MODULE=on` ? I still don't think that will actually test correctly this. 

Maybe even create a go module project on the `/tmp`?

I would love some opinions on this.

jonjohnsonjr:
> I ended up using the go list command to find out the absolute path, that worked just fine.

I think by default `packages.Load` dispatches to `go list` under the hood. Is there a way to call `packages.Load` to do what you're doing by shelling out to `go list`? Not a huge deal, would just be a bit nicer.

> It still missing some test, I can't figure out how to do it.

Which test and what's the issue?

paivagustavo:
> I think by default packages.Load dispatches to go list under the hood. Is there a way to call packages.Load to do what you're doing by shelling out to go list? Not a huge deal, would just be a bit nicer.



I guess `packages.Load` could do it, but is it really necessary?

There is two points I can see for this:
- `packages.Load` maybe an unnecessary abstract on top of `go list` for this use case.
- on the other hand, it might be safer to use it since any breaking changes to `go list` is going to handled by `packages.Load`

I actually prefer to use `go list` direct because I feel that it's simpler but if you want to go on the other path I'm totally down with it too.


> Which test and what's the issue?

I feel that current tests doesn't assure that ko actually works with modules.

jonjohnsonjr:
> I actually prefer to use go list direct because I feel that it's simpler

That seems fine since we already shell out to `go build`. Maybe just add a TODO to use `packages` instead if we find that using `go list` directly isn't reliable.

> I feel that current tests doesn't assure that ko actually works with modules.

Ah, got it. Creating a module inside a `ioutil.TempDir` seems reasonable to me.

paivagustavo:
@jonjohnsonjr I've added some tests for it. I had to add an extra abstraction to set the working directory of the build process, so I can fake that I'm running ko from the go module project that is on the `ioutil.Tempdir`.

I have duplicated some test code just to validate the idea. If you feel like this is enough i'll refactor to be easier to add more test later.

pkg/build/gobuild.go
```diff
@@ -82,26 +85,54 @@ func NewGo(options ...Option) (Interface, error) {
 	return gbo.Open()
 }
 
+// goList is a wrapper for the `go list` command to retrieve information about a package.
+func (g *gobuild) goList(importPath, projection string) (string, error) {
```
jonjohnsonjr: What does `projection` here mean? Is that a known term for `go list`?

pkg/build/gobuild.go
```diff
@@ -82,26 +85,54 @@ func NewGo(options ...Option) (Interface, error) {
...
+
+	stdout, err := cmd.Output()
+
+	if err != nil {
+		return "", err
+	}
+
+	outputs := strings.Split(string(stdout), "\n")
+
+	if len(outputs) == 0 {
```
jonjohnsonjr: I think I'd pull this error handling/checking outside of `goList` and make `goList` return `([]string, error)` or rename `goList` to sound less generic.

pkg/build/gobuild.go
jonjohnsonjr: nit: there's a lot of extra newlines in here, I'd drop most of them (I usually like to leave a newline after an error handling clause as a kind of visual "checkpoint")

pkg/build/gobuild.go
```diff
@@ -82,26 +85,54 @@ func NewGo(options ...Option) (Interface, error) {
...
+		return "", errors.New("could not find specified import path: " + importPath)
+	}
+
+	output := outputs[0]
```
jonjohnsonjr: Will this ever return more than 0 or 1 results?

pkg/build/gobuild_test.go
```diff
@@ -26,39 +27,110 @@ import (
...
-	base, err := random.Image(1024, 3)
+func createModuleProject(t *testing.T) string {
+	tmpDir, err := ioutil.TempDir("", "module-test")
 	if err != nil {
-		t.Fatalf("random.Image() = %v", err)
+		t.Error(err.Error())
 	}
+	os.Mkdir(filepath.Join(tmpDir, "kodata"), 0777)
+
+	writeFile(t, filepath.Join(tmpDir, "main.go"), `
```
jonjohnsonjr: Is it possible to pull these out into some checked-in files, and just copy them to a tmpdir? It's kind of ugly with them inlined here. Is this just replicating [these files](https://github.com/google/ko/tree/master/cmd/ko/test)?  You could just copy that directory to `tmpDir` and then write the `go.mod` file (and maybe fix up symlinks? not sure).

pkg/build/options.go
```diff
@@ -44,3 +44,12 @@ func withBuilder(b builder) Option {
 		return nil
 	}
 }
+
+// withWd is a functional option for overriding the working directory
+// where ko is going run. This is exposed for testing.
+func withWd(wd string) Option {
```
jonjohnsonjr: Can you avoid adding this by just doing [`os.Chdir`](https://golang.org/pkg/os/#Chdir) in the test? (You probably would want to defer a `Chdir` back to the result of `Getwd` to avoid messing up other tests).

jonjohnsonjr:
@paivagustavo any chance of picking this back up?

paivagustavo:
Hi @jonjohnsonjr I totally forgot about this. I'll try to finish finish this week :)

pkg/build/gobuild.go
paivagustavo: changed to format, since it is the name of the flag on go list.

pkg/build/gobuild.go
paivagustavo: changed the method name to findImportPath to be more specific.

pkg/build/gobuild.go
paivagustavo: Thanks, this was one of the first codes in go I have written and had a lot of bad habit from other languages :)

paivagustavo:
Just replying here to say that I'll finish this tomorrow :)

paivagustavo:
I have done a wrong push :sob:  and so this PR was automatically closed. Please redirect to #50

PR #24, opened 2019-04-09
 Add --disable-optimizations flag to make debugging easier.

Harwayne:
Add --disable-optimizations flag to make debugging easier. 

---

I used this, along with a different base image (`ubuntu:bionic-20190307`), so that I could interactively debug a `ko` built and applied container running in my Kubernetes cluster, using `squash` and VS Code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# Ignore GoLand (IntelliJ) files.
+.idea/
+
--- a/cmd/ko/commands.go
+++ b/cmd/ko/commands.go
@@ -70,6 +70,7 @@ func addKubeCommands(topLevel *cobra.Command) {
 	no := &NameOptions{}
 	fo := &FilenameOptions{}
 	ta := &TagsOptions{}
+	do := &DebugOptions{}
 	apply := &cobra.Command{
 		Use:   "apply -f FILENAME",
 		Short: "Apply the input files with image references resolved to built/pushed image digests.",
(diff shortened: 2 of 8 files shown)


dprotaso:
Out of curiosity do you just ssh to the container and use gdb?

jonjohnsonjr:
I believe the strategy is to override the `ko` base image with something containing [delve](https://github.com/go-delve/delve) and then use [squash](https://github.com/solo-io/squash) to debug. Adam landed [some changes](https://github.com/solo-io/squash/commits?author=Harwayne) to enable debugging from vscode using the squash plugin.

Harwayne:
As @jonjohnsonjr mentioned, I am using `squash`, in particular its VS Code plugin. I tried to document the steps I needed to take to get everything working [here](https://github.com/Harwayne/squash/blob/f5c12109a300387d30555db3d8caa96d502471e3/editor/vscode/README.md#how-to-run-it), but they were notes meant for me more than anything else. And since then I've merged PRs into `squash`, so the next release won't require building it myself.

dprotaso:
Cool thanks for the context

.gitignore
```diff
@@ -0,0 +1,3 @@
+# Ignore GoLand (IntelliJ) files.
```
matzew: :heart:

matzew:
I'd like to see this for better debugging options. Thanks @Harwayne

Issue #21, opened 2019-03-27
`ko version` to report build/version information


ImJasonH:
Using https://github.com/ahmetb/govvv

cezkuj:
I would like to take care of it. Could you advise on activities needed besides adding 'version' command? govvv changes build process (as you don't use go build directly anymore) and I don't see any build wrapper (Makefile/scripts etc.) to hide that. Same problem with using package, users 'go getting' ko wouldn't have it working.

ImJasonH:
Thanks for your interest @cezkuj !

We don't currently have an official release process or any release scripts, which is probably a prerequisite for this.

For users can install via `go get` we can have `go version` return the commit hash that's checked out (by inspecting `$GOPATH/src/github.com/google/ko/.git/refs/heads/master` (or else whatever is `HEAD`) at runtime, if the const value isn't provided.

FWIW, I also found [an alternative approach](https://stackoverflow.com/questions/11354518/golang-application-auto-build-versioning) to `govvv` which is to use `-ldflags` to set the value of a package-level const at build-time.

cezkuj:
As far as I see, govvv also uses -ldflags approach. I feel like that's most common way to get build info.  Not sure if it is worth to have hard dependency on other project to reduce few less lines here and there - your call.
Shall I add Makefile/bash/bazel/anything else to wrap building and releasing process within that issue?
Or just add code for version command?

ImJasonH:
Right you are. I don't have a strong opinion about taking a dependency on `govvv`, it seems pretty straightforward and if we ever needed to stop using it we could.

I think if you add a release script that populates consts which `go version` prints that'd be sufficient. We'll tag a release after that and document how to get an official release where `go version` prints something other than the commit hash.

cezkuj:
Overall I agree, but one thing is a bit unclear to me, what do you mean by 
> go version prints something other than the commit hash.
I don't see any way to get it from go version, do you mean git rev-parse (or some equivalents)?

ImJasonH:
Yeah, using `git rev-parse` would work, basically:

```
 // provided by govvv
var Version, BuildDate string

func goVersion() {
  if Version == "" {
    hash := gitRevParseHead() // output of `git rev-parse HEAD`
    fmt.Println("version (go get):, hash)
  } else {
    fmt.Println("version:", Version)
  }
}
```

[...]

PR #25, opened 2019-04-16
Resolving #21 issue. Adding ko version and simple release script

cezkuj:
Output from version command implemented as we have discussed in comments, easily we can extend it with other build information from govvv, if needed.
Also, added basic release script, I didn't have any guidelines so I added what seemed valid for me, let's discuss if it should be improved.
Cheers!


hack/release.sh
```diff
@@ -0,0 +1,28 @@
...
+# Usage:
+#   hack/release.sh $VERSION
+#
+# Example:
+#   hack/release.sh v0.2
+
+VERSION=$1
+
+KO_ROOT="$(cd "$(dirname "$0")" && pwd)/.."
+govvv build -o $KO_ROOT/build/ko $KO_ROOT/cmd/ko -version $VERSION
```
jonjohnsonjr: Perhaps add a `go get github.com/ahmetb/govvv` before this?

cezkuj: Good point. Added.

cezkuj:
@jonjohnsonjr @ImJasonH 

I've updated branch so it doesn't have conflicts anymore. Could you review it?

ImJasonH:
I think we want to run `git rev-parse` in `ko`'s directory in `GOPATH`, not in the current working directory, which is probably another repo and not ko's.

cezkuj:
Hmm, agreed but, using GOPATH wouldn't be possible if migrated to go modules. It is definitely drawback - not sure if valid for now.

For the time being we could use git dir based on GOPATH. Smth like that:
```go
gitDir := fmt.Sprintf("--git-dir=%v/src/github.com/google/ko/.git", os.Getenv("GOPATH"))
hash, err := exec.Command("git", gitDir, "rev-parse", "HEAD").Output()
```

PR #30, opened 2019-04-26
Adding ko create

n3wscott:
I wanted `ko create -f config/file.yaml` 
but I had to `ko publish -f config/file.yaml | kubectl create -f -`

Fixed that by just adding `create` to `ko`.


n3wscott:
@jonjohnsonjr

PR #33, opened 2019-04-30
Return better error messages

jonjohnsonjr:
This hoists the publisher creation out of the loop for `ko publish` to
avoid needlessly resolving credentials multiple times.

This pulls the publisher creation outside of resolveFilesToWriter so we
can fail earlier if e.g. KO_DOCKER_REPO is unset, otherwise we start up
the kubectl goroutine and it would asynchronously fail with:

  error: no objects passed to apply

### DIFF
--- a/pkg/commands/apply.go
+++ b/pkg/commands/apply.go
@@ -15,14 +15,14 @@
 package commands
 
 import (
-	"github.com/google/ko/pkg/commands/options"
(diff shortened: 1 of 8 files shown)


jonjohnsonjr:
I'm thinking we could pull `makePublisher` and `makeBuilder` out into some initialization step at some point since they're dependency-injectiony and just parse options into these behaviors.

Issue #6, opened 2019-01-02
ko should produce multi-platform images


jonjohnsonjr:
Right now we assume amd64/linux and use distroless.

This is blocked on full manifest list support in the library, but ideally supplying a multi-platform base image would produce a multiplatform ko image.

(this is mostly a placeholder for reference until I flesh this out more)

jonjohnsonjr:
The general idea for this was to drive the platform selection via the base image. Specifying a manifest list as the base image should produce a manifest list. If you want to build for a single platform, you could specify that platform's image.

I like this approach because we don't have to add any additional configuration or flags or anything, it's totally data-driven. One drawback is that it might be surprising that if you specify e.g. `busybox`, you'd end up building a binary for every single platform, but we could print a warning or something to make that clear.

We currently just pass through environment variables to `go build` for GOOS and GOARCH to set the platform/os, but we could inspect the base image's configuration instead for [`architecture`](https://github.com/google/go-containerregistry/blob/1c6c7f61e8a5402b606c3c6db169fdcd1b0712b7/pkg/v1/config.go#L27) and [`os`](https://github.com/google/go-containerregistry/blob/1c6c7f61e8a5402b606c3c6db169fdcd1b0712b7/pkg/v1/config.go#L33).

This will be a bit of a pain to rework some of `ko`'s build/publish interfaces, but it shouldn't be too difficult.

zxDiscovery:
May I ask I question? Am I must export the env `GOARCH=ppc64le` if I want to build the ppc64le image on power when I use the current `ko`? Thanks.
@jonjohnsonjr

jonjohnsonjr:
@zxDiscovery correct,  if you use a ppc64le specific image, you don''t need to export GOARCH anymore. Let me know if that doesn't work for you.

zxDiscovery:
I use ko to build the ppc64le image on ppc64le platform, but the image can't work normally when I didn't export `GOARCH=ppc64le`. 
```
root@dusk1:~# docker run -it 666ffcf0b33c
standard_init_linux.go:190: exec user process caused "exec format error"
```
After I export `GOARCH=ppc64le`, the images can work normally.
My environment information as follows:
```
root@dusk1:~# uname -o
GNU/Linux
root@dusk1:~# uname -a
Linux dusk1.fyre.ibm.com 4.15.0-45-generic #48-Ubuntu SMP Tue Jan 29 16:27:02 UTC 2019 ppc64le ppc64le ppc64le GNU/Linux
[...]

PR #38, opened 2019-06-17
Multi-platform ko

jonjohnsonjr:
This is a sketch of what I had in mind for https://github.com/google/ko/issues/6

For this PR, I've set the default base image for this repo to `ubuntu` so that you can just fetch this branch and run `ko publish ./cmd/ko` to test out the change. It should build a different binary for every platform and publish a manifest list instead of just a manifest.

```
$ ko publish ./cmd/ko
2019/06/17 14:04:04 Using base index.docker.io/library/ubuntu:latest for github.com/google/ko/cmd/ko
2019/06/17 14:04:04 No matching credentials were found, falling back on anonymous
2019/06/17 14:04:06 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:12 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:15 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:19 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:23 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:27 Building github.com/google/ko/cmd/ko
2019/06/17 14:04:30 Publishing gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0:latest
2019/06/17 14:04:32 existing manifest: sha256:e66f9706d017b28a5f63fd2b1a9395a43852bb8691c2a3d789bf6187ed2a83ef
2019/06/17 14:04:32 existing manifest: sha256:6ae8959c22069d43bbaa848496237a653f4644244f638d11aa05dee555e2d7d9
2019/06/17 14:04:32 existing manifest: sha256:d835632ac512ee32b2489d44c15ecb03a139785b2e3d475a07670288fc95242e
2019/06/17 14:04:32 existing manifest: sha256:ce8f3571502ac3b5c86f1d527c712abd4cad953c97f24ba9b28a6505e502f1ad
2019/06/17 14:04:32 existing manifest: sha256:5c60f74c26ec1726f06e6c35ca1a3aee0c567d37001143ec70e9cc6f5b1a763b
2019/06/17 14:04:33 existing manifest: sha256:c1940d8d12f448cf17212cf76e012d8b8480e0b033fac5ff6068402a8f0f590f
2019/06/17 14:04:34 gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0:latest: digest: sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621 size: 1411
2019/06/17 14:04:34 Published gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621
gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621

$ crane manifest gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621 | jq .
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
  "manifests": [
    {
      "size": 1073,
      "digest": "sha256:e66f9706d017b28a5f63fd2b1a9395a43852bb8691c2a3d789bf6187ed2a83ef",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "size": 1073,
      "digest": "sha256:6ae8959c22069d43bbaa848496237a653f4644244f638d11aa05dee555e2d7d9",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v7"
      }
    },
    {
      "size": 1073,
      "digest": "sha256:d835632ac512ee32b2489d44c15ecb03a139785b2e3d475a07670288fc95242e",
      "platform": {
        "architecture": "arm64",
        "os": "linux",
        "variant": "v8"
      }
    },
    {
      "size": 1073,
      "digest": "sha256:ce8f3571502ac3b5c86f1d527c712abd4cad953c97f24ba9b28a6505e502f1ad",
      "platform": {
        "architecture": "386",
        "os": "linux"
      }
    },
    {
      "size": 1073,
      "digest": "sha256:5c60f74c26ec1726f06e6c35ca1a3aee0c567d37001143ec70e9cc6f5b1a763b",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      }
    },
    {
      "size": 1073,
      "digest": "sha256:c1940d8d12f448cf17212cf76e012d8b8480e0b033fac5ff6068402a8f0f590f",
      "platform": {
        "architecture": "s390x",
        "os": "linux"
      }
    }
  ]
}
```

I haven't actually tested pulling/running this on different platforms (I don't have any), so YMMV, but publishing works 👍 

I don't actually like this very much because I've made every package deal with a new abstraction (I've called it `Steve` so that we don't ever merge this) that can represent either base image or a base manifest list. A better approach would probably be to make everything deal with a manifest list to simplify things.

When pulling base image:
- if it's a manifest list, pass that along as the base
- if it's an image, wrap it as a singleton manifest list

When publishing the artifacts:
- if it's a singleton manifest list, unwrap it and publish just the image
- otherwise, publish it as a manifest list

This isn't perfect (what if the user actually does want to publish a singleton manifest list?), so maybe it's not the best idea. Open to suggestions :)


jonjohnsonjr:
https://github.com/google/go-containerregistry/issues/474 would make my proposed implementation easier.

pkg/build/gobuild.go
```diff
@@ -355,3 +362,130 @@ func (gb *gobuild) Build(s string) (v1.Image, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return steve.Image(img)
+	}
+
+	return nil, fmt.Errorf("oops")
+}
+
+type index struct {
```
jonjohnsonjr: All of this could go away with https://github.com/google/go-containerregistry/issues/474

vdemeester:
> I don't actually like this very much because I've made every package deal with a new abstraction (I've called it `Steve` so that we don't ever merge this) that can represent either base image or a base manifest list. A better approach would probably be to make everything deal with a manifest list to simplify things.
> 
> When pulling base image:
> 
>     * if it's a manifest list, pass that along as the base
> 
>     * if it's an image, wrap it as a singleton manifest list
> 
> 
> When publishing the artifacts:
> 
>     * if it's a singleton manifest list, unwrap it and publish just the image
> 
>     * otherwise, publish it as a manifest list
> 
> 
> This isn't perfect (what if the user actually does want to publish a singleton manifest list?), so maybe it's not the best idea. Open to suggestions :)

Could there be a flag or environment variable to customize this behavior ? (like it does what you describe by default but the user could force to build a singleton manifest list)

jonjohnsonjr:
> Could there be a flag or environment variable to customize this behavior ? (like it does what you describe by default but the user could force to build a singleton manifest list)

Hm.. that gives me an idea. We could add a `--platform` flag to select which platform to use from a manifest list, and have some sentinel value (`all`?) to indicate we publish a manifest list. This would make it possible to publish a singleton manifest list, but also allow targeting a specific platform.

vdemeester:
> Hm.. that gives me an idea. We could add a `--platform` flag to select which platform to use from a manifest list, and have some sentinel value (`all`?) to indicate we publish a manifest list. This would make it possible to publish a singleton manifest list, but also allow targeting a specific platform.

I like that idea :heart_eyes_cat: 

> I don't actually like this very much because I've made every package deal with a new abstraction (I've called it Steve so that we don't ever merge this) that can represent either base image or a base manifest list. A better approach would probably be to make everything deal with a manifest list to simplify things.

on the `Steve` package (:joy_cat:), I would also think a better approach would be to make everything deal with manifest list :angel:

vincent-pli:
@jonjohnsonjr 
I think when user run `ko build`, in most cases he does not expect to `all platform`,  even he pick up a multi-arch image as base image.
Could we use the `GOARCH` to tell `ko` which arch he expect to. based on that, `ko` could check if the base image support that arch and raise error if not support.

If `GPARCH` not set, `ko` could release a `all-platform` image.

jonjohnsonjr:
> I think when user run ko build, in most cases he does not expect to all platform, even he pick up a multi-arch image as base image.

Agreed, my first thought was to just default to amd64/linux (since MacOS -> Linux Kubernetes is so common), and it's the [default in docker distribution](https://docs.docker.com/registry/spec/manifest-v2-2/#backward-compatibility).

Using GOARCH/GOOS also makes sense.

Another idea I just had was to try to derive the os/arch based on the cluster we're deploying to:
```
$ kubectl get nodes -ojsonpath="{.items[0].status.nodeInfo.architecture}/{.items[0].status.nodeInfo.operatingSystem}"
amd64/linux
```

Though, that would add a little bit of latency to everything...

Some combination of those things is probably going to be the best user experience, e.g.:

1. If the base image is a single image, use that image's platform.
2. If the `--platform` flag is set, use that (`all` == build every platform).
3. If `GOOS` and `GOARCH` environment variables are set, use them.
4. If we can determine the platform via `kubectl get nodes`, use that.
5. Else, use amd64/linux.

vincent-pli:
That's great, maybe option 2 and option 3 are little overlapped, but whatever, expect to see the new implements

lysannef:
Hi @jonjohnsonjr , @junawaneshivani and I have been working on testing the changes in this PR on ppc64le platform.
However we have been facing an issue with the go build of this [repo](https://github.com/jonjohnsonjr/ko/tree/multi-platform-wip)
We have tried the build in a Kubernetes environment as well.

**Error encountered:**
 ```
# github.com/jonjohnsonjr/ko/cmd/ko
cmd/ko/main.go:33:26: cannot use cmds (type *"github.com/jonjohnsonjr/ko/vendor/github.com/spf13/cobra".Command) as type *"github.com/google/ko/vendor/github.com/spf13/cobra".Command in argument to commands.AddKubeCommands
```
Are we missing any step?

**Comment by vincent-pli** [comment]
@lysannef 
You should change the `path` of project to `github.com/google/ko`

**Comment by lysannef** [comment]
@vincent-pli These are the steps I took:
```
git clone https://github.com/google/ko.git
cd ko
git fetch origin +refs/pull/38/merge   ( so as to reflect the changes made by @jonjohnsonjr )
git checkout FETCH_HEAD
cd cmd/ko
go build
```
trying a go build in this repository, still throws the same error.

> @lysannef
You should change the path of project to github.com/google/ko

Could you please elaborate on the same. We are trying to test the changes made in @jonjohnsonjr 's repository at the multi-platform-wip branch. How would changing to the original ko repo work?

**Comment by vincent-pli** [comment]
So what's the path of your `ko` project?

**Comment by lysannef** [comment]
> So what's the path of your ko project?

@vincent-pli The path we are testing:  github.com/jonjohnsonjr/ko/cmd/ko
Changing the path to the google/ko repository with path : github.com/ko/cmd/ko and tried building it which resulted in the same error.


Using the path suggested by you : github.com/google/ko 
the build proceeds however none of jonjohnsonjr's changes are reflecting and the log is as follows
```
2019/07/08 00:47:27 Building github.com/google/ko/cmd/ko
2019/07/08 00:47:35 Using base index.docker.io/library/ubuntu:latest for github.com/google/ko/cmd/ko
2019/07/08 00:47:35 No matching credentials were found, falling back on anonymous
2019/07/08 00:47:39 Loading ko.local/ko-099ba5bcefdead87f92606265fb99ac0:333f19bd74774f8eccc0277a79375c80c6f810abb7dfda2496bdaabc089af622
2019/07/08 00:47:47 Loaded ko.local/ko-099ba5bcefdead87f92606265fb99ac0:333f19bd74774f8eccc0277a79375c80c6f810abb7dfda2496bdaabc089af622
2019/07/08 00:47:47 Adding tag latest
2019/07/08 00:47:48 Added tag latest
ko.local/ko-099ba5bcefdead87f92606265fb99ac0:333f19bd74774f8eccc0277a79375c80c6f810abb7dfda2496bdaabc089af622

```

**Comment by vincent-pli** [comment]
I'm little confuse,  so I just try it on ppc64le
1. Git clone git@github.com:jonjohnsonjr/ko.git to `GOPATH` as path: `/root/go/src/github.com/google/ko`
2. Run `go build` in `root/go/src/github.com/google/ko/cmd/ko`, then get binary `ko`, mv it to `/root/go/bin`
3. run `ko publish ./cmd/ko` in `root/go/src/github.com/google/ko`

It works, and get correct result.
BTW, are you a ibmer?

**Comment by lysannef** [comment]
> Git clone git@github.com:jonjohnsonjr/ko.git to GOPATH as path: /root/go/src/github.com/google/ko
Run go build in root/go/src/github.com/google/ko/cmd/ko, then get binary ko, mv it to /root/go/bin
run ko publish ./cmd/ko in root/go/src/github.com/google/ko

I followed these steps and this is the result I get:
```
$ ko publish ./cmd/ko
2019/07/08 02:42:23 Building github.com/google/ko/cmd/ko
2019/07/08 02:42:29 Using base gcr.io/distroless/static:latest for github.com/google/ko/cmd/ko
2019/07/08 02:42:29 No matching credentials were found, falling back on anonymous
2019/07/08 02:42:35 Loading ko.local/ko-099ba5bcefdead87f92606265fb99ac0:9d2e16d79eb082396f5374b765ee5b6c62f2485d13bac726fe6335aad39f2a11
2019/07/08 02:42:40 Loaded ko.local/ko-099ba5bcefdead87f92606265fb99ac0:9d2e16d79eb082396f5374b765ee5b6c62f2485d13bac726fe6335aad39f2a11
2019/07/08 02:42:40 Adding tag latest
2019/07/08 02:42:40 Added tag latest
```


Based on the ouput given [above](https://github.com/google/ko/pull/38#issue-289020815) I dont see the additional images being created , like 
```
2019/06/17 14:04:34 gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0:latest: digest: sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621 size: 1411
2019/06/17 14:04:34 Published gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621
gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:96ac9a314510e380cf3dd3aae3abc51dad577eecf53ebf6691f0c4d665d90621
```

> BTW, are you a ibmer?
Yes I am an IBMer

**Comment by vincent-pli** [comment]
@lysannef you can find me by ST: Peng QL Li 😄

**Comment by lysannef** [comment]
@jonjohnsonjr We have validated the working of these changes on ppc64le architecture. ppc64le architecture images are generated , and the manifest is as below.

```
 docker manifest inspect localhost:5000/knative/ko-099ba5bcefdead87f92606265fb99ac0@sha256:9fe6837835e5fe4f19bd037f2d85bc350a32a9904e088fdc74d6c46fb627b56a --insecure
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "size": 1235,
         "digest": "sha256:15271f24993d960daeea17c856be3f5cc89acdae902e6a5f6634c9c1d94c4ff9",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "size": 1235,
         "digest": "sha256:5092ff253eb247d1dfaba556f575bcbd69f2371505a2e459b6593055841f8f0a",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "size": 1235,
         "digest": "sha256:fd7eb6a9c0ab9494e49204bd51090939d23d754e32e9f1f1810e482646372353",
         "platform": {
            "architecture": "arm64",
            "os": "linux",
            "variant": "v8"
         }
      },
      {
         "size": 1235,
         "digest": "sha256:8b5f22a0386daecb29fc2a1705eba24800adf788a06cb953d2e698a23eba9e50",
         "platform": {
            "architecture": "386",
            "os": "linux"
         }
      },
      {
         "size": 1235,
         "digest": "sha256:2de4326fd4c4c15b2e08ef785b7225dcf895a1640147aa25dfd2f353cf9a6072",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "size": 1235,
         "digest": "sha256:b8398cec700268886cb888081b45d7aeacef8a726bd28c33658190ff8734ed78",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}
```

Have there been any plans about up-streaming these changes and around when would that be expected?

**Comment by jonjohnsonjr** [comment]
Pinging back here for a quick update. I just landed https://github.com/google/go-containerregistry/pull/520 which I believe is everything I need upstream to implement this "for real". I'll try to revisit this PR in the next couple weeks.

@mattmoor had a nice idea that the default flag value for platforms should depend on which command we're invoking:

* For `ko apply`, we're almost certainly deploying to a single platform, which is probably amd64/linux, so we should default to `--platform=amd64/linux`. (Possibly use `kubectl` to determine the target cluster's platform as well)
* For `ko resolve`, we're *probably* producing a release, so we could default to `--platform=all`.
* For `ko publish`, we don't know where it's going, I'm tempted to just default to `--platform=amd64/linux` for now.

**Comment by junawaneshivani** [comment]
Hi @jonjohnsonjr, any further progress on this PR?

**Comment by jonjohnsonjr** [comment]
> any further progress on this PR?

Just pushed something that I'm pretty happy with. I'm not 100% certain what we want to do with the `--platform` flag, but this seems pretty good.

**Comment by junawaneshivani** [comment]
@jonjohnsonjr I see there have been few commits since I last checked.. Does the `--platform` flag produce images for the right architecture now ? (given we have used a multi-arch base image )

**Comment by junawaneshivani** [comment]
@jonjohnsonjr I see there have been few commits since I last checked.. Does the `--platform` flag produce images for the right architecture now ? (given we have used a multi-arch base image )

**Comment by jonjohnsonjr** [comment]
This branch seems to work for me (from an image perspective), but I don't have various architectures to test against, so I'm not sure if the cross-platform go compilation stuff is working correctly.

I just rebased this against master, any affirmation that this branch works would help towards code review 😄

**Comment by junawaneshivani** [comment]
@jonjohnsonjr I will work on validating the master branch for power architecture. Thanks 😄

**Comment by junawaneshivani** [comment]
@jonjohnsonjr .. I was not able to build ppc64le specific images... Let me know if I am missing something here..
```
root@p006vm31 google]# git clone https://github.com/jonjohnsonjr/ko/tree/multi-platform-wip
Cloning into 'multi-platform-wip'...
fatal: repository 'https://github.com/jonjohnsonjr/ko/tree/multi-platform-wip/' not found
[root@p006vm31 google]# git clone --branch multi-platform-wip https://github.com/jonjohnsonjr/ko/
Cloning into 'ko'...
remote: Enumerating objects: 117, done.
remote: Counting objects: 100% (117/117), done.
remote: Compressing objects: 100% (81/81), done.
remote: Total 5363 (delta 48), reused 74 (delta 34), pack-reused 5246
Receiving objects: 100% (5363/5363), 7.60 MiB | 4.99 MiB/s, done.
Resolving deltas: 100% (2451/2451), done.
[root@p006vm31 google]# pwd
/root/go/src/github.com/google
[root@p006vm31 google]# cd ko/cmd/ko/
[root@p006vm31 ko]# go build
[root@p006vm31 ko]# ls
ko  main.go  test
[root@p006vm31 ko]# which ko
/root/go/bin/ko
[root@p006vm31 ko]# cp ko /root/go/bin/
cp: overwrite '/root/go/bin/ko'? y
[root@p006vm31 ko]# cat .ko.yaml
defaultBaseImage: registry.access.redhat.com/ubi7/ubi-minimal:latest
[root@p006vm31 ko]# echo $GOOS
linux
[root@p006vm31 ko]# echo $GOARCH
ppc64le
[root@p006vm31 ko]# ko publish --local ./cmd/ko
2020/02/21 06:09:40 Using base registry.access.redhat.com/ubi7/ubi-minimal:latest for github.com/google/ko/cmd/ko
2020/02/21 06:09:41 Building github.com/google/ko/cmd/ko
2020/02/21 06:09:49 Loading ko.local/ko-099ba5bcefdead87f92606265fb99ac0:89b4ce3c5dbe0f9124f057aaa136c433e0aaf3e2b95230e96ae6dc6662cb7fcc
2020/02/21 06:09:51 Loaded ko.local/ko-099ba5bcefdead87f92606265fb99ac0:89b4ce3c5dbe0f9124f057aaa136c433e0aaf3e2b95230e96ae6dc6662cb7fcc
2020/02/21 06:09:51 Adding tag latest
2020/02/21 06:09:51 Added tag latest
ko.local/ko-099ba5bcefdead87f92606265fb99ac0:89b4ce3c5dbe0f9124f057aaa136c433e0aaf3e2b95230e96ae6dc6662cb7fcc
[root@p006vm31 ko]# docker image inspect ko.local/ko-099ba5bcefdead87f92606265fb99ac0:89b4ce3c5dbe0f9124f057aaa136c433e0aaf3e2b95230e96ae6dc6662cb7fcc | grep Arch
        "Architecture": "amd64",
```

**Comment by lisa** [comment]
@junawaneshivani In regard to getting `Architecture: amd64`, that can be a misnomer. If you built on a native amd64 host (eg cross-compiling for ppc64le), by default, you will get this behaviour. It turns out that the `Architecture` in the image spec is not (currently) used, and therefor not the best indicator to see what the arch of binaries inside. 

If you'd like to set that to the correct architecture indicator, I wrote a Makefile to do that, which is a part of a larger presentation given at a conference. You can find the specific Makefile steps in [lisa/lisa19-containers](https://github.com/lisa/lisa19-containers/blob/badce255bb6e2c0ac2ae71666d6d839a6be0aa34/functions.mk#L16-L41), and the larger context is available in [that repo's readme](https://github.com/lisa/lisa19-containers/blob/master/README.md).

**Comment by jonjohnsonjr** [comment]
> Let me know if I am missing something here

The introduction of the `--platform` flag supersedes the GOOS/GOARCH environment variables from before, since we (I think) always want our builds to match the underlying base image, I don't think it makes sense to allow overriding GOOS/GOARCH anymore, but that might be a bad assumption!

Just tested this with the platform flag:

```
$ cat .ko.yaml 
defaultBaseImage: registry.access.redhat.com/ubi7/ubi-minimal:latest

$ crane config $(ko publish --platform linux/ppc64le ./cmd/ko) | jq .
2020/02/21 11:22:24 Using base registry.access.redhat.com/ubi7/ubi-minimal:latest for github.com/google/ko/cmd/ko
2020/02/21 11:22:26 Building github.com/google/ko/cmd/ko
2020/02/21 11:22:32 Publishing gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0:latest
2020/02/21 11:22:33 Published gcr.io/jonjohnson-test/ko/ko-099ba5bcefdead87f92606265fb99ac0@sha256:f7f23fa7530725d1aa0c3b31467ac51c198f859db17aa6ee8da51d494e08502b
{
  "architecture": "ppc64le",
  "author": "github.com/google/ko",
  "created": "2020-01-28T10:59:36.069978Z",
  "docker_version": "1.13.1",
  "history": [
    {
      "created": "2020-01-28T10:59:12.703525002Z",
      "comment": "Imported from -"
    },
    {
      "created": "2020-01-28T10:59:36.069978Z"
    },
    {
      "author": "ko",
      "created": "0001-01-01T00:00:00Z",
      "created_by": "ko publish github.com/google/ko/cmd/ko",
      "comment": "kodata contents, at $KO_DATA_PATH"
    },
    {
      "author": "ko",
      "created": "0001-01-01T00:00:00Z",
      "created_by": "ko publish github.com/google/ko/cmd/ko",
      "comment": "go build output, at /ko-app/ko"
    }
  ],
  "os": "linux",
  "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:5e0e8674338a2c81041a1ff7197711332f1fad73d57da7d3131f6adbeb249496",
      "sha256:932a3377207350683ba8e6d0d733dfac7c1513942011d7341673d3e1755834f5",
      "sha256:4f76c7cc15474e09ddbebb4b39759df491cde9af6c85700612e7ec756492d571",
      "sha256:d2c3519551e749b6f5e96e333bbf9e0ed1b180ba711aca62e992e4dcaf38f57e"
    ]
  },
  "config": {
    "Cmd": [
      "/bin/bash"
    ],
    "Entrypoint": [
      "/ko-app/ko"
    ],
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/ko-app",
      "container=oci",
      "KO_DATA_PATH=/var/run/ko"
    ],
    "Hostname": "2a85853ff11a",
    "Image": "caf773ea068dc19a68a8bf71855b93a2776c00f49b8c9db627f9abfccbcf63a9",
    "Labels": {
      "architecture": "ppc64le",
      "authoritative-source-url": "registry.access.redhat.com",
      "build-date": "2020-01-28T10:59:03.338808",
      "com.redhat.build-host": "ppc64le-c2-vm-06.prod.osbs.eng.bos.redhat.com",
      "com.redhat.component": "ubi7-minimal-container",
      "com.redhat.license_terms": "https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI",
      "description": "The Universal Base Image Minimal is a stripped down image that uses microdnf as a package manager. This base image is freely redistributable, but Red Hat only supports Red Hat technologies through subscriptions for Red Hat products. This image is maintained by Red Hat and updated regularly.",
      "distribution-scope": "public",
      "io.k8s.description": "The Universal Base Image Minimal is a stripped down image that uses microdnf as a package manager. This base image is freely redistributable, but Red Hat only supports Red Hat technologies through subscriptions for Red Hat products. This image is maintained by Red Hat and updated regularly.",
      "io.k8s.display-name": "Red Hat Universal Base Image 7 Minimal",
      "io.openshift.tags": "minimal rhel7",
      "maintainer": "Red Hat, Inc.",
      "name": "ubi7-minimal",
      "release": "303",
      "summary": "Provides the latest release of the minimal Red Hat Universal Base Image 7.",
      "url": "https://access.redhat.com/containers/#/registry.access.redhat.com/ubi7-minimal/images/7.7-303",
      "vcs-ref": "7e9dd8b61a03561c4b5f70b214bc26d19c4f25a4",
      "vcs-type": "git",
      "vendor": "Red Hat, Inc.",
      "version": "7.7"
    },
    "ArgsEscaped": true
  }
}
```

Let me know if that works and if you think I should bring back GOOS/GOARCH handling. We _could_ make that a way to specify the default platform instead of assuming linux/amd64?

**Comment by junawaneshivani** [comment]
Thanks @jonjohnsonjr ... I had tried using the `--platform` tag but with `ppc64le/linux` value instead of `linux/ppc64le` which gave an error `no child with platform`.. I referred below comment..

> For ko apply, we're almost certainly deploying to a single platform, which is probably amd64/linux, so we should default to --platform=amd64/linux. (Possibly use kubectl to determine the target cluster's platform as well)

I am now able to build ppc64le images by using the right tag value... 

```
[root@p006vm31 ko]#  GO111MODULE=on  ko publish --local --platform=linux/ppc64le ./cmd/ko
2020/02/23 08:54:06 Using base registry.access.redhat.com/ubi7/ubi-minimal:latest for github.com/google/ko/cmd/ko
2020/02/23 08:54:07 Building github.com/google/ko/cmd/ko
2020/02/23 08:54:15 Loading ko.local/ko-099ba5bcefdead87f92606265fb99ac0:222cf6c149e162e11c492aa2ab2d0fbfdc6073caea9425ed6d19131c8004b2c5
2020/02/23 08:54:24 Loaded ko.local/ko-099ba5bcefdead87f92606265fb99ac0:222cf6c149e162e11c492aa2ab2d0fbfdc6073caea9425ed6d19131c8004b2c5
2020/02/23 08:54:24 Adding tag latest
2020/02/23 08:54:25 Added tag latest
ko.local/ko-099ba5bcefdead87f92606265fb99ac0:222cf6c149e162e11c492aa2ab2d0fbfdc6073caea9425ed6d19131c8004b2c5
[root@p006vm31 ko]# docker inspect ko.local/ko-099ba5bcefdead87f92606265fb99ac0:222cf6c149e162e11c492aa2ab2d0fbfdc6073caea9425ed6d19131c8004b2c5 | grep Arch
        "Architecture": "ppc64le",
```


> Let me know if that works and if you think I should bring back GOOS/GOARCH handling. We could make that a way to specify the default platform instead of assuming linux/amd64?

Yes, I do believe that instead of assuming `linux/amd64` if we could use `GOOS/GOARCH` for setting the default value for `--platform` tag, it would be great... 

All in all the changes look good to me... This would help many communities in providing ppc64le support.. Thanks for all the help so far.. 😄

**Comment by jonjohnsonjr** [comment]
@junawaneshivani thanks!

Can you also confirm that running these images _actually_ works on a ppc64le machine? I'm under the impression that it's sufficient to just set GOOS/GOARCH to perform cross-compilation, but the existence of tools like https://github.com/laher/goxc make me think it may be harder than that...

**Comment by junawaneshivani** [comment]
@jonjohnsonjr Yes, one of my colleague validated that the images work on power..

**Comment by barthy1** [comment]
@jonjohnsonjr I verified your code with s390x arch. 
I did the build on `linux/amd64` with `defaultBaseImage: alpine`
```
# GO111MODULE=on  ko publish --local --platform=linux/s390x ./cmd/ko
2020/03/17 22:02:29 Using base alpine for github.com/google/ko/cmd/ko
2020/03/17 22:02:31 Building github.com/google/ko/cmd/ko
2020/03/17 22:02:36 Loading ko.local/ko-099ba5bcefdead87f92606265fb99ac0:43ed13bc4c7ac96db03fe97f164405700e09f7f309bbf7897418a27b6bb6b46f
2020/03/17 22:02:36 Loaded ko.local/ko-099ba5bcefdead87f92606265fb99ac0:43ed13bc4c7ac96db03fe97f164405700e09f7f309bbf7897418a27b6bb6b46f
2020/03/17 22:02:36 Adding tag latest
2020/03/17 22:02:36 Added tag latest
ko.local/ko-099ba5bcefdead87f92606265fb99ac0:43ed13bc4c7ac96db03fe97f164405700e09f7f309bbf7897418a27b6bb6b46f
# docker inspect ko.local/ko-099ba5bcefdead87f92606265fb99ac0:43ed13bc4c7ac96db03fe97f164405700e09f7f309bbf7897418a27b6bb6b46f | grep Arch
        "Architecture": "s390x",
```
and was able to run it on `linux/s390x` machine.
```
# docker run -it --rm --entrypoint="/bin/uname" ko.local/ko-099ba5bcefdead87f92606265fb99ac0:43ed13bc4c7ac96db03fe97f164405700e09f7f309bbf7897418a27b6bb6b46f -a
Linux 30712c36b68d 5.3.7-200.fc30.s390x #1 SMP Fri Oct 18 19:49:30 UTC 2019 s390x Linux
``` 

Same behaviour I've got when I did the build and run on `linux/s390x` machine only.
Great work!

**Comment by barthy1** [comment]
Hi @jonjohnsonjr 
this PR is very important for tekton s390x and ppc64le enablement - https://github.com/tektoncd/pipeline/issues/856
Do you see how we can help you from Z/Power side with your work in this direction?

**Comment by seth-priya** [comment]
@jonjohnsonjr just wanted to check is there is any work remaining in terms of validating these changes with which we @junawaneshivani and myself on the Power side can help with? or any other blockers that need to be worked through? thanks in advance!

File: pkg/build/build.go [hunk 2]
```diff
@@ -28,6 +29,15 @@ type Interface interface {
...
+	// (or a set of images as a v1.ImageIndex).
+	Build(context.Context, string) (Result, error)
+}
+
+// Result represents the product of a Build. This is usually a v1.Image or v1.ImageIndex.
+type Result interface {
+	MediaType() (types.MediaType, error)
+	Size() (int64, error)
+	Digest() (v1.Hash, error)
+	RawManifest() ([]byte, error)
```
mattmoor, line 42: Can you add `var _ v1.Image = (Result)(nil)` below this for Image and ImageIndex?

pkg/build/gobuild.go
```diff
@@ -563,3 +560,72 @@ func updatePath(cf *v1.ConfigFile) {
...
+func (gb *gobuild) buildAll(ctx context.Context, s string, base v1.ImageIndex) (v1.ImageIndex, error) {
+	im, err := base.IndexManifest()
+	if err != nil {
+		return nil, err
+	}
+
+	// Build an image for each child from the base and append it to a new index to produce the result.
+	adds := []mutate.IndexAddendum{}
+	for _, desc := range im.Manifests {
+		// TODO(jonjohnsonjr): This would fail for a nested index, which is exceedingly rare.
```
mattmoor, line 605: Doesn't the descriptor have a mime type, which would let you detect this and fail gracefully?

mattmoor: A unit test would be good for this case.

pkg/build/gobuild.go
```diff
@@ -563,3 +560,72 @@ func updatePath(cf *v1.ConfigFile) {
...
+		base, err := base.Image(desc.Digest)
+		if err != nil {
+			return nil, err
+		}
+		img, err := gb.buildOne(ctx, s, base)
```
mattmoor, line 610: Should we do these in parallel?

mattmoor: `TODO(#1234)` would be fine, I think

pkg/build/recorder.go
```diff
@@ -36,8 +34,7 @@ func (r *Recorder) IsSupportedReference(ip string) bool {
 	return r.Builder.IsSupportedReference(ip)
 }
 
-// Build implements Interface
-func (r *Recorder) Build(ctx context.Context, ip string) (v1.Image, error) {
+func (r *Recorder) Build(ctx context.Context, ip string) (Result, error) {
```
mattmoor, line 37: Why drop the comment?

pkg/commands/config.go
```diff
@@ -37,21 +39,55 @@ var (
...
+			if len(parts) > 0 {
+				p.OS = parts[0]
+			}
+			if len(parts) > 1 {
+				p.Architecture = parts[1]
+			}
+			if len(parts) > 2 {
+				p.Variant = parts[2]
+			}
+			ropt = append(ropt, remote.WithPlatform(p))
```
mattmoor, line 73: Error if `len(parts) > 3`?

pkg/commands/options/build.go
```diff
@@ -24,11 +24,15 @@ import (
...
 }
 
 func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
 	cmd.Flags().IntVarP(&bo.ConcurrentBuilds, "jobs", "j", runtime.GOMAXPROCS(0),
 		"The maximum number of concurrent builds")
 	cmd.Flags().BoolVar(&bo.DisableOptimizations, "disable-optimizations", bo.DisableOptimizations,
 		"Disable optimizations when building Go code. Useful when you want to interactively debug the created container.")
+	cmd.Flags().StringVar(&bo.Platform, "platform", "linux/amd64",
+		"Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]]")
+
```
mattmoor, line 37: Isn't this technically a breaking change if folks are using `GOOS/GOARCH` with `ko` today?

pkg/publish/default_test.go
```diff
@@ -93,133 +74,82 @@ func md5Hash(s string) string {
...
-
-	createdTags := make(map[string]struct{})
-
-	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.Method == http.MethodHead && strings.HasPrefix(r.URL.Path, headPathPrefix) && r.URL.Path != initiatePath {
-			http.Error(w, "NotFound", http.StatusNotFound)
-			return
+	for _, br := range []build.Result{img, idx} {
+		base := "blah"
+		importpath := "github.com/Google/go-containerregistry/cmd/crane"
```
mattmoor, line 111:
```suggestion
		importpath := "github.com/google/go-containerregistry/cmd/crane"
```

pkg/publish/default.go
```diff
@@ -94,8 +95,32 @@ func NewDefault(base string, options ...Option) (Interface, error) {
...
+	case types.OCIManifestSchema1, types.DockerManifestSchema2:
+		img, ok := br.(v1.Image)
+		if !ok {
+			return fmt.Errorf("failed to interpret result as image: %v", br)
+		}
+		return remote.Write(tag, img, opt...)
+	default:
+		return fmt.Errorf("result image media type: %s", mt)
+	}
+}
```
mattmoor, line 120: I sense a pattern emerging

mattmoor:
Generally, this seems like a good first chunk to land, but we should open issues for "smarter" platform modes.

Honestly, looking through, I'm pretty happy with how simple this was.  Most of the diff seemed like changing signatures, which is pretty cool.

I'd like to get an early form of this in, land #162 and then cut 0.6 🤩

mattmoor:
Oh, one other thing I've been thinking is that when the base image is an image index, it would be good for use to log what platforms we see and build for logged to stderr.

pkg/publish/default_test.go
jonjohnsonjr: This was for https://github.com/google/ko/blob/bec089d9c82eadfbd3e772395b1581709bc2055a/pkg/publish/default.go#L100-L101

pkg/build/gobuild.go
jonjohnsonjr: I would be worried even more about cache invalidation, will file an issue to experiment and follow-up, then TODO(it).

pkg/publish/default.go
jonjohnsonjr: Indeed... could probably do better than this but it's easy enough to understand.

pkg/build/recorder.go
jonjohnsonjr: Oops.

pkg/build/build.go
jonjohnsonjr: Done.

pkg/build/gobuild.go
jonjohnsonjr: Done.

pkg/commands/options/build.go
jonjohnsonjr: Done (see pkg/commands/config.go)

pkg/build/gobuild.go
jonjohnsonjr: Done.

pkg/publish/default.go
mattmoor: Yeah, I wasn't expecting you do change anything here, but we could probably find ways to share the boilerplate if it starts to diverge.

jonjohnsonjr:
> Oh, one other thing I've been thinking is that when the base image is an image index, it would be good for use to log what platforms we see and build for logged to stderr.

I did this really lazily by just logging architecture every time in our `Building ...` log, LMK if this is too lazy.

mattmoor:
I saw, seemed fine.

vdemeester:
🎉

PR #40, opened 2019-06-18
Don't encode nil objects

### LINKED ISSUE #39: Files ending with --- generate nulls in output
Repro: 
```
$ echo -e "foo\n---" | ko resolve -f -
foo
--- null

---
```

See https://github.com/knative/serving/issues/4333

jonjohnsonjr:
Fixes #39

### DIFF
--- a/pkg/resolve/resolve.go
+++ b/pkg/resolve/resolve.go
@@ -102,8 +102,10 @@ func ImageReferences(input []byte, builder build.Interface, publisher publish.In
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Not 100% sure on this one -- this might not work well if the input document intentionally contains nulls...

jonjohnsonjr:
Yeah looks like this isn't quite there:
```
$ echo -e "null\n---" | ko resolve -f -

---
```

Issue #31, opened 2019-04-27
ko resolve support for label filtering


mattmoor:
With `ko apply` I can supply a label filter to only deploy the pieces matching a particular label filter.

For example:
```
ko apply -f config -l "networking.knative.dev/ingress-provider!=istio"
```

Will install knative/serving without the Istio elements.  However, there are two problems with this:
1. This still builds filtered things (this is handled by `kubectl`).
2. This doesn't work with `ko resolve`.

I'd like to leverage the latter portion of this to enable us to release separate release bundles, e.g. `config/` would contain all of the "default" implementations, but we can still create split release bundles for core, istio, cert-manager separately using the label filter.

cc @n3wscott @jonjohnsonjr @ImJasonH

mattmoor:
If we can land this in the next week or so, we can use this to release 0.6 👍

jonjohnsonjr:
Fixed by https://github.com/google/ko/pull/46

PR #46, opened 2019-06-25
add selector flag and logic

ilackarms:
this pr is intended to solve https://github.com/google/ko/issues/31

by passing a selector via a cli flag `--selector` or `-l`, users can now filter resources out of manifests

some notes (not sure how the maintainers will feel on this):
- this pr leverages some libraries from `apimachinery` and `controller-runtime` to handle yaml parsing, so `Gopkg.lock` has been updated
-  comments in the original yaml will be lost and map order is not preserved when the a selector is specified (objects have to be deserialized and then serialized again in order to check their labels)
- in the case of `List` objects, individual items are filtered out of the list based on the provided selector
- unit testing could probably be improved, let me know if there are more edge/error cases that are worth testing


ilackarms:
this should work with `create`, `apply` and `resolve`

ImJasonH:
Thanks for doing this, it seems really useful!

Can you update docs to describe this option in more detail? It would be great to have a short section describing how to use this in the README.md.

pkg/resolve/selector.go
```diff
@@ -0,0 +1,116 @@
+package resolve
```
jonjohnsonjr: nit: missing Copyright header

pkg/resolve/selector.go
```diff
@@ -0,0 +1,116 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	yaml2json "k8s.io/apimachinery/pkg/util/yaml"
+	"regexp"
+	"sigs.k8s.io/yaml"
+	"strings"
+)
+
+
```
jonjohnsonjr: nit: extra newline here

pkg/resolve/selector.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// re-join the objects into a single manifest
+	return bytes.Join(outputObjectsYaml, []byte("---\n")), nil
+}
+
+var yamlSeparatorRegex = regexp.MustCompile( "\n---")
+
+func parseUnstructured(rawYaml []byte) ([]runtime.Object, error) {
+	objectYamls := yamlSeparatorRegex.Split(string(rawYaml), -1)
```
jonjohnsonjr: You could use [this](https://github.com/kubernetes/apimachinery/blob/ed135c5b96450fd24e5e981c708114fbbd950697/pkg/util/yaml/decoder.go#L93) for splitting the document -- I'm not sure if it will look any cleaner, though.

pkg/resolve/selector.go
```diff
@@ -0,0 +1,116 @@
...
+	var resources []runtime.Object
+
+	for _, objectYaml := range objectYamls {
+		// empty yaml snippets, such as those which can be
+		// generated by helm should be ignored
+		// else they may be parsed into empty map[string]interface{} objects
+		if isEmptyYamlSnippet(objectYaml) {
+			continue
+		}
+		jsn, err := yaml2json.ToJSON([]byte(objectYaml))
```
jonjohnsonjr: There's [this](https://github.com/kubernetes/apimachinery/blob/ed135c5b96450fd24e5e981c708114fbbd950697/pkg/runtime/serializer/yaml/yaml.go#L39) as well.

pkg/resolve/selector.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	// parse runtime.Objects from the input yaml
+	objects, err := parseUnstructured(input)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, object := range objects {
+		// objects parsed by UnstructuredJSONScheme can only be of
```
jonjohnsonjr: We might want to log something or panic in a default case in case this changes, just to make sure we're not silently dropping documents?

jonjohnsonjr:
@ilackarms it seems like we could work around the comment issue by just passing through the original document's yaml bytes for `unstructured.Unstructured` instead of re-encoding to json and converting back to yaml.

For `unstructured.UnstructuredList`, we could similarly skip the re-encoding if no objects were actually filtered out of the list.

WDYT?

ilackarms:
@jonjohnsonjr great suggestions. the only reason i didn't go with this idea in the first place was specifically because of the list case... but we can be greedy and preserve the original YAML in all cases except when a list is modified

i'll follow up with updates on this pr

pkg/resolve/selector.go
ilackarms: the problem is that this won't let me use `runtime.Decode(unstructured.UnstructuredJSONScheme) `

pkg/resolve/selector.go
ilackarms: does that require me to know the `GroupVersionKind` for the yaml? seems more straightforward to do what i've done here. what do you think?

README.md
```diff
@@ -230,6 +238,8 @@ rebuild, repush, and redeploy their changes.
 `ko apply` will invoke `kubectl apply` under the covers, and therefore apply
 to whatever `kubectl` context is active.
 
+`ko apply`, when supplied the `--selector` flag, will apply only the resources that are selected by the provided selector.
```
jonjohnsonjr: This seems redundant with the other section.

README.md
```diff
@@ -219,6 +219,14 @@ spec:
 It is notable that this is not the default (anymore) because certain popular
 registries (including Docker Hub) do not support multi-level repository names.
 
+`ko resolve`, `ko apply`, and `ko create` accept an optional `--selector` or `-l`
```
jonjohnsonjr: Part of me wants to move this down to the "release management" section, because that's where I see this being most useful... though it could always be useful during development. WDYT?

ilackarms: my thinking was that it's useful for any `ko` user who wants to run these commands... they might not be checking the "release management" section (as in my case, i've been using it as part of the serving e2e tests)

PR #58, opened 2019-07-10
Implement "strict mode" which requires `ko://` prefix for import paths

ImJasonH:
Demonstrating a strawman implementation and UX for "strict mode", as described in #9

In strict mode, image references are only considered if they begin with the string `ko://`. In loose mode, `ko://` refs are considered as image references and built and replaced during YAML resolution.

If we decide to pursue this model, we could start by making strictness opt-in, cut a versioned release of `ko`, then make `--strict=true` by default (opt-out), then eventually maybe even remove the option.

```
$ cat test.yaml
looseImage: github.com/google/ko/cmd/ko
strictImage: ko://github.com/google/ko/cmd/ko
$ KO_DOCKER_REPO=gcr.io/foo go run ./cmd/ko/ resolve -f test.yaml
...
looseImage: gcr.io/foo/ko-099ba5bcefdead87f92606265fb99ac0@sha256:dbe1a3e11de9e701c9e7a5e61d4f3c2dc93eee2552bb7f6831a23c1147e6899a
strictImage: gcr.io/foo/ko-099ba5bcefdead87f92606265fb99ac0@sha256:dbe1a3e11de9e701c9e7a5e61d4f3c2dc93eee2552bb7f6831a23c1147e6899a

---
$ KO_DOCKER_REPO=gcr.io/foo go run ./cmd/ko/ resolve -f test.yaml -S
...
looseImage: github.com/google/ko/cmd/ko
strictImage: gcr.io/foo/ko-099ba5bcefdead87f92606265fb99ac0@sha256:dbe1a3e11de9e701c9e7a5e61d4f3c2dc93eee2552bb7f6831a23c1147e6899a
---
```

Sending this out to get feedback on the overall idea; it sounds like Go modules is going to be pretty painful without this as an option at least.


ImJasonH:
I'd like to push this forward soon, there have been a lot of issues that I think this would solve. The PR isn't ready yet -- it doesn't correctly fail when a `ko://`-prefixed path isn't supported -- but I'll ping when it's ready.

ImJasonH:
```
$ cat test.yaml 
looseImage: github.com/google/ko/cmd/ko          # unresolved in strict mode
strictImage: ko://github.com/google/ko/cmd/ko    # resolved in any mode
badImage: github.com/blargh/whatever             # unresolved in either mode
badStrictImage: ko://github.com/blargh/whatever  # resolved in loose mode; fails in strict mode

# loose mode
$ KO_DOCKER_REPO=gcr.io/my-project go run ./cmd/ko/ resolve -f test.yaml
badImage: github.com/blargh/whatever
badStrictImage: ko://github.com/blargh/whatever
looseImage: gcr.io/my-project/ko-099ba5bcefdead87f92606265fb99ac0@sha256:174089f40c1c0f9f5103c7858a39827ed50b80c6a7b877cf63714b8029b95fcf
strictImage: gcr.io/my-project/ko-099ba5bcefdead87f92606265fb99ac0@sha256:174089f40c1c0f9f5103c7858a39827ed50b80c6a7b877cf63714b8029b95fcf

# strict mode
$ KO_DOCKER_REPO=gcr.io/my-project go run ./cmd/ko/ resolve -S -f test.yaml
2019/08/15 09:42:47 error processing import paths in "test.yaml": Strict reference "ko://github.com/blargh/whatever" is not supported
```

If you remove `badStrictImage` the strict-mode resolved output is:

```
badImage: github.com/blargh/whatever
looseImage: github.com/google/ko/cmd/ko
strictImage: gcr.io/test-argo/ko-099ba5bcefdead87f92606265fb99ac0@sha256:174089f40c1c0f9f5103c7858a39827ed50b80c6a7b877cf63714b8029b95fcf
```

Non-`ko://`-prefixed paths are _ignored_ even if they're suppoorted. Only paths that begin with `ko://` are considered in strict mode.

pkg/resolve/resolve.go
```diff
@@ -45,8 +46,14 @@ func ImageReferences(input []byte, builder build.Interface, publisher publish.In
...
-			if builder.IsSupportedReference(ref) {
-				refs[ref] = struct{}{}
+			if strict && !strings.HasPrefix(ref, "ko://") {
+				return ref, nil
+			}
+			tref := strings.TrimPrefix(ref, "ko://")
+			if builder.IsSupportedReference(tref) {
+				refs[tref] = struct{}{}
+			} else if strict && strings.HasPrefix(ref, "ko://") {
+				return "", fmt.Errorf("Strict reference %q is not supported", ref)
```
jonjohnsonjr: "is not supported" makes sense from our go interfaces, but isn't particularly helpful from a CLI UX perspective.

You could cheat here and just call `builder.Build(tref)` and return whatever error it returns? I'm actually not sure what would happen but it's worth trying.

ImJasonH: Good feedback, changed to:

```
2019/08/15 15:01:51 error processing import paths in "test.yaml": Found strict reference "ko://github.com/blargh/whatever" but github.com/blargh/whatever is not a valid import path
```

pkg/resolve/resolve.go
```diff
@@ -45,8 +46,14 @@ func ImageReferences(input []byte, builder build.Interface, publisher publish.In
 		}
 		// This simply returns the replaced object, which we discard during the gathering phase.
 		if _, err := replaceRecursive(obj, func(ref string) (string, error) {
-			if builder.IsSupportedReference(ref) {
-				refs[ref] = struct{}{}
+			if strict && !strings.HasPrefix(ref, "ko://") {
```
jonjohnsonjr: You could dry this up a bit -- reuse `tref` on line 56 and pull `strings.HasPrefix(ref, "ko://")` out into a boolean, maybe `strictRef`.

ImJasonH: Good call, done.

pkg/commands/options/strict.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/spf13/cobra"
+)
+
+// StrictOptions holds options to require strict references.
+type StrictOptions struct {
+	Strict bool
+}
+
+func AddStrictArg(cmd *cobra.Command, so *StrictOptions) {
+	cmd.Flags().BoolVarP(&so.Strict, "strict", "S", so.Strict,
```
jonjohnsonjr: I'm mildly -1 on including `-S` -- we will eventually run out of flags if we keep this up :P

pkg/commands/options/strict.go
```diff
@@ -0,0 +1,29 @@
...
+		"If true, require package references to be explicitly noted")
```
jonjohnsonjr: ... be explicitly prefixed with 'ko://'

also we will want to update the README (maybe not in this PR if we're still experimenting, but eventually)

pkg/commands/options/strict.go
ImJasonH: Yeah, I kind of agree. I'll remove `-S` until someone complains it's too much to type.

pkg/commands/options/strict.go
ImJasonH: I was planning to update the README in a followup PR, if that's okay.

PR #68, opened 2019-08-12
Add permissions to the kodata directory.

mattmoor:
I suspect that this may be a source of problems reading kodata with nonroot,
but it is probably worth doing either way.


dprotaso:
> Additionally, windows can only set 0222, 0444, or 0666, none of which are executable.

For my curiosity do you have a reference/link?

mattmoor:
@dprotaso this is copy/pasta from (IIRC) @evankanderson's comments elsewhere in this file.

PR #73, opened 2019-08-15
Add support for recursively resolving directory symlinks.

mattmoor:
This adds support for properly resolving directory symlinks within kodata.
I verified that with this I can symlink `.git/refs` into `kodata/` and (with
changes) resolve the `.git/HEAD` symlink to read the appropriate
`ref: refs/heads/...` files with the commit SHA.


pkg/build/gobuild.go
```diff
@@ -318,6 +302,17 @@ func (g *gobuild) tarKoData(importpath string) (*bytes.Buffer, error) {
...
 			return err
 		}
+		newPath := filepath.Join(chroot, path[len(root):])
+
+		path, err = filepath.EvalSymlinks(path)
+		if err != nil {
+			return err
+		}
+
+		// Skip other directories.
```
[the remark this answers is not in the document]
jonjohnsonjr: sg

jonjohnsonjr:
```
    --- FAIL: TestGoBuild/check_app_layer_contents (0.00s)
        gobuild_test.go:265: Digest() = sha256:b732ce9d2fe1042fcbee9d3d4682ed06336c65b55cd04462ae4dc26f752e7699, want sha256:4379f30a6c6f66221c3c54dddd378fcfa5a7304a6655ff783b102069c0f943ab
```

pkg/build/gobuild.go
```diff
@@ -342,6 +338,26 @@ func (g *gobuild) tarKoData(importpath string) (*bytes.Buffer, error) {
...
+}
+
+func (g *gobuild) tarKoData(importpath string) (*bytes.Buffer, error) {
+	buf := bytes.NewBuffer(nil)
+	// Compress this before calling tarball.LayerFromOpener, since it eagerly
+	// calculates digests and diffids. This prevents us from double compressing
+	// the layer when we have to actually upload the blob.
+	//
+	// https://github.com/google/go-containerregistry/issues/413
+	gw, _ := gzip.NewWriterLevel(buf, gzip.BestSpeed)
```
vagababov: check error?

pkg/build/gobuild.go
```diff
@@ -342,6 +338,26 @@ func (g *gobuild) tarKoData(importpath string) (*bytes.Buffer, error) {
...
+	defer gw.Close()
+	tw := tar.NewWriter(gw)
+	defer tw.Close()
+
+	root, err := g.kodataPath(importpath)
+	if err != nil {
+		return nil, err
+	}
+
+	err = walkRecursive(tw, root, kodataRoot)
```
vagababov: why not just `return buf, err`?

pkg/build/gobuild.go
jonjohnsonjr: > The error returned will be nil if the level is valid.

The level is valid 😄

vagababov: ┻━┻︵ \(°□°)/ ︵ ┻━┻

pkg/build/gobuild.go
mattmoor: Doing that without copies sounds like a PITA with recursion.  This way the N calls can write to the same `buf` directly.

jonjohnsonjr: I think he means you have redundant error checking here when you could just do:
```go
return buf, walkRecursive(tw, root, kodataRoot)
```

mattmoor: done

PR #77, opened 2019-08-16
Use GOOS/GOARCH from base image

jonjohnsonjr:
Small progress towards #6

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -41,7 +41,7 @@ const (
 
 // GetBase takes an importpath and returns a base v1.Image.
 type GetBase func(string) (v1.Image, error)
-type builder func(string, bool) (string, error)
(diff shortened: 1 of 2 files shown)


lysannef:
@jonjohnsonjr  This title indicates to  generating multi arch behaviour. Is this in any way related to the PR #38.

jonjohnsonjr:
@lysannef yes, it's some progress towards it. Instead of assuming amd64/linux, this will get the OS and Arch from the underlying base image. The default base image is amd64/linux, but you could [override the base image](https://github.com/google/ko#overriding-the-default-base-image) to use an image built for a different platform, and things should just work.

You could achieve this before by setting `GOOS` and `GOARCH` in your environment, so this isn't a huge change, but it does include some interface changes that I'll need for #38.

PR #79, opened 2019-09-10
Add build.Limiter

jonjohnsonjr:
You can limit the number of concurrent builds with -j (a la make).

The default value for this is GOMAXPROCS, which seems reasonable.

cc @JungukCho


pkg/build/limit.go
```diff
@@ -0,0 +1,55 @@
...
+
+// IsSupportedReference implements Interface
+func (l *Limiter) IsSupportedReference(ip string) bool {
+	return l.Builder.IsSupportedReference(ip)
+}
+
+// Build implements Interface
+func (l *Limiter) Build(ip string) (v1.Image, error) {
+	// TODO(jonjohnsonjr): Build should take a context.Context.
+	if err := l.semaphore.Acquire(context.TODO(), 1); err != nil {
```
ImJasonH: Is there any reason not to use [`sync.WaitGroup`](https://golang.org/pkg/sync/#WaitGroup)? I'm unfamiliar with the `semaphore` package, but it seems like it's main benefit is weightedness, which we're not taking advantage of here.

jonjohnsonjr: WaitGroup is similar, but AFAICT there is no way to bound the maximum number of the counter, whereas NewWeighted takes a maximum weight.

PR #88, opened 2019-09-23
Allow a port to be specified in KO_DOCKER_REPO.

ian-mi:
This allows repo names such as 'localhost:5000' to be used.


jonjohnsonjr:
Use `localhost:5000/foo` (foo can be basically anything).

The problem is that `localhost:5000` is ambiguous due to defaulting. It could represent either what you intend, or an image under the default namespace with a `5000` tag: `docker.io/library/localhost:5000`

PR #95, opened 2019-10-02
Support writing built images in OCI layout format to a file with the --oci-layout-path flag.

jbarrick-mesosphere:
This allows images built by ko to be used in Tekton. https://github.com/tektoncd/pipeline/blob/master/docs/resources.md#image-resource

### DIFF
--- a/pkg/commands/options/build.go
+++ b/pkg/commands/options/build.go
@@ -24,6 +24,7 @@ import (
 type BuildOptions struct {
 	ConcurrentBuilds     int
 	DisableOptimizations bool
+	OCILayoutPath    string
 }
 
 func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
@@ -31,4 +32,5 @@ func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
 		"The maximum number of concurrent builds")
 	cmd.Flags().BoolVar(&bo.DisableOptimizations, "disable-optimizations", bo.DisableOptimizations,
 		"Disable optimizations when building Go code. Useful when you want to interactively debug the created container.")
+	cmd.Flags().StringVarP(&bo.OCILayoutPath, "oci-layout-path", "", "", "Path to save the OCI image spec of the built images")
 }
--- a/pkg/commands/publish.go
+++ b/pkg/commands/publish.go
@@ -68,7 +68,7 @@ func addPublish(topLevel *cobra.Command) {
 			if err != nil {
 				log.Fatalf("error creating publisher: %v", err)
 			}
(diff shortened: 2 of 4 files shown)


jonjohnsonjr:
I'm a little torn on this one. This feels like it should be a separate tool, but it's also nice that this enables `ko` to produce an image without relying on the docker daemon or a registry.

From an implementation perspective, it might make more sense to implement this as a different [Publisher](https://github.com/google/ko/blob/3880b61d2d5f85401f726eda7012157295956360/pkg/publish/publish.go#L22-L28) that gets passed into publishImages. Right now this will both write to an OCI layout and publish to a registry, which might not be desirable.

jbarrick-mesosphere:
> I'm a little torn on this one. This feels like it should be a separate tool, but it's also nice that this enables ko to produce an image without relying on the docker daemon or a registry.

Yeah, I'm not sure about that either - it'd probably be possible to leverage something like skopeo for this task.

> From an implementation perspective, it might make more sense to implement this as a different Publisher that gets passed into publishImages. Right now this will both write to an OCI layout and publish to a registry, which might not be desirable.

It definitely makes sense for me to change this to use the Publisher interface - but, at least for my use-case, I actually want it to publish both to disk and registry. I'm not sure the best way to retain that - can we support multiple publishers at once?

jonjohnsonjr:
> I actually want it to publish both to disk and registry.

Doing this in a backwards compatible way would be somewhat annoying (`--local` and publishing to a registry are mutually exclusive ATM). We could add a `--push` flag that signals explicitly that the user wants to push to a registry, and have some logic where we default to the registry if no other publishers are selected. This would allow you to do `ko publish --oci-layout-path /foo/bar --push` to get both, but omitting `--push` would just write to the layout. WDYT?

It also occurs to me that this only really makes sense for `ko publish` -- for apply or resolve, we can't really deploy from an OCI image layout :/

> I'm not sure the best way to retain that - can we support multiple publishers at once?

We could reasonably add a MultiPublisher (a la [`io.MultiWriter`](https://golang.org/pkg/io/#MultiWriter)).

jonjohnsonjr:
> Yeah, I'm not sure about that either - it'd probably be possible to leverage something like skopeo for this task.

If skopeo doesn't support it, I've been meaning to add something like this to `crane` for a long time https://github.com/google/go-containerregistry/pull/406

jbarrick-mesosphere:
Thanks for the pointers! I'll circle back on this next week and figure out a way forward.

jonjohnsonjr:
@jbarrick-mesosphere if you have any interest in picking this back up, there's some overlap here with #128

PR #102, opened 2019-10-30
[Resolves #71] Add trimpath arg to gobuild

### LINKED ISSUE #71: Use -trimpath for go1.13
Currently, the binary built by `go build` includes the file paths for the source files. For most people, GOPATH is in their home directory, so their username is embedded in the resulting binary, which makes reproducible images difficult for us.

In 1.13, we can use `go build -trimpath` to fix this.

See https://github.com/golang/go/commit/74695644e0c38586a85b9b483830c546efbb6642

stanleynguyen:
I'm not sure about `-trimpath` behavior with pre-1.13 go versions though.
I'm expecting `-trimpath` to not do anything for pre-1.13 versions


jonjohnsonjr:
> I'm expecting -trimpath to not do anything for pre-1.13 versions

I think it actually complains if you pass a non-existent flag:

```
$ go build --foo bar
flag provided but not defined: -foo
```

You could use [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints) to only add `-trimpath` for `go1.13` and above. This isn't perfect, though. Since we shell out to the go command, someone could have a go binary built with 1.13 but have go 1.12 installed, in which case that would fail. We _could_ shell out to `go version` and do some string parsing, but I think that's actually worse.

stanleynguyen:
I think there're 2 appropriate strategies to solve this problem:
- with a version check vendor
```go
package main

import (
	"fmt"
	"runtime"
	"github.com/hashicorp/go-version"
)

func main() {
	fmt.Println(runtime.Version())
	v12, err := version.NewVersion("1.2")
	rtV, err := version.NewVersion(runtime.Version()[2:])
	if err != nil {
		fmt.Println(err)
	}
	if v12.LessThan(rtV) {
    		fmt.Printf("%s is less than %s", v12, rtV)	
	}
}
```
- using build tag and have 2 separate files for 2 different versions of the function `gbbuild_pre13.go` and `gbbuild_13.go` 

Should I go ahead with the latter

stanleynguyen:
This makes me wonder if we should include some test cases to cover the actual execution of the `ko` binary as the current test cases failed to flag `trimpath` even when tested with `go1.12` in Travis

jonjohnsonjr:
> This makes me wonder if we should include some test cases to cover the actual execution of the ko binary as the current test cases failed to flag trimpath even when tested with go1.12 in Travis

Yeah that would be hugely helpful. We could test ko `ko publish`ing itself to a local docker daemon in travis, which would get you most of the coverage. We could even use something like [kind](https://github.com/kubernetes-sigs/kind) or minikube to `ko apply` [this test](https://github.com/google/ko/tree/master/cmd/ko/test) and assert that the output is what we expected.

pkg/build/gbbuild_113.go
```diff
@@ -0,0 +1,68 @@
...
+	file := filepath.Join(tmpDir, "out")
+
+	args := make([]string, 0, 7)
+	args = append(args, "build")
+	if disableOptimizations {
+		// Disable optimizations (-N) and inlining (-l).
+		args = append(args, "-gcflags", "all=-N -l")
+	}
+	args = append(args, "-o", file)
+	args = append(args, "-trimpath")
```
jonjohnsonjr: I'd try to reduce the duplication here as much as possible across versions. Maybe define a separate function (`func addVersionSpecificBuildFlags([]string) []string` or something) that just adds the right flags so that the difference is just `return append(args, "-trimpath")`. This would make it easier to see the actual difference.

stanleynguyen:
@jonjohnsonjr I will include the test cases that covers what we've just discussed as well :+1:

stanleynguyen:
@jonjohnsonjr Added integration test using `kind` on travis env

jonjohnsonjr:
Wow this is really awesome, thanks! Much better than I would have done myself 😄

Issue #96, opened 2019-10-04
Allow using `go test -c` as build command


k15r:
We often use `go test -c` to create binaries that need to run in a cluster to run their tests inside kubernetes. Currently `ko` only supports `go build` as a build command, so for these tests there is currently no way to be applied using `ko`.

I'd be really glad if a flag would allow me to change the build command.

jonjohnsonjr:
@n3wscott had an related idea for a `ko test` command.

How would you want this to work? Similar to `ko apply` but instead of building a main package we create the test binary?

We might be able to do this without a flag, e.g. by checking for [`TestGoFiles`](https://godoc.org/go/build#Package.TestGoFiles) instead of `IsCommand` [here](https://github.com/google/ko/blob/3880b61d2d5f85401f726eda7012157295956360/pkg/build/gobuild.go#L126).

k15r:
i'm fine with any implementation, doesn't need to be a flag, it could also be a new command e.g. `ko test`. If it could figure out automatically which command is required it would allow for some nice scenarios. E.g. that would allow you to install a controller and run its tests with one `ko apply`

jonjohnsonjr:
> E.g. that would allow you to install a controller and run its tests with one ko apply

I think this would be hard to do if there are tests in the same `main` package, but if they were two separate packages, I could imagine something like:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: controller
spec:
  replicas: 1
  selector:
    matchLabels:
      app: controller
  template:
    metadata:
      labels:
        app: controller
    spec:
      containers:
      - name: controller
        image: github.com/google/ko/cmd/ko
---
apiVersion: batch/v1
kind: Job
metadata:
  name: controller-test
spec:
  template:
    spec:
      containers:
      - name: controller-test
        image: github.com/google/ko/cmd/ko/test
      restartPolicy: Never
```

Which would run `github.com/google/ko/cmd/ko` as the controller in a Deployment and `github.com/google/ko/cmd/ko/test` as the test in a Job. This would work with `ko apply`, but then you'd have to look at the Job status to determine if the test succeeds (I guess this isn't too hard?)

WDYT?

k15r:
Thats exactly what i would need! I like it!

jonjohnsonjr:
> We might be able to do this without a flag, e.g. by checking for TestGoFiles instead of IsCommand here.

I think you could just do this part and it would just work.

k15r:
[...]

PR #104, opened 2019-11-01
add support for build using `go test`

k15r:
as described in #96:
currently only `real` go builds are supported. With this PR importpaths that only contain tests will also be built using `go test -c`


k15r:
@googlebot I signed it!

k15r:
@googlebot I signed it!

pkg/build/gobuild_test.go
```diff
@@ -50,7 +50,7 @@ func TestGoBuildIsSupportedRef(t *testing.T) {
 
 	// Unsupported import paths.
 	for _, importpath := range []string{
-		filepath.FromSlash("github.com/google/ko/pkg/build"),       // not a command.
+		filepath.FromSlash("github.com/google/ko/pkg/commands/options"),       // not a command.
```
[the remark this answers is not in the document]
k15r: fixed

k15r:
i had to limit the 'test-builds' to strictly defined (`ko://<importpath>`) builds. otherwise the image field (and potentially other words) could break the yaml file. 
`image` as an import path gets resolved to image `Package image implements a basic 2-D image library.`. It has no main, but some tests :(

pkg/build/build.go
```diff
@@ -24,7 +24,7 @@ type Interface interface {
 	// IsSupportedReference determines whether the given reference is to an importpath reference
 	// that Ko supports building.
 	// TODO(mattmoor): Verify that some base repo: foo.io/bar can be suffixed with this reference and parsed.
-	IsSupportedReference(string) bool
+	IsSupportedReference(string, bool) bool
```
k15r: Instead of a flag we could think of not returning bool but more detailed information what kind of reference it is. 
eg. 
```
type ReferenceType string

const (
    MAIN ReferenceType = "main"
    TEST ReferenceType = "test"
    NONE ReferenceType = "none"
)
```

Or - in order to remove the logic from the caller - we could just allow the caller to specify what he needs 
e.g.
```
type AllowedReferences struct {
	Test bool
	Main bool
}

type Option func(*AllowedReferences)

func AllowTest() Option {
	return func(a *AllowedReferences) {
		a.Test = true
	}
}

func AllowMain() Option {
	return func(a *AllowedReferences) {
		a.Main = true
	}
}

func NewAllowedReferences(o ...Option) *AllowedReferences {
	ref := &AllowedReferences{}
	for _, a := range o {
		a(ref)
	}
	return ref
}
```
And add this to  `issupportedreference`

pkg/build/build.go
```diff
@@ -23,8 +23,10 @@ import (
 type Interface interface {
 	// IsSupportedReference determines whether the given reference is to an importpath reference
 	// that Ko supports building.
+	// `tests` activates 'test-builds'. So importpaths that do not provide `package main` but provide tests (*_test.go).
 	// TODO(mattmoor): Verify that some base repo: foo.io/bar can be suffixed with this reference and parsed.
-	IsSupportedReference(string) bool
+
+	IsSupportedReference(ip string, tests bool) bool
```
k15r: Instead of a flag we could think of not returning bool but more detailed information what kind of reference it is.
eg.

type ReferenceType string

const (
    MAIN ReferenceType = "main"
    TEST ReferenceType = "test"
    NONE ReferenceType = "none"
)
Or - in order to remove the logic from the caller - we could just allow the caller to specify what he needs
e.g.

type AllowedReferences struct {
	Test bool
	Main bool
}

type Option func(*AllowedReferences)

func AllowTest() Option {
	return func(a *AllowedReferences) {
		a.Test = true
	}
}

func AllowMain() Option {
	return func(a *AllowedReferences) {
		a.Main = true
	}
}

func NewAllowedReferences(o ...Option) *AllowedReferences {
	ref := &AllowedReferences{}
	for _, a := range o {
		a(ref)
	}
	return ref
}
And add this to issupportedreference

jonjohnsonjr:
Oof, this is frustrating, but it makes sense.

If we _have_ to change the signature of `IsSupportedReference`, I think I'd rather the bool represent whether we're using strict mode or not, rather than the type of package we're looking for...

We might be able to avoid changing the signature at all? If we don't strip the `ko://` prefix and just pass it down through to the builder, we could move the strict ref check into the go builder, and try to build tests if and only if it's a strict reference, the package is not main, and the package contains tests.

This means that more packages have to understand strict ref vs import path, but maybe that's worth it? Otherwise we're just passing these bools around anyway... 

cc @ImJasonH do you have any opinion on this?

k15r:
> We might be able to avoid changing the signature at all? If we don't strip the `ko://` prefix and just pass it down through to the builder, we could move the strict ref check into the go builder, and try to build tests if and only if it's a strict reference, the package is not main, and the package contains tests.
that would also allow different prefixes for different builders ...
e.g: `ko://` normal go build (even if if provides tests)
`ko-test://` test build (even if it provides a main)
`ko-<whatever>://` whatever people want to do differently
> cc @ImJasonH do you have any opinion on this?

jonjohnsonjr:
> that would also allow different prefixes for different builders ...

Yeah I think this is a pretty good model. We've discussed doing something like this before (I have a _really_ rough proof of concept [here](https://github.com/google/ko/compare/master...jonjohnsonjr:kopp) that supports Dockerfiles in addition to go packages via `ko-docker` and `ko-go` prefixes). I think we'll eventually want to split out builders into separate binaries so that it's more pluggable (a la [docker credential helpers](https://docs.docker.com/engine/reference/commandline/login/#credential-helpers)).

k15r:
imho implementing that change (splitting into multiple binaries) would go beyond the scope of this PR. 😄

jonjohnsonjr:
> imho implementing that change (splitting into multiple binaries) would go beyond the scope of this PR.

Oh yes, absolutely, I just wanted to mention it because I think that's a direction we (might) eventually head, and it's very close to you suggested :smile:

k15r:
So what's the way forward?

jonjohnsonjr:
Just got back from kubecon/vacation -- sorry for late reply.

I'm in favor of just passing through the strict prefix, i.e. what I suggested in https://github.com/google/ko/pull/104#issuecomment-551946810

Unless @ImJasonH has a strong opinion on it :)

jonjohnsonjr:
I've changed my mind again! (sorry)

Let's make this very very explicit and set up for a world where ko++ can build arbitrary things.

I think a prefix `ko-go-test://` makes this very obvious what it will do, and you clearly opt into it. I think ideally, that would dispatch to a separate implementation of `build.Interface` from the default go builder. The implementation would be almost identical, so we probably want to make [this](https://github.com/google/ko/blob/6aff039ca948315443c5bcde8a4c17a8513a75c7/pkg/build/gobuild.go#L159-L168) code something you can change with an option (that would clean up the debug stuff a bit too).

PR #107, opened 2019-11-07
Ignore null YAML documents when using a label selector.

Harwayne:
Resolving a YAML that ends in an empty YAML node.
```shell
go run ./cmd/ko resolve -l foo=bar -f - <<END
apiVersion: v1
kind: A
metadata:
  labels:
    foo: bar
---
END
```
Before this change:
```
error processing import paths in "-": error evaluating selector: yaml doesn't represent a k8s object
exit status 1
```

After this change:
```
apiVersion: v1
kind: A
metadata:
  labels:
    foo: bar

---

```

### DIFF
--- a/pkg/resolve/selector.go
+++ b/pkg/resolve/selector.go
@@ -47,6 +47,11 @@ func MatchesSelector(doc *yaml.Node, selector labels.Selector) (bool, error) {
 }
 
 func docKind(doc *yaml.Node) (string, error) {
+	// Null nodes will fail the check below, so simply ignore them.
+	if doc.Tag == "!!null" {
+		return "", nil
+	}
+
 	it := FromNode(doc).
 		Filter(Intersect(
 			WithKind(yaml.MappingNode),
--- a/pkg/commands/resolver_test.go
+++ b/pkg/commands/resolver_test.go
@@ -16,6 +16,7 @@ package commands
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"testing"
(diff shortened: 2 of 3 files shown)


pkg/resolve/selector.go
```diff
@@ -47,6 +47,11 @@ func MatchesSelector(doc *yaml.Node, selector labels.Selector) (bool, error) {
 }
 
 func docKind(doc *yaml.Node) (string, error) {
+	// Null nodes will fail the check below, so simply ignore them.
+	if doc.Tag == "!!null" {
+		return "", nil
+	}
+
```
dprotaso: I'd move this up a level to the `MatchesSelector` func. Then you can just return `false` there

Harwayne: I was thinking about that, but decided not to because `docKind` is called from both `MatchesSelector` and `listMatchesSelector`. I'm not sure why this would come up for `listMatchesSelector`, but wanted to be consistent just in case.

pkg/resolve/selector.go
```diff
@@ -47,6 +47,11 @@ func MatchesSelector(doc *yaml.Node, selector labels.Selector) (bool, error) {
 }
 
 func docKind(doc *yaml.Node) (string, error) {
+	// Null nodes will fail the check below, so simply ignore them.
+	if doc.Tag == "!!null" {
```
jonjohnsonjr: Where does `!!nulll` come from? Does that yaml package export something we can use instead? I would never be able to come up with this string on my own :P

Harwayne: I believe it is the canonical null type in YAML, similar to !!str or !!int. See https://yaml.org/type/null.html.

PR #120, opened 2020-01-16
Add kubectl Check for ko delete

danielhelfand:
This pull request adds an error message in case `kubectl` is not installed. I'm assuming most users of `ko` will have `kubectl` installed, but it's nice to have a clear message in the edge case when it's not available.

### DIFF
--- a/pkg/commands/delete.go
+++ b/pkg/commands/delete.go
@@ -15,10 +15,11 @@
 package commands
 
 import (
-	"github.com/spf13/cobra"
 	"log"
 	"os"
 	"os/exec"
+
+	"github.com/spf13/cobra"
 )
 
 // runCmd is suitable for use with cobra.Command's Run field.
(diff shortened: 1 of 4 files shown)


pkg/commands/delete.go
```diff
@@ -46,6 +52,15 @@ func passthru(command string) runCmd {
 	}
 }
 
+// check if kubectl is installed
+func isKubectlAvailable() bool {
+	cmd := exec.Command("kubectl")
```
jonjohnsonjr: You could use [`exec.LookPath`](https://golang.org/pkg/os/exec/#LookPath) for this, and we probably want to do this for `apply`, `create`, and `run` as well.

danielhelfand: @jonjohnsonjr Awesome! I can definitely use `LookPath` instead. Thanks for the better suggestion. Would you recommend a good place to define `isKubectlAvailable()` so I can use it across all commands?

jonjohnsonjr: Anywhere is fine! I'd just leave it here, probably.

danielhelfand: I put in `commands.go` for now, but more than willing to move it back to any of the commands that use it.

PR #121, opened 2020-01-24
Add goreleaser GitHub action

jonjohnsonjr:
Based on https://goreleaser.com/actions/

This seems to work in testing against my fork. It will automatically publish a release on GitHub releases whenever we add a tag.


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
```
ImJasonH: These `-`s on their own line might be valid YAML but they look weeeird

jonjohnsonjr: Agreed

PR #134, opened 2020-02-13
Add additional output formats (tarball and layout)

### LINKED ISSUE #128: Tarball output format?
It would be really useful if ko could output images in the tarball format, similar to what you get with `docker save`. Ideally, ko could output a release.yaml, but output image tarballs rather than publishing them (perhaps writing the tarball to disk?)

Many other tools that build images support outputting `docker save`-format tarballs; for instance:
* [bazel rules_docker](https://github.com/bazelbuild/rules_docker#container_bundle-1)
* [buildkit](https://github.com/moby/buildkit#docker-tarball)
* [Kaniko](https://github.com/GoogleContainerTools/kaniko#--tarpath)
* [img](https://github.com/genuinetools/img#export-an-image-to-docker)

jonjohnsonjr:
Fixes #128 
Closes #95

This is staged in a few commits that should be easy to review independently:

* Create a MultiPublisher
* Add publish.{Tarball,Layout}Publisher
* Consolidate publisher options
* Add flags for tarball/layout publishers


jonjohnsonjr:
@bioball @josephlewis42 @jbarrick-mesosphere 

Would this work for all of your needs?

bioball:
Awesome!

I just tried it, and it doesn't seem like all the images within `ko resolve` are getting added to the tarball. This is what I see when building Tekton:

```
ko resolve --tarball=output.tar -f config > release.yaml
```

jonjohnsonjr:
>  doesn't seem like all the images within ko resolve are getting added to the tarball

Ahh yeah I forgot to go back and fix this, thanks. There's not an easy way to append an image to a tarball right now (since the interface supports streaming writes). Hacked a workaround in to just buffer images until the end and write them out in a Close().

bioball:
Looks like it works!

josephlewis42:
This looks like it will fit our purposes nicely!

Issue #152, opened 2020-04-29
vendor paths are broken with go modules


mattmoor:
_We've been talking about this a bit in Knative slack in #ko, but I wanted to track this here in an issue, which is more durable and searchable._

The gist is that there are a number of projects that vendor a binary and then embed the following in yaml to publish the vendored binary:  `github.com/foo/bar/vendor/github.com/baz/blah`

I was under the mistaken impression that this worked fine because Tekton uses this for the `gcs-fetcher` image, but it turns out that isn't actually building from the local vendor tree, but the "latest" release of Tekton.  We caught this on the Knative side because we're still tagging our releases as "pre-release" so Github's "latest" release voodoo doesn't mask the issue for us like it did for Tekton.

For Tekton, the following works for the same reason:
```
ko publish github.com/tektoncd/pipeline/vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher
```

However, a similar command in Knative Serving fails (due to the "no latest release" problem):
```
ko publish knative.dev/serving/vendor/github.com/tsenart/vegeta
```

Interestingly, when I adapt these both to relative paths, they BOTH fail, but for a different reason!
```
ko publish ./vendor/github.com/GoogleCloudPlatform/cloud-builders/gcs-fetcher/cmd/gcs-fetcher
ko publish ./vendor/github.com/tsenart/vegeta
```

In this case they fail because `qualifyLocalImport` (due to `packages.Load`) returns the unqualified import path (e.g. `github.com/baz/blah`) , but prior to go modules it returned the fully-qualified path (e.g. `github.com/foo/bar/vendor/github.com/baz/blah`).

The nearest workaround I've been able to find thus far is to drop the condition around this call:
https://github.com/google/ko/blob/f45bc13ded83af4cfa337910d4ba892015f5d1dc/pkg/build/gobuild.go#L142-L150

... which would let us reference `ko://github.com/baz/blah` from within `github.com/foo/bar` and (I believe) have it respect the version of that dependency from `go.mod` 🤞 

The main thing that gives me pause here is that the PR (https://github.com/google/ko/pull/60) that introduced go mod support referenced this as a pretty important performance optimization 😬

I'm hoping that either @ImJasonH or @jonjohnsonjr has some insights here.

mattmoor:
Not that for Serving, I have been poking at this in the context of: https://github.com/knative/serving/pull/7734, which as of now has not yet landed.

[...]

PR #154, opened 2020-04-30
[modules] Dependent command packages can now be built with ko 

dprotaso:
Thus if you have a (in)direct command package as a dependency
say `myhost.com/go/package/cmd/run` you can now publish this
with the following ko command

```bash
$  ko publish myhost.com/go/package/cmd/run
```

This follows go's module/vendor semantics. Thus go1.14 by default will use vendored modules. With go1.13 you need the explicit flags to use vendored deps

```bash
$  GOFLAGS="-mod=vendor" ko publish myhost.com/go/package/cmd/run
```

Subsequently in yaml files you can drop the superfluous prefixes

ie. this (in a non-module world)  

```yaml
apiVersion: batch/v1
kind: Job
spec:
  containers:
  - name: vegeta
    image: knative.dev/serving/vendor/github.com/tsenart/vegeta
```
becomes this in the module world

```yaml
apiVersion: batch/v1
kind: Job
spec:
  containers:
  - name: vegeta
    image: github.com/tsenart/vegeta
```


dprotaso:
cc https://github.com/google/ko/issues/152

pkg/build/gobuild.go
```diff
@@ -79,34 +79,73 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
...
 //
 // Related: https://github.com/golang/go/issues/26504
-func moduleInfo() *modInfo {
-	output, err := exec.Command("go", "list", "-mod=readonly", "-m", "-json").Output()
+func moduleInfo() (*modules, error) {
+	modules := modules{
+		deps: make(map[string]*modInfo),
+	}
+
+	// TGDO - optimize
```
ImJasonH, line 102: What's this?

ImJasonH:
I'm not sure I understand what this is trying to achieve. Is the effect that you can `ko build` a non-`package main` package?

pkg/build/gobuild.go
dprotaso: see updated comment

dprotaso:
whoops needs a rebase - one minute

dprotaso:
Failure in CI is go1.12 

> pkg/build/gobuild.go:134: Errorf format %w has unknown verb w

1.12 is EOL so I made https://github.com/google/ko/pull/156

mattmoor:
@ImJasonH If I'm reading the description correctly, this is the `ko publish` complement of my change.

pkg/build/gobuild.go
```diff
@@ -75,38 +81,80 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
...
+		err := dec.Decode(&info)
+		if err == io.EOF {
+			// all done
+			break
+		}
+
+		modules.deps[info.Path] = &info
+
+		if info.Main {
+			modules.main = &info
```
mattmoor, line 130: couldn't we hit N of these?  Which one wins?

mattmoor:
Ok, so reading through the change, I want to make sure I understand...

With my change, a vendored main module `foo` will work if it is written as `ko://foo`, but what you are doing is dumping all of the referenced modules, so that just `foo` still works.

Since in `ko publish` qualify local import turns `./vendor/foo` into `foo` with modules, this unblocks letting `ko publish ./vendor/foo` or `ko publish foo` from working.

@dprotaso is that right?

pkg/build/gobuild.go
dprotaso: I believe there's only one

dprotaso:
> this unblocks letting ko publish ./vendor/foo or ko publish foo from working.

This unblocks `ko publish foo`
This _does not unblock_ `ko publish ./vendor/foo`

pkg/build/gobuild.go
dprotaso: main != `package main` but specifically "main module" = root module

mattmoor:
I wonder, would simply having `qualifyLocalImport` slap a `ko://` on the front of what it produces work?

pkg/build/gobuild.go
mattmoor: that's not confusing *at all* 🙄

dprotaso:
🤷‍♂️

mattmoor:
This just makes me want to deprecated non-`ko://` faster, but it seems like an improvement of the status quo.  thanks for chasing this down.

PR #159, opened 2020-05-01
[modules] support references that point to modules in the vendored folder

dprotaso:
This should unblock using references like:

   `knative.dev/serving/vendor/github.com/tsenart/vegeta`
  `./vendor/github.com/tsenart/vegeta`

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -461,7 +461,7 @@ func (gb *gobuild) Build(ctx context.Context, s string) (v1.Image, error) {
 	}
 
 	// Do the build into a temporary file.
-	file, err := gb.build(ctx, pkg.ImportPath, platform, gb.disableOptimizations)
+	file, err := gb.build(ctx, pkg.Dir, platform, gb.disableOptimizations)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -200,8 +200,21 @@ func (g *gobuild) importPackage(ref reference) (*gb.Package, error) {
 	// * path is a module
 
 	_, isDep := g.mod.deps[ref.Path()]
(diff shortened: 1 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -444,7 +461,7 @@ func (gb *gobuild) Build(ctx context.Context, s string) (v1.Image, error) {
 	}
 
 	// Do the build into a temporary file.
-	file, err := gb.build(ctx, ref.Path(), platform, gb.disableOptimizations)
+	file, err := gb.build(ctx, pkg.Dir, platform, gb.disableOptimizations)
```
dprotaso, line 464: curious if this triggers the determination to fail

dprotaso:
What's confuses me is if you have different versions of a package vendored - I'm unclear what go build picks up

pkg/build/gobuild.go
dprotaso: switched to import path

dprotaso:
Supplying the `pkg.Dir` means we're passing absolute path to `go build`

ie.
```
/usr/local/bin/go build -o /var/folders/qn/bxc0sm8j5dgcx260_4r3vr7w0000gn/T/ko453017338/out -trimpath /Users/dprotasowski/serving/vendor/github.com/tsenart/vegeta
```

The determinism test is actually just capturing call args and dumping them into the image - so it actually may not be representative of what's happening in reality

pkg/build/gobuild.go
dprotaso: switch back to pkg dir and changed the test

mattmoor: I'm guessing this is needed because it piggybacks on the normalization being done in importPackage, but this is the part of the change that makes me most nervous 🤔

dprotaso: this is passing the absolute path to the package now

dprotaso:
closing - given we'll be using stricts refs in the future

Issue #171, opened 2020-06-01
Integration tests for go module support


Cynocracy:
ko, when run from within the GOPATH, uses GOPATH mode for resolving images even when environment variables instruct otherwise.

ko version 0.5.1

Cynocracy:
One caveat here: I have 2 machines with this setup and they're behaving differently :( One seems to do fine with the module type ko:// references to vendor. The other does not. I'll have to take a look and see if ko has debugging tips anywhere, I'd love to figure out this mystery :)

Cynocracy:
Thoughts on changing to using https://golang.org/pkg/go/build/ from exec to see if that helps resolve inconsistencies?

Cynocracy:
In slack I got a "sure, why not?" So I'm going to play around with that.

/assign Cynocracy

Cynocracy:
Oh, this repo isn't as fancy. I'll do it myself :P

Cynocracy:
No luck, but please consider it assigned.

Cynocracy:
This /may/ have been due to a bug in our setup, but we're investigating a bit more now that we got our internal repo into a good state. That should help us understand a bit better why this is occurring (and play around with fixes :)

Cynocracy:
OK, here's my attempt at repro.

Setup
1. Add a tools.go file to ko's root, and import 'github.com/go-training/helloworld'
2. `go mod vendor && go mod tidy`
3. Modify vendor to change message output
4. Make sure you are in the GOPATH (where you might fallback to GOPATH mode) and you run `go get -d githu`b.com/go-training/helloworld`

Scenarios:
a. GO111MODULE=on GOFLAGS="-mod=vendor" should always output the changed text
b. GO111MODULE=auto GOFLAGS="-mod=vendor" should output the changed text iff in the module path, otherwise the original text.
i.  In the module root
ii. Outside module root
c. GO111MODULE=off should output the original text.
d. GO111MODULE=on should fail outside a module.

Results:
a. GO111MODULE="on" GOFLAGS="-mod=vendor" ko publish github.com/go-training/helloworld

docker run result
Hello World TEST!!

b. 
i.
Hello World TEST!!
  
ii.
Hello World!!

c. 
Hello World!!

d.
go: cannot find main module; see 'go help modules'


So: WAI :+1: Most likely the issues I was seeing were related to our own infra around ko and not ko itself.

Any objections to me adding something similar as an integration test?

jonjohnsonjr:
[...]

PR #179, opened 2020-07-27
Add integration tests for various go mod corner cases.

Cynocracy:
"Fixes" https://github.com/google/ko/issues/171 (where I was confused by the behavior of modules) by ensuring ko behaves similarly to go build in various edge cases.


hack/tools.go
```diff
@@ -0,0 +1,6 @@
+package hack
+
+import (
+  "github.com/go-training/helloworld"
```
jonjohnsonjr, line 4: From travis:

> hack/tools.go:4:3: import "github.com/go-training/helloworld" is a program, not an importable package

I think you need a `_` before the import (and a `gofmt`).

Cynocracy: Thanks :)

Re: whether we should keep this behavior; totally agree, ideally we would replicate the behavior of go build exactly, which might be a moving target.

Cynocracy:
Oh, because it's not actually obvious: I wrote the test cases before checking behavior :) So at least *I'm* convinced this behavior is correct.

jonjohnsonjr:
Same error :/

I think you need a build tag that doesn't get used so that it doesn't try to build it, we did something similar in go-containerregistry: https://github.com/google/go-containerregistry/pull/599/files#diff-39bad9c50a0f30bd6c067b83431bec76R1

Cynocracy:
Ah, I missed that I didn't have the tools build tag, yeah that's definitely necessary.

Cynocracy:
Any suggestions on what to set for KO_DOCKER_REPO ?

jonjohnsonjr:
> Any suggestions on what to set for KO_DOCKER_REPO ?

I think the current integration tests us the `--local` flag to just write it into the docker daemon, which doesn't require KO_DOCKER_REPO to be set, if you wanna do that.

Cynocracy:
Ah, perfect, that works for me, thank you :)

Cynocracy:
Ready! I think :)

PR #180, opened 2020-08-10
Allow images to be loaded into kind using 'kind.local'.

### LINKED ISSUE #123: Document how to use ko with kind
Feedback from tgik.io/102 that we should explain how to use `ko` with [kind](https://github.com/kubernetes-sigs/kind).

### LINKED ISSUE #149: Add `pkg/v1/kind` for writing images to KinD
I think we want something like:
https://github.com/google/go-containerregistry/tree/master/pkg/v1/daemon

... but to write it directly to KinD.

cc @BenTheElder for pointers on the best API.

I'd love to get some mechanism like this so that we can support something like: `KO_DOCKER_REPO=kind.local`

markusthoemmes:
Fixes #123 
Fixes #149 

This allows `ko` to load images into kind nodes when the user is setting the `KO_DOCKER_REPO` to `kind.local`.

This replicates some of `kind`'s CLI in that it talks to the container hosting the respective K8s node and loads an image into it by `exec`ing into the containerd instance of the respective container and loading the image in via stdin.


pkg/publish/kind.go
```diff
@@ -0,0 +1,86 @@
...
+			return nil, err
+		}
+		log.Printf("Added tag %v", tagName)
+	}
+
+	return &digestTag, nil
+}
+
+func (t *kindPublisher) Close() error {
+	return nil
```
jonjohnsonjr, line 85: If there's some benefit from bundling images together (from your slack thread), you could buffer everything and write them all out here, similar to what i did for tarball: https://github.com/google/ko/blob/ecce857be69db48d914837da0d234c54d7f1a419/pkg/publish/tarball.go#L88-L97

markusthoemmes: I opted for not doing that as it puts publishing images "after the fact" in the `ko apply` case, which ain't great.

With the reworked approach (vendoring kind and talking to docker directly) this seemed fine performance wise, especially with the tagging optimization.

jonjohnsonjr: Oh, of course -- my suggestion doesn't make much sense :)

pkg/publish/kind/write.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+
+	"sigs.k8s.io/kind/pkg/cluster"
+	"sigs.k8s.io/kind/pkg/cluster/nodes"
+)
+
+// Tag adds a tag to an already existent image.
+func Tag(src, dest name.Tag) error {
+	return onEachNode(func(n nodes.Node) error {
+		cmd := n.Command("ctr", "--namespace=k8s.io", "images", "tag", "--force", src.String(), dest.String())
```
jonjohnsonjr, line 32: I'm not super familiar with kind -- is this namespace flag somethign that would need to be configurable? Similar question re: `cluster.DefaultName`on line 76.

I understand this is a PoC, just curious how much config stuff we'll need to plumb through here for a final version.

pkg/publish/kind.go
markusthoemmes: It does, one intermediate stage had that code even. Then I realize the caveat. I suppose, as a middle-ground, we could try to close the publisher earlier instead of deferring it. But maybe a followup.

pkg/publish/kind/write.go
markusthoemmes: This namespace is hardcoded in kind too.

In theory, `kind` can have different cluster names and you could also potentially choose which nodes you load images on. I opted for the default cluster name here and opted to just publish to all nodes.

pkg/publish/kind/write.go
```diff
@@ -0,0 +1,85 @@
...
+// onEachNode executes the given function on each node. Exits on first error.
+func onEachNode(f func(nodes.Node) error) error {
+	nodeList, err := getNodes()
+	if err != nil {
+		return err
+	}
+
+	for _, n := range nodeList {
+		if err := f(n); err != nil {
+			return err
```
ImJasonH, line 65: Maybe include a node identifier in this error message? Or does the error include details about the node that encountered the error?

pkg/publish/kind/write.go
```diff
@@ -0,0 +1,113 @@
...
+func Write(tag name.Tag, img v1.Image) error {
+	return onEachNode(func(n nodes.Node) error {
+		pr, pw := io.Pipe()
+
+		grp := errgroup.Group{}
+		grp.Go(func() error {
+			return pw.CloseWithError(tarball.Write(tag, img, pw))
+		})
+
+		cmd := n.Command("ctr", "--namespace=k8s.io", "images", "import", "-").SetStdin(pr)
```
mattmoor, line 66: these are going to run serially.  I wonder about running `onEachNode` in an `errgroup`?

mattmoor: Seems like complexity for a follow-up

mattmoor (approved):

cmd/ko/test/main.go
```diff
@@ -35,5 +37,10 @@ func main() {
 	if err != nil {
 		log.Fatalf("Error reading %q: %v", file, err)
 	}
-	log.Print(string(bytes))
+	log.Printf(string(bytes))
+
+	// Cause the pod to "hang" to allow us to check for a readiness state.
+	sigs := make(chan os.Signal)
+	signal.Notify(sigs, syscall.SIGTERM)
```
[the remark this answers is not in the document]
mattmoor: looking here, it looks like TERM is wrong on windows and we want Interrupt (for both): https://github.com/knative/pkg/blob/master/signals/signal_windows.go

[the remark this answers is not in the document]
evankanderson: What setup do I need to run this test? I have Windows and I think I have Kind on there (I've been using the Docker k8s cluster recently, but have tried several).

pkg/publish/kind/write.go
markusthoemmes: ack, I thought about that too but then wanted to get the general mechanism blessed first.

cmd/ko/test/main.go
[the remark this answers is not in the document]
markusthoemmes: @evankanderson kind should be fine. Just running `integration-tests.sh` should do it.

pkg/publish/kind/write.go
mattmoor: Yep, this seems great.  We should add `kind.local` to the list of built-in domains that Serving ignores for digest resolution too 😎

mattmoor:
@markusthoemmes Can you update the README to call this out?  Generally this seems great.

pkg/publish/kind/write.go
markusthoemmes: I'll do that as soon as this lands :joy:

mattmoor (approved):

cmd/ko/test/main.go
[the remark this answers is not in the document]
evankanderson: It turns out that `.\integration-tests.sh` doesn't work at all unless you install WSL. 😁 

I'm interpreting that into windows-ese now, as running in WSL2 isn't really what I think you meant.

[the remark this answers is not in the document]
evankanderson: Kubectl logs output:
```
kubectl logs kodata
2020/09/04 23:00:49 ../kenobi
2020/09/04 23:00:49 Error reading "/var/run/ko/refs/heads/master": open /var/run/ko/refs/heads/master: not a directory
```

PR #188, opened 2020-09-04
Pin k8s.io/api for good measure

mattmoor:


mattmoor:
I was comparing `vendor/k8s.io` with what we have `replace` directives for, and everything else was present or doesn't have a `0.18.8` tag (one had a `release-1.18` branch, but no tag 😞 )

mattmoor:
Alright, should be rfal 🤞

jonjohnsonjr (approved):

PR #189, opened 2020-09-04
Fix k8s.io/cli-runtime "require"

mattmoor:
The "require" is behind the version in "replace", and since cli-runtime references the package mentioned in the errors we are seeing, rule this out as a source!


vagababov (commented):
/lgtm

PR #197, opened 2020-09-25
Add a handful of files to trigger more analyses.

mattmoor:
`.codecov.yaml` - Enable codecov.io coverage reports
`.golangci.yaml` - Enable the enumerated checks in golangci-lint
`boilerplate.*.txt` - Enable file header boilerplate checks for these extensions

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,41 @@
+run:
+  timeout: 5m
+
+  build-tags:
+    - e2e
+    - performance
+    - probe
+    - preupgrade
+    - postupgrade
+    - postdowngrade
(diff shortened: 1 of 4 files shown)


jonjohnsonjr (approved):

.golangci.yaml
```diff
@@ -0,0 +1,41 @@
+run:
+  timeout: 5m
+
+  build-tags:
```
jonjohnsonjr, line 4: these seems largely knative-y

mattmoor: That's b/c I just copied the strictest I was aware of from there 😉

PR #209, opened 2020-09-30
feat(config): search parent dirs recursively for config file

### LINKED ISSUE #172: Support .ko.yaml while in subdir
Today if I invoke `ko apply -f foo/` while in a subdirectory `bar/` then it is possible that a `.ko.yaml` file that's a sibling of `bar/` won't be picked up because we (currently) only look at siblings of `foo/`.

Fortunately, `AddConfigPath` seems to support multiple invocations, so we can crawl the directory hierarchy from the current working directory to populate the path.

I think we want this here: https://github.com/google/ko/blob/d45c5277500220120e9196f431856c0cdd83e92b/pkg/commands/config.go#L104

stanleynguyen:
Resolves #172

### DIFF
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -21,6 +21,8 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"syscall"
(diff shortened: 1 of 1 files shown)


stanleynguyen:
Hi @mattmoor are we still considering this feature?

PR #216, opened 2020-10-05
Don't panic when MultiPublisher is given zero publishers

### LINKED ISSUE #215: ko resolve segfaults when push=false
When push=false there are no publishers and multiPublisher.Publish will return a nil name.Reference causing a segfault:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x1610af5]

goroutine 13 [running]:
github.com/google/ko/pkg/resolve.ImageReferences.func1(0x1, 0x665a99)
        /usr/local/google/home/ianmi/go/pkg/mod/github.com/google/ko@v0.6.0/pkg/resolve/resolve.go:66 +0x115
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc000362a80, 0xc0000b6230)
        /usr/local/google/home/ianmi/go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup.go:57 +0x59
created by golang.org/x/sync/errgroup.(*Group).Go
        /usr/local/google/home/ianmi/go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup.go:54 +0x66

ImJasonH:
Fixes #215 

cc @ian-mi

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -160,6 +160,16 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 			}
 			publishers = append(publishers, dp)
 		}
+
+		// If not publishing, at least generate a digest to simulate
+		// publishing.
+		if len(publishers) == 0 {
+			publishers = append(publishers, nopPublisher{
+				repoName: repoName,
+				namer:    namer,
+			})
+		}
+
 		return publish.MultiPublisher(publishers...), nil
 	}()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


jonjohnsonjr:
> This ignores -P or -B flags, and always uses -P behavior.

Hm... what if we [always append the default publisher](https://github.com/google/ko/blob/dc30efcfd8b6f6a8aa381c715390013f9c1fa23b/pkg/commands/resolver.go#L151) and just added `push` as a default publisher Option and guarded all this with a boolean:

https://github.com/google/ko/blob/dc30efcfd8b6f6a8aa381c715390013f9c1fa23b/pkg/publish/default.go#L128-L151

Then that keeps everything in sync and we don't have to worry about naming logic drift, plus we can just return an error from multipublisher `if len(publishers) == 0` and it's pretty easy to follow the logic.

jonjohnsonjr (approved):

jonjohnsonjr:
```
~/go/src/github.com/google/ko ((b890235...)) $ ko publish -B --push=false ./cmd/ko/
2020/10/06 09:44:44 Using base gcr.io/distroless/static:nonroot for github.com/google/ko/cmd/ko
2020/10/06 09:44:46 Building github.com/google/ko/cmd/ko for linux/amd64
gcr.io/jonjohnson-test/ko/ko@sha256:77582412aa5b7919ff07951ebdac05fd145463c802ddf254bca3065292c3059f
```

Dope.

PR #222, opened 2020-10-13
Fix things pointed out by golangci-lint

markusthoemmes:
This fixes the linting issues as pointed out by `golangci-lint`. Also reduces its config a bit as most of it isn't necessary here.


ImJasonH (approved):
Thanks for fixing these! 🎉

Issue #224, opened 2020-10-21
test ko publish multi-arch


mattmoor:
Our ability to test `ko apply` multi-arch is limited by support for KinD multi-arch images.

This doesn't stop us from testing images produced via `ko publish`.  It'd be good to leverage this and Travis' broad multi-arch support to validate the `ko` multi-arch images are WAI.

The default base not supports: amd64, arm64, s390x, and PowerPC.  I think Travis has limited support for all of these.

mattmoor:
cc @jonjohnsonjr @ImJasonH

PR #227, opened 2020-10-23
Run the hello test on multiple architectures

mattmoor:
Fixes: https://github.com/google/ko/issues/224

### DIFF
--- a/pkg/publish/daemon.go
+++ b/pkg/publish/daemon.go
@@ -17,6 +17,7 @@ package publish
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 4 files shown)


pkg/publish/daemon.go
```diff
@@ -47,9 +48,43 @@ func (d *demon) Publish(br build.Result, s string) (name.Reference, error) {
...
-	// There's no way to write an index to a daemon, so attempt to downcast it to an image.
-	img, ok := br.(v1.Image)
-	if !ok {
+	// There's no way to write an index to a kind, so attempt to downcast it to an image.
+	var img v1.Image
+	switch i := br.(type) {
+	case v1.Image:
+		img = i
+	case v1.ImageIndex:
+		im, err := i.IndexManifest()
```
ImJasonH, line 57: Does this seem like a helper we should add to ggcr? `idx.ImageByArch(os,arch)`?

mattmoor: Yeah probably.  I copied this from kind.

I think I left a similar comment on @jonjohnsonjr's original PR.

ImJasonH: fyi https://github.com/google/go-containerregistry/pull/797

ImJasonH (approved):

PR #231, opened 2020-10-28
Tag contintuent images in manifest lists with os+arch

### LINKED ISSUE #230: When pushing manifest lists, tag constituent images with arch+os
It could be useful to make, e.g., `gcr.io/foo/github.com/bar/baz/cmd/quux:linux-arm64` available, and it shouldn't be much more work/latency for ko to support it.

imjasonh:
Fixes #230 

When running `ko publish --platform=all`, each resulting image will now be tagged with the image's OS + architecture, e.g., `:latest-linux-amd64` or `:v1.2.3-linux-arm64`

### DIFF
--- a/pkg/publish/default.go
+++ b/pkg/publish/default.go
@@ -143,7 +143,7 @@ func tagImages(idx v1.ImageIndex, tag name.Tag, opt []remote.Option) error {
 			return err
 		}
 
-		log.Printf("Tagging %v", itag)
+		log.Print("Tagging ", itag)
 		if err := remote.Tag(itag, img, opt...); err != nil {
 			return err
 		}
@@ -170,12 +170,12 @@ func (d *defalt) Publish(br build.Result, s string) (name.Reference, error) {
 		}
 
 		if i == 0 {
-			log.Printf("Publishing %v", tag)
+			log.Print("Publishing ", tag)
 			if err := pushResult(tag, br, ro); err != nil {
 				return nil, err
 			}
 		} else {
-			log.Printf("Tagging %v", tag)
+			log.Print("Tagging ", tag)
 			if err := remote.Tag(tag, br, ro...); err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
I'm kind of afraid that this would be surprising behavior and accidentally overwrite things that aren't intended. This is expected for `latest`, but other tags have me worried.

ImJasonH:
> I'm kind of afraid that this would be surprising behavior and accidentally overwrite things that aren't intended. This is expected for `latest`, but other tags have me worried.

So the concern would be that users already have an image tagged `:v1.2.3-linux-arm64` which they care about, then they run `ko publish --platform=all --tags=v1.2.3` where the base image happens to support linux/arm64, and `ko` surprisingly tags over their image.

If that scenario seems plausible and negatively surprising to you I'm okay dropping this PR. I'd rather not add a flag for it.

This PR should include a docs change at least, that would help with the surprise.

jonjohnsonjr:
> I'd rather not add a flag for it.

I just don't love this as the default behavior. It's not completely obvious to me that this is something I'd want. Also, as is, this PR  has a collision for platforms with variants, e.g.:

```
$ crane manifest alpine | jq .
2020/11/19 12:34:36 No matching credentials were found, falling back on anonymous
{
  "manifests": [
    {
      "digest": "sha256:d7342993700f8cd7aba8496c2d0e57be0666e80b4c441925fc6f9361fa81d10e",
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      },
      "size": 528
    },
    {
      "digest": "sha256:c4f0f03cda416f3e4cfebcfea9910463121651b019c6677053ece71084699f47",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v6"
      },
      "size": 528
    },
    {
      "digest": "sha256:d0f78a6ddf7a457dc72dbd44eab67209454ddb1e6d2323fa8e27275bc13dc320",
      "platform": {
        "architecture": "arm",
        "os": "linux",
        "variant": "v7"
      },
      "size": 528
    },
...
```

PR #233, opened 2020-10-31
IsSupportedReference returns descriptive error

ImJasonH:
This can be useful to determine what they need to do to make a ko
publish work.


mattmoor (approved):
I've wanted this forever, but without strict it was prohibitively noisy.  Sooooo glad to see this

Issue #143, opened 2020-03-13
korc file


mattmoor:
It would be useful to be able to have a `.korc` file akin to a `.bazelrc` file that allows for certain flags to be passed by default.

For example, passing `-B` or `-P` gets tedious, and if I exclusively use a registry that supports deep nesting like GCR, I may want `-P` to avoid the noisy hash-based names we use by default.  For a registry like ECR, where I have to create repositories manually before pushing, passing `-B` all the time may be necessary for the push to succeed.  Another good example of this might be if I want to always pass `--strict`.

Initially I'd though `KO_FLAGS=-B`, as we do in our e2e tests, but some of these flags are context-specific (e.g. `-B` won't work with `ko delete`).

Bazel's solution to this is to allow a file in your home directory (and a few other contexts) where you can specify per-subcommand flags, e.g.

```
apply -B --strict
# For releases, use full paths
resolve -P --strict
delete --ignore-not-found
```

Wanted to put this out into the world in case someone was sufficiently motivated to implement it 😇

jonjohnsonjr:
Yeah I'd love this and I'd love to deprecate the equivalent flags. Naming convention is per user not per project.

PR #234, opened 2020-11-02
Refactor the publish.Namer, add `--bare` option for image naming

mattmoor:
In some environments (e.g. ECR) each repository needs to be pre-created, which gets tedious.

Given that `ko` does everything by digest, this collapses all of the images published into the same repository: `${KO_DOCKER_REPO}` with no suffix to discriminate between images.

This is also beneficial from a security standpoint in certain contexts, as `KO_DOCKER_REPO=docker.io/{user}` might publish new repositories as public by default.

### DIFF
--- a/pkg/commands/options/publish.go
+++ b/pkg/commands/options/publish.go
@@ -41,6 +41,8 @@ type PublishOptions struct {
 	PreserveImportPaths bool
 	// BaseImportPaths uses the base path without MD5 hash after KO_DOCKER_REPO.
 	BaseImportPaths bool
+	// Naked uses a tag on the KO_DOCKER_REPO without anything additional.
+	Naked bool
 }
 
 func AddPublishArg(cmd *cobra.Command, po *PublishOptions) {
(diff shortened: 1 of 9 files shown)


mattmoor:
For example with ECR, my registry domain is `912237561242.dkr.ecr.us-west-2.amazonaws.com`, and I have to create each of: `912237561242.dkr.ecr.us-west-2.amazonaws.com/{foo}` today.

With this, I can use `KO_DOCKER_REPO=912237561242.dkr.ecr.us-west-2.amazonaws.com/ko-images` and `ko apply --naked` will publish all images to the above repository and substitute their digests into the yamls, e.g. `912237561242.dkr.ecr.us-west-2.amazonaws.com/ko-images@sha256:deadbeef`

In these cases, no part of the import path is incorporated into the repository name, which is where the name "naked" comes from.  I'm open to alternative naming suggestions, but struggled to find a more descriptive term than this.

jonjohnsonjr:
My only real issue with this is having a global flag called `--naked`. I wish we had a separate mechanism for selecting a namer that wasn't flag-based, ref:

https://github.com/google/ko/issues/172
https://github.com/google/ko/issues/143

mattmoor:
@ImJasonH also doesn't like the name either.  He suggested `--bare`, which I think bears the same meaning as `--naked`, but without the same (level of) vulgar connotation 🤷 

I am happy to switch to `--bare`, though I'd probably drop the short flag since we already have `-B`.

jonjohnsonjr:
I'd prefer dropping the short flag regardless, since they're such a precious resource. In fact, it would be great if this was really cumbersome to type so that you got annoyed into implementing a better mechanism for this :) 

Some dumb ideas that might inspire better ideas:
* `--flat`
* `--direct`
* `--nosuffix`
* `--norepo[sitory]`
* `--noname`
* `--nemo`

In this mode, we should probably warn or error if you pass a single tag and we publish multiple images, because the tag will get overwritten. Alternatively (lazily), just document this caveat in the README or flag description.

jonjohnsonjr (approved):

PR #238, opened 2020-11-04
Leverage remote.MultiWrite to efficiently publish images

imjasonh:
This adds a MultiPublish method to the publish.Interface, which
publishers can use to optimize publishing many images, and which the
defalt publisher implements by leveraging remote.MultiWrite, which
dedupes blob uploads, reuses auth handshakes and parallelizes blob and
manifest uploads.

All other publisher implementations implement MultiPublish by naively
calling their Publish method in a loop, but some could possibly leverage
more smarts in the future.


pkg/resolve/resolve.go
```diff
@@ -33,61 +33,68 @@ import (
...
 			} else if strict {
-				return fmt.Errorf("found strict reference but %s is not a valid import path", ref)
+				return fmt.Errorf("found strict reference but %s is not a valid import path", ip)
 			}
 		}
 	}
 
 	// Next, perform parallel builds for each of the supported references.
-	var sm sync.Map
+	var sm sync.Map // importpath string -> build.Result
```
jonjohnsonjr, line 53: @mattmoor can you sanity check the changes in this file?

pkg/publish/default.go
```diff
@@ -119,6 +119,52 @@ func pushResult(tag name.Tag, br build.Result, opt []remote.Option) error {
...
+				return nil, err
+			}
+			m[tag] = br
+		}
+
+		h, err := br.Digest()
+		if err != nil {
+			return nil, err
+		}
+		ref := fmt.Sprintf("%s/%s@%s", d.base, d.namer(sanitized), h)
```
mattmoor, line 149: I think there's a merge conflict with my change here.
```suggestion
		ref := fmt.Sprintf("%s@%s", d.namer(d.base, sanitized), h)
```

pkg/publish/default.go
```diff
@@ -119,6 +119,52 @@ func pushResult(tag name.Tag, br build.Result, opt []remote.Option) error {
...
+		no = append(no, name.Insecure)
+	}
+
+	out := map[string]name.Reference{}
+	m := map[name.Reference]remote.Taggable{}
+	for ip, br := range in {
+		// https://github.com/google/go-containerregistry/issues/212
+		sanitized := strings.TrimPrefix(strings.ToLower(ip), build.StrictScheme)
+		for _, tagName := range d.tags {
+			tag, err := name.NewTag(fmt.Sprintf("%s/%s:%s", d.base, d.namer(sanitized), tagName), no...)
```
mattmoor, line 135: same as below

pkg/resolve/resolve.go
```diff
@@ -33,61 +33,68 @@ import (
...
-			img, err := builder.Build(ctx, ref)
+			img, err := builder.Build(ctx, ip)
 			if err != nil {
 				return err
 			}
-			digest, err := publisher.Publish(img, ref)
-			if err != nil {
-				return err
-			}
-			sm.Store(ref, digest.String())
```
mattmoor, line 66: We used to publish in go routines, and now we don't.  This feels like it could slow things down for high-arity resolves by doing all of the pushes after all of the builds have completed 🤔

ImJasonH: MultiPublish does publish blobs and manifests in goroutines, after all the builds are complete. It might matter if your publish fails and you could have saved some time building, but I don't know if we should optimize for that.

Am I misunderstanding this comment?

ImJasonH:
FYI this is blocked on https://github.com/google/go-containerregistry/pull/809

pkg/resolve/resolve.go
mattmoor: We used to have:
```
build -> publish. \
build -> publish. --- done!
build -> publish. /
```

We now have:
```
build  \     / publish  \
build   ---  publish  --- done
build  /     \ publish /
```

The new join point between builds and publishes seems to hinder parallelism, which seems counter to what we want, right?

ImJasonH: Ah I see, yeah. I believe the win from avoiding auth handshakes and deduping blobs will make this faster overall, but I don't think we can speak in absolutes about every use case and destination repo.

@jonjohnsonjr wdyt 🤔

jonjohnsonjr: This will largely depend on the inputs and resources available so it's hard to say what would be actually faster, but I think this change is more likely to slow things down, since we're unable to use network for uploading one image while we use CPU to build the other...

mattmoor: I wonder if there is a streaming interface we might consider that could get the best of both worlds.  What do we need to know in advance for this to work?

ImJasonH: So I ran some basic tests comparing before/after this change resolving Tekton's configs, with and without a warm Go build cache, and this change comes out slightly ahead in both cases.

```
# Cold build cache
$ go clean -cache
$ time ko resolve --platform=all -f config/
...
real	5m29.013s
user	26m30.454s
sys	7m51.594s

$ go clean -cache
$ time new-ko resolve --platform=all -f config/
...
real	6m12.956s
user	30m18.792s
sys	8m35.511s

# Warm build cache
$ time ko resolve --platform=all -f config/
...
real	1m46.570s
user	4m55.177s
sys	3m29.805s

$ time new-ko resolve --platform=all -f config/
...
real	1m38.875s
user	4m47.177s
sys	3m23.662s
```

We could have a streaming write that takes a `<-chan struct{name.Reference, remote.Taggable}`, and keeps a cache of blobs/manifests it's published, does a new auth handshake for each new repo it sees, consumes the chan with N workers, etc. 🤔 

There's still the problem of needing to publish sub-manifests before committing indexes, but that's doable. I'll keep `MultiWrite` in ggcr since I think it can still be useful, and if/when streaming write exists it would just be another facade on it.

jonjohnsonjr: Those results make sense to me. Did you clear the registry between runs as well?

ImJasonH: I didn't clear the registry, but I don't think any case actually pushed any new layers, which I think is usually the common case. I'd expect new-ko to be faster if there are actually many layers to upload or mount.

Issue #155, opened 2020-04-30
Support `ko login`


mattmoor:
`ko` consumes the Docker keyring, but generally it is a pain in the butt to configure this without Docker installed.

I propose we add `ko login` mimicing the behavior of `docker login` so that folks can reasonable authenticate without having docker installed.

This likely needs work in github.com/google/go-containerregistry, and come to think of it a `crane login` command sounds useful too!

mattmoor:
@jonjohnsonjr since you added `crane auth login` 👀

jonjohnsonjr:
Fixed by https://github.com/google/ko/pull/243

alexellis:
Hi guys, the README file says `ko auth login` exists, but go get with the latest stable release doesn't seem to support it.

Is the resolution to pull the master version or is that a bug?

Alex

jonjohnsonjr:
Yeah... https://github.com/google/ko/pull/243#issuecomment-724269021

I can refactor some things to make this less gross (we just want `ko login` not `ko auth login`, IMO.

alexellis:
I would like to see some output, if you're open to it after running the login command like `Wrote to... ~/.docker/config.json` or some other output if the file wasn't updated (because it already existed)

jonjohnsonjr:
> I would like to see some output, if you're open to it after running the login command like Wrote to... ~/.docker/config.json or some other output if the file wasn't updated (because it already existed)

Seems reasonable to me!

jonjohnsonjr:
Once we update dependencies, we should be able to merge a fix for this.

PR #243, opened 2020-11-07
Add crane's auth command to facilitate logging into a registry

tcnghia:
Fixes https://github.com/google/ko/issues/155

/assign @jonjohnsonjr @mattmoor

### DIFF
--- a/cmd/ko/main.go
+++ b/cmd/ko/main.go
@@ -21,6 +21,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/logs"
 	"github.com/google/ko/pkg/commands"
 
+	cranecmd "github.com/google/go-containerregistry/cmd/crane/cmd"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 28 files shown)


cmd/ko/main.go
```diff
@@ -38,6 +39,10 @@ func main() {
 	}
 	commands.AddKubeCommands(cmds)
 
+	// Also add the auth group from crane to facilitate logging into a
+	// registry.
+	cmds.AddCommand(cranecmd.NewCmdAuth())
```
mattmoor, line 44: Can you document it?

tcnghia: Done

mattmoor (approved):

jonjohnsonjr:
I would have just snagged the `login` command, not both, but the cobra doc is currently hardcoded to `crane`:

```
$ ko auth login --help
Log in to a registry

Usage:
  ko auth login [OPTIONS] [SERVER] [flags]

Examples:
  # Log in to reg.example.com
  crane auth login reg.example.com -u AzureDiamond -p hunter2

Flags:
  -h, --help              help for login
  -p, --password string   Password
      --password-stdin    Take the password from stdin
  -u, --username string   Username
```

Easy, hacky solution would be to just replace `logincmd.Example` with your own hard-coded string.

More involved solution would be to rework `NewCmdAuthLogin` to figure out what command group it's in and print the right thing. I'm not sure if that's possible.

mattmoor:
@jonjohnsonjr We could make the function use `os.Args[0]` 😉

jonjohnsonjr:
That's not sufficient because for `crane` that would just be `crane login` but in reality we need `crane auth login`, which make me regret this command group entirely, and I think they should have just been top level.

Issue #249, opened 2020-11-19
Watch stops after first build when using ko:// prefix


halvards:
When running `resolve` with watch and the `ko://` prefix, the watch loop stops after the first iteration with this message:

`watching dependencies: adding importpath to dep graph: package ko:/[import-path] is not in GOROOT ([GOROOT]/src/ko:/[import-path])`

It seems to lose one of the slashes (`ko:/[import-path]`), and I'm not sure why the image reference is appended to `GOROOT`, with `/src` in between. 

I tried with both `v0.6.2` and latest from master (`bd1ec54`).

Full output:

```
$ ko resolve -f manifests -W
2020/11/20 07:57:05 Using base gcr.io/distroless/static:nonroot for github.com/halvards/ghr-test
2020/11/20 07:57:10 Building github.com/halvards/ghr-test for linux/amd64
2020/11/20 07:57:12 Publishing ghcr.io/halvards/ghr-test-290f42b8e1fff6505c8434193ee80807:latest
2020/11/20 07:57:13 existing blob: sha256:dc8715d5ff921a96369bfbaa96d3266ee7a63b635f5ba5d33bdb4a087c924a18
2020/11/20 07:57:14 existing blob: sha256:e59bd8947ac7af2c8f4403b183326986ab554bbc262739cf5d9d596c7c7a3aca
2020/11/20 07:57:14 existing blob: sha256:fe64bc90abda5ce3e0111a4a5f0e2c1ef75bdc3b49c9e8ffd334e45bca8a036c
2020/11/20 07:57:14 existing blob: sha256:72164b581b02b1eb297b403bcc8fc1bfa245cb52e103a3a525a0835a58ff58e2
2020/11/20 07:57:15 ghcr.io/halvards/ghr-test-290f42b8e1fff6505c8434193ee80807:latest: digest: sha256:0e28daaa75fa95dd8f537516b57ad909c7048d811f0848d3faadcd9635565be5 size: 750
2020/11/20 07:57:15 Published ghcr.io/halvards/ghr-test-290f42b8e1fff6505c8434193ee80807@sha256:0e28daaa75fa95dd8f537516b57ad909c7048d811f0848d3faadcd9635565be5
apiVersion: v1
kind: Pod
metadata:
  name: ghr-test
spec:
  containers:
    - name: ghr-test
[...]

PR #250, opened 2020-11-20
Make --watch work with ko://

jonjohnsonjr:
Fixes https://github.com/google/ko/issues/249

The dep-notify package doesn't know anything about ko, so we need to
trim the strict prefix before sending import paths over to it.


mattmoor (approved):

jonjohnsonjr:
These non-linux/amd64 platforms really like sitting in a queue on travis.

PR #257, opened 2020-12-04
Move ko binary to root of project.

n3wscott:
It has become clear `ko` and this repo is a single tool repo. Let's embrace that, and allow `go get github.com/google/ko` to work as 99% of folks try to do without reading the readme.


ko.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/ko/pkg/commands"
+
+	cranecmd "github.com/google/go-containerregistry/cmd/crane/cmd"
+	"github.com/spf13/cobra"
+)
+
+func main() {
```
jonjohnsonjr, line 28: Is this filename (`ko.go`) a convention for single tool repos? I'd prefer `main.go` I think.

n3wscott: the filename dictates what the installed binary will be I think. Let me double check.

n3wscott: it worked with main.go. Switched.

README.md
```diff
@@ -13,10 +13,10 @@
 
 `ko` can be installed and upgraded by running:
 
-**Note**: Golang version `1.12.0` or higher is required.
+**Note**: Golang version `1.14.0` or higher is required.
 
 ```shell
-GO111MODULE=on go get github.com/google/ko/cmd/ko
+go get github.com/google/ko
```
ImJasonH, line 19: What should we do for users that install from cmd/ko today in scripts? This change would break them without a lot of help about what changed or how to fix it.

Can we keep cmd/ko/main.go as an alias for a release or two with a deprecation warning?

hack/update-deps.sh
```diff
@@ -29,3 +29,8 @@ go mod vendor
 # Delete all vendored broken symlinks.
 # From https://stackoverflow.com/questions/22097130/delete-all-broken-symbolic-links-with-a-line
 find vendor/ -type l -exec sh -c 'for x; do [ -e "$x" ] || rm "$x"; done' _ {} +
+
+
+# HACK HACK HACK:
+# Until we can cleanly remove it, copy root main.go and move it to ./cmd/ko
+cp ${PROJECT_ROOT}/main.go ${PROJECT_ROOT}/cmd/ko/
```
jonjohnsonjr, line 36: @ImJasonH ^

Is it possible to just commit a symlink to the root? Or would that break everything?

ImJasonH: Missed this 😇

I'd prefer something explicit since it also gives us an opportunity to inject a deprecation warning

n3wscott: Added a message, dropped the copy,

README.md
n3wscott: That was always my plan to keep both. We have a lot of tooling that hardcodes installing ko from the `./cmd/ko` path.

n3wscott: > This change would break them without a lot of help about what changed or how to fix it.

It never broke them, I left `./cmd/ko` in the PR.

n3wscott: I did drop `./cmd/ko/test`

README.md
```diff
@@ -13,10 +13,10 @@
 
 `ko` can be installed and upgraded by running:
 
-**Note**: Golang version `1.12.0` or higher is required.
+**Note**: Golang version `1.14.0` or higher is required.
```
ImJasonH, line 16: Why is 1.14 required now?

ImJasonH (approved):

README.md
n3wscott: That is what the go mod file says.

Issue #253, opened 2020-11-30
Feature request - option to publish N architectures


alexellis:
I tend to build for armhf, arm64 and Intel, so could use an option to give the three architectures. When @mattmoor gave me a demo earlier today (very impressed with the workflow) - it seemed like I could only give one architecture or the predefined list in all?

mattmoor:
cc @jonjohnsonjr 

Yeah, right now there is only `--platform=all` and `--platform=os/arch`

jonjohnsonjr:
> or the predefined list in all

This isn't actually predefined, it's derived from the base image. The default base image happens to be `gcr.io/distroless/static:nonroot`:

```
$ crane manifest gcr.io/distroless/static:nonroot
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 426,
         "digest": "sha256:572620cc4108b51662307290f1c9c6cd4cb6b42eedc68d218c7ba50e709fbd3e",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 426,
         "digest": "sha256:c367236866f91e24bb793a2e764918b9fd1bef8d4a4e292bc5f184f02318b216",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 426,
         "digest": "sha256:7b1b2d30014678efbfc19980b924f134cfcbe38bdac4a33a5e2d8d15dc12c3e6",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 426,
         "digest": "sha256:e83342ea9d8278ecee63d0b2c867515a03d21389e1c1b9227a924fb1d9bfd9f6",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 426,
         "digest": "sha256:796ff193c9ae25911401d4e687b9646ad81e4bec028876073cc5dcfbed694c14",
         "platform": {
            "architecture": "s390x",
[...]

PR #259, opened 2020-12-07
Allow comma-separated list of platforms

jonjohnsonjr:
Fixes https://github.com/google/ko/issues/253

Fixes https://github.com/google/ko/issues/274

```
$ ko publish -B --platform=linux/amd64,linux/arm64 ./cmd/ko/
2020/12/07 13:57:20 Using base gcr.io/distroless/static:nonroot for github.com/google/ko/cmd/ko
2020/12/07 13:57:21 Building github.com/google/ko/cmd/ko for linux/amd64
2020/12/07 13:57:25 Building github.com/google/ko/cmd/ko for linux/arm64
2020/12/07 13:57:29 Publishing gcr.io/jonjohnson-test/ko/ko:latest
2020/12/07 13:57:30 existing blob: sha256:72164b581b02b1eb297b403bcc8fc1bfa245cb52e103a3a525a0835a58ff58e2
2020/12/07 13:57:30 existing blob: sha256:e59bd8947ac7af2c8f4403b183326986ab554bbc262739cf5d9d596c7c7a3aca
2020/12/07 13:57:31 pushed blob: sha256:314bd22f324d73c81d23dd94a14e3a4df6a591381e39b9164163208115c23ae4
2020/12/07 13:57:33 pushed blob: sha256:dd714425e64113dddc2cf7a21a1b7f47c20ae29dc2072284b64fe1e7b82a5307
2020/12/07 13:57:33 gcr.io/jonjohnson-test/ko/ko@sha256:cb6a0c6cb99e6c7b1d7dfd598484d0d25b74e15a69837f3d87c08c58ee7f88e2: digest: sha256:cb6a0c6cb99e6c7b1d7dfd598484d0d25b74e15a69837f3d87c08c58ee7f88e2 size: 751
2020/12/07 13:57:34 existing blob: sha256:72164b581b02b1eb297b403bcc8fc1bfa245cb52e103a3a525a0835a58ff58e2
2020/12/07 13:57:34 existing blob: sha256:d34b09c140e4b67f2c6214e24f26a68bc6b17c28d42f0dda89db38ad03488416
2020/12/07 13:57:35 pushed blob: sha256:99af2cd1b289d3a47aba52934e2ac8c4037e92563e8a4d119b5d0e040f5b9ed3
2020/12/07 13:57:37 pushed blob: sha256:052d3194d06e9194d60e7dcdce4c7d8e91df952f34dfddd461ce181732342db2
2020/12/07 13:57:37 gcr.io/jonjohnson-test/ko/ko@sha256:ca19f1aa2adc983e55c1eb7d6af1ed8574991ff6762eccec8ef7b5e5d965d1ce: digest: sha256:ca19f1aa2adc983e55c1eb7d6af1ed8574991ff6762eccec8ef7b5e5d965d1ce size: 751
2020/12/07 13:57:38 gcr.io/jonjohnson-test/ko/ko:latest: digest: sha256:03ed41dd8676e9988fb4958366cf8d38fbb39e576d5b09bb1cc108c937e6d7ff size: 529
2020/12/07 13:57:38 Published gcr.io/jonjohnson-test/ko/ko@sha256:03ed41dd8676e9988fb4958366cf8d38fbb39e576d5b09bb1cc108c937e6d7ff
gcr.io/jonjohnson-test/ko/ko@sha256:03ed41dd8676e9988fb4958366cf8d38fbb39e576d5b09bb1cc108c937e6d7ff
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -454,13 +454,25 @@ a multi-architecture
 (aka "manifest list"), with support for each OS and architecture pair supported
 by the base image.
 
-If `ko` is invoked with `--platform=<some-OS>/<some-platform>` (e.g.,
+If `ko` is invoked with `--platform=<some-OS>/<some-arch>` (e.g.,
 `--platform=linux/amd64` or `--platform=linux/arm64`), then it will attempt to
 build an image for that OS and architecture only, assuming the base image
 supports it.
 
-When `--platform` is not provided, `ko` builds an image with the OS and
-architecture based on the build environment's `GOOS` and `GOARCH`.
+If the base image is a manifest list with more platforms than you want to build,
+invoking `ko` with comma-separated list of platforms (e.g.
+`--platform=linux/amd64,linux/arm/v6`) will produce a manifest list
+containing only the provided platforms. Note that if the base image does not
+contain platforms that are provided by this flag, `ko` will be unable to build
+a corresponding image, and this is not an error. The resulting artifact will be
+a multi-platform image containing the intersection of platforms from the base
+image and the `--platform` flag. This is especially relevant for projects that
+use multiple base images, as you must ensure that every base image contains all
+the platforms that you'd like to build.
+
+When `--platform` is not provided, if both `GOOS` and `GOARCH` environment
+variables are set, `ko` will build an image for `${GOOS}/${GOARCH}`, otherwise
+`ko` will build a `linux/amd64` image.
 
 ## Enable Autocompletion
 
--- a/README.md
+++ b/README.md
@@ -471,8 +471,11 @@ use multiple base images, as you must ensure that every base image contains all
 the platforms that you'd like to build.
 
 When `--platform` is not provided, if both `GOOS` and `GOARCH` environment
-variables are set, `ko` will build an image for `${GOOS}/${GOARCH}`, otherwise
-`ko` will build a `linux/amd64` image.
+variables are set, `ko` will build an image for `${GOOS}/${GOARCH}[/v${GOARM}]`,
+otherwise `ko` will build a `linux/amd64` image.
+
+Note that using both `--platform` and GOOS/GOARCH will return an error, as it's
+unclear what platform should be used.
 
 ## Enable Autocompletion
 
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -54,6 +54,11 @@ type buildContext interface {
 	Import(path string, srcDir string, mode gb.ImportMode) (*gb.Package, error)
 }
 
+type platformMatcher struct {
+	spec      string
+	platforms []v1.Platform
+}
+
 type gobuild struct {
 	getBase              GetBase
 	creationTime         v1.Time
@@ -61,8 +66,7 @@ type gobuild struct {
 	disableOptimizations bool
 	mod                  *modules
 	buildContext         buildContext
-	platform             string
-	parsedPlatform       []v1.Platform
+	platformMatcher      *platformMatcher
 }
 
 // Option is a functional option for NewGo.
@@ -82,7 +86,7 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 	if gbo.getBase == nil {
 		return nil, errors.New("a way of providing base images must be specified, see build.WithBaseImages")
 	}
-	parsed, err := parseSpec(gbo.platform)
+	matcher, err := parseSpec(gbo.platform)
 	if err != nil {
 		return nil, err
 	}
@@ -93,8 +97,7 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 		disableOptimizations: gbo.disableOptimizations,
 		mod:                  gbo.mod,
 		buildContext:         gbo.buildContext,
-		platform:             gbo.platform,
-		parsedPlatform:       parsed,
+		platformMatcher:      matcher,
 	}, nil
 }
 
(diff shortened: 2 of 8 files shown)


ImJasonH:
Can you include README.md in this change?

In particular, what happens if I pass `ko resolve --platform=A,B` and some of the images built in the process support both platforms while some only support one or the other (or neither).

jonjohnsonjr:
Will do -- also want to fix the TODO in there, but wanted to hack something together to play with the UX.

jonjohnsonjr:
PTAL, cc @mattmoor

pkg/build/gobuild_test.go
```diff
@@ -607,3 +608,81 @@ func TestGoarm(t *testing.T) {
...
+		err:  true,
+	}, {
+		spec: "",
+		platform: &v1.Platform{
+			Architecture: "amd64",
+			OS:           "linux",
+		},
+		result: false,
+	},
+	} {
```
mattmoor, line 672:
```suggestion
	}} {
```

README.md
```diff
@@ -454,13 +454,25 @@ a multi-architecture
...
+contain platforms that are provided by this flag, `ko` will be unable to build
+a corresponding image, and this is not an error. The resulting artifact will be
+a multi-platform image containing the intersection of platforms from the base
+image and the `--platform` flag. This is especially relevant for projects that
+use multiple base images, as you must ensure that every base image contains all
+the platforms that you'd like to build.
+
+When `--platform` is not provided, if both `GOOS` and `GOARCH` environment
+variables are set, `ko` will build an image for `${GOOS}/${GOARCH}`, otherwise
+`ko` will build a `linux/amd64` image.
```
mattmoor, line 475: should thus mention GOARM?

jonjohnsonjr: I don't think so? Defaulting doesn't consider GOARM at all.

pkg/build/gobuild.go
```diff
@@ -73,19 +75,26 @@ type gobuildOpener struct {
...
+	}
 	return &gobuild{
 		getBase:              gbo.getBase,
 		creationTime:         gbo.creationTime,
 		build:                gbo.build,
 		disableOptimizations: gbo.disableOptimizations,
 		mod:                  gbo.mod,
 		buildContext:         gbo.buildContext,
+		platform:             gbo.platform,
+		parsedPlatform:       parsed,
```
mattmoor, line 97: Feels like we've got a datastructure we're hiding in a couple fields of gobuild.  Should we create a little PlatformMatcher that we build from the string, which captures this state and surfaces methods for matching a platform?

README.md
mattmoor: I wonder if it should 🤔

jonjohnsonjr: Oh you're saying we can use `GOARM` to break platform.Variant ties in base image resolution if all of GOOS/GOARCH/GOARM are set? Not a bad idea.

jonjohnsonjr: Though... a little ambiguous. Currently local env overrides the platform from the base image (GOOS and GOARCH) [here](https://github.com/google/ko/blob/960e6a3f1c039d3b7635c7834b10dab59d548b2a/pkg/build/gobuild.go#L259-L274).

So we can't use GOARM for both selection and overriding in every context. I'm having a hard time wrapping my head around how the defaulting, selection, and overriding precedence rules should work.

I guess maybe if you say `--platform=linux/arm` and set `GOARM=6`, we would take the first `linux/arm` image and always use `GOARM=6`, but if you don't set `--platform` and _do_ set all of GOOS/GOARCH/GOARM, we will use that whole thing as our default platform? We now need to map from `GOARM` to `platform.Variant` (just prepend a `v`, I guess?).

ImJasonH: Just a reminder that if these questions seem too complex to answer, we can still just not support comma-delimited archs and still only support `--all`, until such a time as we have a user that wants it and can tell us how they'd like to use it.

mattmoor: Alex Ellis was asking, and knative/serving would probably also be fine with restricting multi-arch to amd64/arm64 since Envoy only ships those.

ImJasonH: Sure. For Tekton at least, I think we're fine producing images that happen to work on some archs because we produce them with `--platform=all`, but we won't commit to supporting them unless explicitly stated otherwise. Knative etc could follow the same path.

In particular this makes it easier to experiment with supporting new archs since it's already built and released using `--platform=all`, and you don't have to add `--platform=foo,bar,new-thing`.

mattmoor: Yeah, that's where we are now.  I was just looking to save some trees in places where Envoy didn't work 😉

jonjohnsonjr: >  these questions seem too complex to answer

I don't know if they're particularly complex to answer. Any behavior we pick is fine, IMO, but I'm trying to determine what the least surprising thing would be.

pkg/build/gobuild.go
jonjohnsonjr: Done, kind of. Not sure if we need this anywhere else, so didn't expose it.

README.md
jonjohnsonjr: > So we can't use GOARM for both selection and overriding in every context. I'm having a hard time wrapping my head around how the defaulting, selection, and overriding precedence rules should work.

It was true: https://github.com/google/ko/issues/274

jonjohnsonjr: Done.

mattmoor:
@googlebot I consent!

mattmoor:
@googlebot I consent.

jonjohnsonjr:
@googlebot I consent.

jonjohnsonjr:
> 2020/12/21 21:57:23 error creating builder: error setting up builder options: cannot use --platform with GOOS="linux"

This failed in the way I wanted it to, wdyt @mattmoor

mattmoor (approved):
Seems fine to me.  `--platform` is the future :)

PR #263, opened 2020-12-13
Update ggcr and use the new tarball.WithCompressCaching

mattmoor:
This simplifies the logic for creating layers and relies on the new option instead of pre-gzipping (and gunzip being faster).  The net effect of this should be the elimination of an extra gunzip, but it should not materially impact the memory pressure because we were ALREADY caching the gzipped layers in-memory, we were just doing it on the caller side, which provided less benefit!


mattmoor:
CLA bot, you have the highest ratio of annoying to useful of anything in my life.  🤬

mattmoor:
@googlebot I signed it!

mattmoor:
@googlebot rescan

jonjohnsonjr (approved):

mattmoor:
@googlebot rescan

PR #266, opened 2020-12-14
documentation: auth without docker, not without ko

arve0:
I believe it should without docker, right? Using ko without ko sounds wrong.


jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

PR #271, opened 2020-12-17
Allow ko to produce estargz layers.

mattmoor:
This pulls in the latest google/go-containerregistry, which enables folks to set `GGCR_EXPERIMENT_ESTARGZ=1` to enable `ko` to start producing layers compatible with the estargz format, which enables the containerd estargz-snapshotter to lazy-load parts of images.


jonjohnsonjr:
> GGCR_EXPERIMENT_ESTARGZ=1

Should we document this or do you want to soft launch for now?

jonjohnsonjr (approved):

PR #288, opened 2021-01-06
fixed: comment and readme are incorrect, there is no ko login

n3wscott:
`ko login` is not a valid command, it is `ko auth login`.


jonjohnsonjr:
https://github.com/google/ko/pull/277

jonjohnsonjr:
image

PR #309, opened 2021-02-06
support mips64le architecture.

houfangdong:
I am going to submit mips64le architecture，The main reasons are:
1、The mips64le architecture is also the mainstream cpu architecture (amd64, arm64), which is used by many users;
2、Golang supports cross compilation, and mips64le is officially supported;
Therefore, I hope that the release package also supports the mips64le architecture, which is convenient for more users.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -19,6 +19,7 @@ builds:
   - arm64
   - s390x
   - 386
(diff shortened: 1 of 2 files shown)


ImJasonH (approved):
@jonjohnsonjr any problem with this?

PR #313, opened 2021-02-18
options: update --bare help message

zhouhaibing089:
The current `--bare` help message is a little bit confusing: It says
that it will not work with `--tags` properly, but in fact, I would
assume it means that it may not work properly instead.

With `--bare`, the name will be `${KO_DOCKER_REPO}:<tag>`.


ImJasonH:
cc @mattmoor who added this iirc

mattmoor:
This seems fine to me.

ImJasonH (approved):

PR #318, opened 2021-02-23
Streamline README.md

ImJasonH:
[Rendered](https://github.com/imjasonh/ko/blob/readme/README.md)

Recent feedback ([1](https://twitter.com/thomrstrom/status/1364028929514696704), [2](https://twitter.com/patricknflynn/status/1364213599426785289)) indicate that it's not clear enough that `ko` is a useful option for simply building container images, separate from its Kubernetes support.

This reorders README.md to start with installation options, then highlight `ko publish` first, then sidetrack into configuration options, then mention Kubernetes support (`ko resolve` -> `ko apply` -> `ko delete`), and end with an FAQ.

It's about 60% as many lines and half as many characters as the current README.

PR #321, opened 2021-02-25
make installation step download 0.8.1

sebgoa:


ImJasonH:
Rather than sign ourselves up to update README every release (or forget to, more likely), #318 makes the user choose a version explicitly.

sebgoa:
fair enough. This is my drive by fix it style without checking other PR :)

closing this in favor of #318

PR #324, opened 2021-03-03
Add --image-label to add labels to built images

### LINKED ISSUE #323: Pass labels to v1.Config during build
Hi, thanks for working on this awesome project. This is very useful in our Go-centered projects.

One thing which can be nice to have is to add `Labels` to the image config. Today, some container registries have API(s) around image labels. For e.g, it is possible to [add a specified label in Quay][1] to tell the registry when this image will expire so will be cleaned up automatically. This is an awesome feature for CI images, by the way.

```Dockerfile
LABEL quay.expires-after=20h
```

[kaniko][2] allows us to do this with a flag `--label` as describe [here][3]. I also looked at the ko code base and if I understand correctly, here is the place where we can tweak a bit to support labels as well?

https://github.com/google/ko/blob/c14c08e982592bcea40a3bd95de94836bfbe40f9/pkg/build/gobuild.go#L603-L607

Do you think this is something reasonable to be added? Thanks. :)

[1]: https://access.redhat.com/documentation/en-us/red_hat_quay/3/html/use_red_hat_quay/working_with_tags#setting_tag_expiration_from_a_dockerfile
[2]: https://github.com/GoogleContainerTools/kaniko
[3]: https://github.com/GoogleContainerTools/kaniko#--label

ImJasonH:
Fixes #323

cc @zhouhaibing089

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -83,6 +83,7 @@ type gobuild struct {
 	mod                  *modules
 	buildContext         buildContext
 	platformMatcher      *platformMatcher
+	labels               map[string]string
 }
 
 // Option is a functional option for NewGo.
@@ -96,6 +97,7 @@ type gobuildOpener struct {
 	mod                  *modules
 	buildContext         buildContext
 	platform             string
+	labels               map[string]string
 }
 
 func (gbo *gobuildOpener) Open() (Interface, error) {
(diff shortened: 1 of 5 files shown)


zhouhaibing089 (approved):
Thanks!

jonjohnsonjr (approved):
Nice, I love clean diffs like this :)

PR #332, opened 2021-03-30
Add a tag-only flag to publisher.

chhsia0:
The `--tag-only` flag resolves images into tag-only references. This flag is useful when digests are not preserved when images are repopulated. It can only be used when a single tag is specified, and the tag is not `latest`.

### DIFF
--- a/pkg/commands/options/publish.go
+++ b/pkg/commands/options/publish.go
@@ -26,6 +26,8 @@ import (
 // PublishOptions encapsulates options when publishing.
 type PublishOptions struct {
 	Tags []string
+	// TagOnly resolves images into tag-only references.
+	TagOnly bool
 
 	// Push publishes images to a registry.
 	Push bool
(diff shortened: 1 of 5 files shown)


ImJasonH:
I'm not sure I understand the use case, in what cases are references by digest not preferred?

Images by tag are problematic because the tag can change and you can get unexpected/inconsistent behavior across replicas/nodes.

jonjohnsonjr:
I can imagine a couple situations where this could be useful, e.g. when writing to the docker daemon, but not when pushing to the registry. Agree with Jason, I'd like to understand the problem a little more.

chhsia0:
Sure I could explain more. Our use case is to use `ko` to generate manifests that supports air-gapped environments, where we repopulate necessary images to containerd from tarballs (similar to `docker load`). However, such repopulation does not preserve image digests, so we have to use tags to refer to a certain image version in the manifests so they can be deployed in a normal and an air-gapped environment.

jonjohnsonjr:
> I also put a constraint for this tag to disallow people from generating image references that uses the (commonly) moving latest tag.

I'm not sure if we want to be that opinionated about `latest`. I generally agree with you, but it feels like an artificial constraint. That being said, it _is_ easier to relax this if people complain than it would be to introduce this constraint and break people who rely on it, so I don't have a strong opinion.

I'm ok to merge unless @imjasonh has objections

imjasonh:
Sgtm

jonjohnsonjr (approved):

PR #340, opened 2021-04-19
Add support for Go build flags

### LINKED ISSUE #316: Support setting Go build flags via configuration file
Although, build flags [like `-ldflags` can be set using an environment variable](https://github.com/google/ko#how-can-i-set-ldflags), there is a problem if you want to set multiple linker flags like that. Since the [Go docs on the environment variables now state that spaces are not allowed](https://golang.org/cmd/go/#hdr-Environment_variables) in this context and specifying the flags multiple times does not work due to the way overrides are handled, I do not currently see a good way to specify multiple `ldflags`.  The use-case I have in mind would be to set `ldflags` for `-s`, `-w`, and `-X`, which is not an uncommon combination. In [GoReleaser this is solved by section in the configuration](https://goreleaser.com/customization/build/) to specify the respective settings, which I think is quite nice. Any chance that the `.ko.yaml` could be used for this kind of configuration, too? Thanks in advance for your feedback.

HeavyWombat:
Fixes #316

There are use cases, where multiple Go build flags need to be set. However, the environment variable to pass flags to Go build has some limits for `ldFlags`.

Proposed changes:
- Add GoReleaser inspired configuration section to `.ko.yaml` to support setting specific Go build and ldFlags to be used by the build.
- Add support for environment variables `KO_GOBUILD_FLAGS` to add extra build flags and `KO_GOBUILD_LDFLAGS` to add extra ldFlags.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,42 @@ baseImageOverrides:
   github.com/my-user/my-repo/cmd/foo: registry.example.com/base/for/foo
 ```
 
+### Overriding Go build settings
+
+By default, `ko` builds the binary with no additional build flags other than
+`--trimpath` (depending on the Go version). You can replace the default build
+arguments by providing build flags and ldflags using a
+[GoReleaser](https://github.com/goreleaser/goreleaser) influenced `builds`
+configuration section in your `.ko.yaml`.
+
+```yaml
+builds:
+- id: foo
+  main: ./foobar/foo
+  flags:
+  - -tags
+  - netgo
+  ldflags:
+  - -s -w
+  - -extldflags "-static"
+  - -X main.version={{.Env.VERSION}}
+- id: bar
+  main: ./foobar/bar/main.go
+  ldflags:
+  - -s
+  - -w
+```
+
+For the build, `ko` will pick the entry based on the respective import path
+being used. It will be matched against the local path that is configured using
+`dir` and `main`. In the context of `ko`, it is fine just to specify `main`
+with the intended import path.
+
+_Please note:_ Even though the configuration section is similar to the
+[GoReleaser `builds` section](https://goreleaser.com/customization/build/),
+only the `flags` and `ldflags` fields are currently supported. Also, the
+templating support is currently limited to environment variables only.
+
 ## Naming Images
 
 `ko` provides a few different strategies for naming the image it pushes, to
@@ -335,9 +371,16 @@ is a common way to embed version info in go binaries (In fact, we do this for
 this flag directly; however, you can use the `GOFLAGS` environment variable
 instead:
 
-```
+```sh
 GOFLAGS="-ldflags=-X=main.version=1.2.3" ko publish .
 ```
+
+## How can I set multiple `ldflags`?
+
+Currently, there is a limitation that does not allow to set multiple arguments
+in `ldflags` using `GOFLAGS`. Using `-ldflags` multiple times also does not
+work. In this use case, it works best to use the [`builds` section](#overriding-go-build-settings)
+in the `.ko.yaml` file.
 
 ## Why are my images all created in 1970?
 
--- a/pkg/build/config.go
+++ b/pkg/build/config.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2021 Google LLC All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package build
+
+import "strings"
+
+// Note: The structs, types, and functions are based upon GoReleaser build
+// configuration to have a loosely compatible YAML configuration:
+// https://github.com/goreleaser/goreleaser/blob/master/pkg/config/config.go
+
+// StringArray is a wrapper for an array of strings.
+type StringArray []string
+
+// UnmarshalYAML is a custom unmarshaler that wraps strings in arrays.
+func (a *StringArray) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var strings []string
+	if err := unmarshal(&strings); err != nil {
+		var str string
+		if err := unmarshal(&str); err != nil {
+			return err
+		}
+		*a = []string{str}
+	} else {
+		*a = strings
+	}
+	return nil
+}
+
+// FlagArray is a wrapper for an array of strings.
+type FlagArray []string
+
+// UnmarshalYAML is a custom unmarshaler that wraps strings in arrays.
+func (a *FlagArray) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var flags []string
+	if err := unmarshal(&flags); err != nil {
+		var flagstr string
+		if err := unmarshal(&flagstr); err != nil {
+			return err
+		}
+		*a = strings.Fields(flagstr)
+	} else {
+		*a = flags
+	}
+	return nil
+}
+
+// Config contains the build configuration section. The name was changed from
+// the original GoReleaser name to match better with the ko naming.
+//
+// TODO: Introduce support for more fields where possible and where it makes
+///      sense for `ko`, for example ModTimestamp, Env, or GoBinary.
+//
+type Config struct {
+	// ID only serves as an identifier internally
+	ID string `yaml:",omitempty"`
+
+	// Dir is the directory out of which the build should be triggered
+	Dir string `yaml:",omitempty"`
+
+	// Main points to the main package, or the source file with the main
+	// function, in which case only the package will be used for the importpath
+	Main string `yaml:",omitempty"`
+
+	// Ldflags and Flags will be used for the Go build command line arguments
+	Ldflags StringArray `yaml:",omitempty"`
+	Flags   FlagArray   `yaml:",omitempty"`
+
+	// Other GoReleaser fields that are not supported or do not make sense
+	// in the context of ko, for reference or for future use:
+	// Goos         []string    `yaml:",omitempty"`
+	// Goarch       []string    `yaml:",omitempty"`
+	// Goarm        []string    `yaml:",omitempty"`
+	// Gomips       []string    `yaml:",omitempty"`
+	// Targets      []string    `yaml:",omitempty"`
+	// Binary       string      `yaml:",omitempty"`
+	// Env          []string    `yaml:",omitempty"`
+	// Lang         string      `yaml:",omitempty"`
+	// Asmflags     StringArray `yaml:",omitempty"`
+	// Gcflags      StringArray `yaml:",omitempty"`
+	// ModTimestamp string      `yaml:"mod_timestamp,omitempty"`
+	// GoBinary     string      `yaml:",omitempty"`
+}
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -33,6 +33,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/containerd/stargz-snapshotter/estargz"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -67,7 +68,7 @@ For more information see:
 // GetBase takes an importpath and returns a base image.
 type GetBase func(context.Context, string) (Result, error)
 
-type builder func(context.Context, string, string, v1.Platform, bool) (string, error)
+type builder func(context.Context, string, string, v1.Platform, Config) (string, error)
 
 type buildContext interface {
 	Import(path string, srcDir string, mode gb.ImportMode) (*gb.Package, error)
@@ -84,6 +85,7 @@ type gobuild struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	disableOptimizations bool
+	buildConfigs         map[string]Config
 	mod                  *modules
 	buildContext         buildContext
 	platformMatcher      *platformMatcher
@@ -100,6 +102,7 @@ type gobuildOpener struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	disableOptimizations bool
+	buildConfigs         map[string]Config
 	mod                  *modules
 	buildContext         buildContext
 	platform             string
@@ -121,6 +124,7 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 		kodataCreationTime:   gbo.kodataCreationTime,
 		build:                gbo.build,
 		disableOptimizations: gbo.disableOptimizations,
+		buildConfigs:         gbo.buildConfigs,
 		mod:                  gbo.mod,
 		buildContext:         gbo.buildContext,
 		labels:               gbo.labels,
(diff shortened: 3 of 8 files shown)


HeavyWombat:
@googlebot I signed it!

README.md
```diff
@@ -123,6 +123,31 @@ baseImageOverrides:
   github.com/my-user/my-repo/cmd/foo: registry.example.com/base/for/foo
 ```
 
+### Overriding Go build settings
+
+By default, `ko` builds the binary with no additional arguments other than `--trimpath` (if available) and the ones required to name the binary accordingly. You can extend the `go build` arguments by providing extra build flags using a [GoReleaser](https://github.com/goreleaser/goreleaser) influenced `builds` configuration section in your `.ko.yaml`.
```
jonjohnsonjr, line 128: I haven't fully thought through this idea, but:

It would be nice if the default build config (I'm thinking of `--trimpath` specifically here) were expressed in terms of this goreleaser config thing, and supplying your own build args (via the `builds` clause in `.ko.yaml`) would override it completely. This would give you a way to actually drop the `--trimpath` flag if you don't want it.

We would basically have good defaults for builds but offer the flexibility to do whatever you want. We can document the default or (better) have some way to print the default `.ko.yaml` file so folks can fiddle with it.

README.md
```diff
@@ -123,6 +123,31 @@ baseImageOverrides:
...
+  - -s -w
+  - -extldflags "-static"
+- id: bar
+  main: ./foobar/bar/main.go
+  ldflags:
+    - -s
+    - -w
+```
+
+For the build, `ko` will pick the entry based on the respective import path being used. It will be matched against the local path that is configured using `dir` and `main`. In the context of `ko`, it is fine just to specify `dir` with the intended import path.
```
jonjohnsonjr, line 147: Can this work for `ko` builds outside of the current directory?

README.md
```diff
@@ -324,6 +349,18 @@ instead:
 GOFLAGS="-ldflags=-X=main.version=1.2.3" ko publish .
 ```
 
+## How can I set multiple `ldflags`?
+
+Currently, there is a limitation that does not allow to set multiple arguments in `ldflags` using `GOFLAGS`. Using `-ldflags` multiple times also does not work. For this, `ko` supports two ways to have more complex Go build setups:
+
+- Using [`builds` section](#overriding-go-build-settings) in `.ko.yaml`.
+- Using environment variables `KO_GOBUILD_FLAGS` and `KO_GOBUILD_LDFLAGS`.
```
jonjohnsonjr, line 357: Given the config file, do we need this? I think the magic viper stuff allows setting these via automatically generated environment variables:

https://github.com/google/ko/blob/82cabb40bae577ce3bc016e5939fd85889538e8b/pkg/commands/config.go#L133

I guess `KO_GOBUILD_FLAGS` and `KO_GOBUILD_LDFLAGS` would apply globally to all builds, whereas the viper thing would require you to do this _per build_.

Another convenience is that you could specify this on the command line without having to produce an entire config file... even still, I think I'd rather this be a flag rather than an env... WDYT?

I'm not sure if that convenience is worth the introducing the new environment variable or flag, but I'm open to being convinced.

pkg/build/gobuild.go
```diff
@@ -38,6 +38,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/tarball"
 	"github.com/google/go-containerregistry/pkg/v1/types"
+	goreleaserCfg "github.com/goreleaser/goreleaser/pkg/config"
```
jonjohnsonjr, line 41: I think I would prefer copying what we actually need from this package, i.e. [spork](https://twitter.com/jonjonsonjr/status/1372415386516262915) it.

Importing this changes the dependency graph quite a bit, and we basically only need a couple structs and a serialization method (I think).

Maybe we start with this under `internal/goreleaser` for now?

jonjohnsonjr (commented):
Thanks for doing this -- a few comments, but it's looking pretty good.

pkg/build/options.go
```diff
@@ -45,6 +46,13 @@ func WithDisabledOptimizations() Option {
 	}
 }
 
+func WithGoReleaserBuildConfig(buildCfgOverrides map[string]goreleaserCfg.Build) Option {
```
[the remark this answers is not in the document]
HeavyWombat: Thanks for the hint! I have to check how I missed the linter warning for that.

README.md
HeavyWombat: It is a good point. I guess the tricky part is how to document it properly so that users understand to put `--trimpath` in there and not expect it to be automatically inserted as before. Will change that.

README.md
HeavyWombat: To be honest, the environment variables were more like an afterthought and mostly to fix the issue of having `ldflag` arguments that are dynamic, i.e. putting the result of `git describe --tags --always` into a variable. Since it is dynamic,  I figured the configuration file is not exactly the best place.

Anyway, I think that there is the use case for having a way to have dynamic content in your build arguments. The question would then be what is the best way to achieve that. I like that GoReleaser has Go template support for the strings being used, e.g. `"{{ .CommitTimestamp }}"` or `{{ .Version }}`. That would also be an option.

pkg/build/gobuild.go
HeavyWombat: Thanks for this feedback. I was annoyed by the changes, too. It will be ... sporked. You should probably provide a declension table. Would be fitting if it is an irregular verb.

README.md
jonjohnsonjr: >  I like that GoReleaser has Go template support for the strings being used, e.g. "{{ .CommitTimestamp }}" or {{ .Version }}. That would also be an option.

This seems reasonable to me. Where do those values come from? Can we make them work for any go project (not just within git). I know go tooling has some support for various non-git version control systems...

I see a bunch here: https://goreleaser.com/customization/templates/

We can at least start with supporting `.Env`, which should allow folks to hack their way around most problems.

README.md
HeavyWombat: Just to be on the safe side that I do not misunderstand the question, the setup would be something like the following:
```yaml
---
defaultBaseImage: registry.access.redhat.com/ubi8-minimal:latest

builds:
- dir: github.com/shipwright-io/build/cmd/manager
  flags: [ "-tags", "netgo" ]
  ldflags:
  - -s -w
  - -extldflags "-static"
```

```plain
/tmp $ KO_DOCKER_REPO=ko.local go run $GOPATH/src/github.com/google/ko/main.go publish --bare github.com/shipwright-io/build/cmd/manager
2021/04/22 22:59:39 adding import path github.com/shipwright-io/build/cmd/manager to build config overrides
2021/04/22 22:59:39 Using base registry.access.redhat.com/ubi8-minimal:latest for github.com/shipwright-io/build/cmd/manager
2021/04/22 22:59:39 No matching credentials were found for "registry.access.redhat.com/ubi8-minimal", falling back on anonymous
2021/04/22 22:59:40 go build command: /usr/local/bin/go build -tags netgo -ldflags=-s -w -extldflags "-static" -o /var/folders/dc/1k_907394hx5mft__7g269vh0000gn/T/ko458001556/out -trimpath github.com/shipwright-io/build/cmd/manager
2021/04/22 22:59:40 Building github.com/shipwright-io/build/cmd/manager for linux/amd64
2021/04/22 22:59:44 Loading ko.local:1881b84bc433524bed5364e6979f124aca4f6c9663b229baf68f97b06cd9f4c5
2021/04/22 22:59:47 Loaded ko.local:1881b84bc433524bed5364e6979f124aca4f6c9663b229baf68f97b06cd9f4c5
2021/04/22 22:59:47 Adding tag latest
2021/04/22 22:59:47 Added tag latest
ko.local:1881b84bc433524bed5364e6979f124aca4f6c9663b229baf68f97b06cd9f4c5
```

It is currently working, because the key used in the build configuration override map is the qualified import path. I figure I will add an `IsLocalImport` call and the documentation needs to updated to be more precise.

jonjohnsonjr: Yep, you got it. Thanks!

README.md
HeavyWombat: Sorry it took so long, but I could not find cycles to work on this in the last week. The template support works for environment variables now. I find that solution much nicer now. Eventually, we can add even more entries to the data map for the template rendering.

pkg/build/gobuild.go
HeavyWombat: Quick question: I sporked the required pieces into the `ko` repository. Which boilerplate would be applied to the file?

jonjohnsonjr: Keep theirs, we've done something similar with k8s stuff: https://github.com/google/go-containerregistry/blob/main/internal/retry/wait/kubernetes_apimachinery_wait.go

pkg/build/options.go
```diff
@@ -45,6 +46,15 @@ func WithDisabledOptimizations() Option {
 	}
 }
 
+// WithGoReleaserBuildConfig is a functional option for providing build
+// settings based on a GoReleaser style configuration.
+func WithGoReleaserBuildConfig(buildCfgOverrides map[string]config.Build) Option {
```
jonjohnsonjr, line 51: I think this is impossible for external dependents to use because `config.Build` is `internal/`. (cc @halvards)

We may want to have a `pkg/config` (or something) that exposes the usable subset of goreleaser's `config.Build`, then do some translation to talk advantage of their stuff.

HeavyWombat: Good point. Thanks. I moved it to `pkg/config`. However, I am wondering whether maybe something like `pkg/ref/goreleaser/config` would be better, because `pkg/config` is pretty generic.

HeavyWombat: @jonjohnsonjr Did you have to think about the name of package?

jonjohnsonjr: Sorry, missed this comment. That's a good point.

My intuition is that this would only affect build config, so it might make more sense to flip the name and call this `build.Config` instead of `config.Build`?

I'm indifferent re: config vs configuration, but this seems reasonable enough:

```
builder.Build(..., build.WithConfig(build.Config{}))
```

We already have `build.Result` so it fits in well. If we want to have a separate configuration for e.g. the publisher stuff (currently set by KO_DOCKER_REPO), that could be mirrored as `publisher.Config`. I think scoping the configuration to the relevant package is probably best, but I'd be interested in other ideas.

Sorry for the back and forth on this :P

HeavyWombat: No problem, not at all. I am here to learn. Naming things can be hard and `build.Config` makes perfect sense while also improves the readability of the code. I will change that right away.

pkg/build/options.go
```diff
@@ -45,6 +45,15 @@ func WithDisabledOptimizations() Option {
 	}
 }
 
+// WithConfig is a functional option for providing build settings based on a
+// GoReleaser Build influenced configuration.
+func WithConfig(buildCfgOverrides map[string]Config) Option {
```
jonjohnsonjr, line 50: Document what the key is here.

pkg/build/config.go
```diff
@@ -0,0 +1,80 @@
...
+		*a = strings.Fields(flagstr)
+	} else {
+		*a = flags
+	}
+	return nil
+}
+
+// Config contains the build configuration section. The name was changed from
+// the original GoReleaser name to match better with the ko namings.
+type Config struct {
```
jonjohnsonjr, line 61: I would comment out anything that we don't currently respect and put a big fat TODO comment at the top, maybe pointing to an existing issue or a new issue to track progress.

I believe we only respect `Ldflags` and `Flags` currently?

pkg/build/config.go
```diff
@@ -0,0 +1,80 @@
...
+	ID           string      `yaml:",omitempty"`
+	Goos         []string    `yaml:",omitempty"`
+	Goarch       []string    `yaml:",omitempty"`
+	Goarm        []string    `yaml:",omitempty"`
+	Gomips       []string    `yaml:",omitempty"`
+	Targets      []string    `yaml:",omitempty"`
+	Dir          string      `yaml:",omitempty"`
```
jonjohnsonjr, line 68: @halvards we might be able to use `Dir` from this instead of https://github.com/google/ko/pull/365

pkg/build/config.go
HeavyWombat: Yes, it is only `Ldflags` and `Flags` with this PR. In the project I am currently working on, I see the potential that maybe something like `ModTimestamp` and `Env` would be interesting, too. I would like to tackle that with subsequent PRs.

jonjohnsonjr:
I think we'll want to combine the changes from https://github.com/google/ko/pull/365 with this PR.

This PR uses `build.Config.Dir` with `build.Config.Main` to qualify local import paths, but we should probably also do what #365 is doing and use `Dir` as the working directory for the `go build`. I'm also not sure how this interacts with `kodata` -- we might want to check on that.

HeavyWombat:
> I think we'll want to combine the changes from #365 with this PR.
> 
> This PR uses `build.Config.Dir` with `build.Config.Main` to qualify local import paths, but we should probably also do what #365 is doing and use `Dir` as the working directory for the `go build`. I'm also not sure how this interacts with `kodata` -- we might want to check on that.

Changing the code to make sure that `go build` uses the directory specified in `build.Config.Dir` as the working directory required a bit of a rewrite of some signatures and the flow. I pushed the idea I have for that and this also placed the `kodata` directory under `/var/run/ko`. However, it only does that for the `go build`. In #365, the working directory change is also done for `go list` and `go env`. Especially, for `go list` I am not sure if changing the working directory is desired in all cases, or at least I am not sure how it will work if the modules file is at the project root directory while the respective working directory would be a sub-directory. If we want to have the working directory fix for all `go` commands, I was wondering whether it would be easier to introduce the working directory as a `Value` into the Context that is available in all functions already rather than adding a directory string to all function signatures.

jonjohnsonjr:
Apologies for taking a while to get around to looking at this. It's been a while since I got too deep into the `ko` codebase (kodebase?), so it's difficult to reason about everything affected by this change. I'm having a hard time really understanding the semantics of this config file... would you mind adding a comment to the `Main` and `Dir` fields describing _exactly_ what they control? It will help me juggle this and #365 better, I think.

> Especially, for go list I am not sure if changing the working directory is desired in all cases, or at least I am not sure how it will work if the modules file is at the project root directory while the respective working directory would be a sub-directory.

Interesting. This has been challenging in the past when dealing with go submodules within a repository. 

> I was wondering whether it would be easier to introduce the working directory as a Value into the Context that is available in all functions already rather than adding a directory string to all function signatures.

I think I prefer the simplicity of changing the function signatures. It makes it easier to review at least.

HeavyWombat:
> Apologies for taking a while to get around to looking at this. It's been a while since I got too deep into the `ko` codebase (kodebase?), so it's difficult to reason about everything affected by this change. I'm having a hard time really understanding the semantics of this config file... would you mind adding a comment to the `Main` and `Dir` fields describing _exactly_ what they control? It will help me juggle this and #365 better, I think.
> 
> > Especially, for go list I am not sure if changing the working directory is desired in all cases, or at least I am not sure how it will work if the modules file is at the project root directory while the respective working directory would be a sub-directory.
> 
> Interesting. This has been challenging in the past when dealing with go submodules within a repository.
> 
> > I was wondering whether it would be easier to introduce the working directory as a Value into the Context that is available in all functions already rather than adding a directory string to all function signatures.
> 
> I think I prefer the simplicity of changing the function signatures. It makes it easier to review at least.

@jonjohnsonjr I added comments to `Dir` and `Main` the way I understand them. This way it is more clear when looking at the code without requiring any context of GoReleaser. Also, I changed the config code to be able to properly test the permutations of `Dir` and `Main`. However, I do think that most users of this feature will only use `Main`.

jonjohnsonjr:
@imjasonh want to take a look? I think we're getting close to something that's actually not bad!

imjasonh (approved):
Looks great! Thanks for your hard work on this @HeavyWombat 

@jonjohnsonjr Anything else you want to see before we merge this? We should probably cut a release after so people can use it more easily.

jonjohnsonjr:
I want to experiment for a bit before cutting a release, I think.

HeavyWombat:
> I want to experiment for a bit before cutting a release, I think.

Thanks for looking into it. If there is anything I can help with, please let me know.

PR #342, opened 2021-04-20
WIP: Serialize image loads to kind

imjasonh:
This seemed to fix issues publishing multiple images to kind, which were cleared up when image loads were serialized.

### DIFF
--- a/pkg/publish/kind/write.go
+++ b/pkg/publish/kind/write.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
 	"golang.org/x/sync/errgroup"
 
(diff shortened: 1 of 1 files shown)


pkg/publish/kind/write.go
```diff
@@ -53,9 +54,15 @@ func Tag(src, dest name.Tag) error {
 	})
 }
 
+// writeMu serializes image imports to kind, to avoid locking situations
+// observed when publishing multiple images to KinD at the same time.
+var writeMu sync.Mutex
```
jonjohnsonjr, line 59: This works but package-level mutex is kinda gross.

We could do this here, given that we usually have one publisher that we hold (I think): https://github.com/google/ko/blob/75ab99123d2c6b830fc11c935f7d3a2a2d740256/pkg/publish/kind.go#L105

ImJasonH:
I'm not seeing the issue anymore after deleting and recreating my KinD cluster. This is weird. Keeping this PR open in WIP/draft mode to collect further data.

jonjohnsonjr:
Another workaround may be to use set `--jobs 1` which should overcome this?

PR #349, opened 2021-04-28
bump ggcr dep to v0.5.0

imjasonh:


jonjohnsonjr:
I'm okay with that breaking change in daemon, I think, because mutating package globals is a really gross thing to do. You should be able to rewrite those tests using `daemon.WithClient`.

pkg/publish/daemon.go
```diff
@@ -43,8 +44,12 @@ func NewDaemon(namer Namer, tags []string) Interface {
 	return &demon{namer, tags}
 }
 
+var defaultClient = func() (daemon.Client, error) {
+	return client.NewClientWithOpts(client.FromEnv)
```
jonjohnsonjr, line 48: I might have this be `func getOpts(ctx) { return WithContext(ctx) }` and override it to include `daemon.WithClient` for tests, so we don't have to import the docker client just to do what `daemon` does by default?

imjasonh: Dang that's nice.

jonjohnsonjr (approved):

PR #351, opened 2021-04-29
Add flag and PublishOption for destination repo

halvards:
This enables programmatically setting the destination image repository
when embedding ko's `publish` functionality in other tools.

See discussion in https://github.com/google/ko/pull/348 for additional context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,16 @@ to
 
 ## Choose Destination
 
-`ko` depends on an environment variable, `KO_DOCKER_REPO`, to identify where it
-should push images that it builds. Typically this will be a remote registry,
-e.g.:
+`ko` must be configured with a destination for where it should push images that
+it builds. You can configure this with either the `--docker-repo` flag or the
+`KO_DOCKER_REPO` environment variable. Typically, the value will be a remote
+registry, e.g.:
 
 - `KO_DOCKER_REPO=gcr.io/my-project`, or
-- `KO_DOCKER_REPO=my-dockerhub-user`
+- `--docker-repo=my-dockerhub-user`
+
+If both the flag and the environment variable are set, the flag value takes
+precedence.
 
 # Build an Image
 
--- a/README.md
+++ b/README.md
@@ -68,9 +68,6 @@ e.g.:
 - `KO_DOCKER_REPO=gcr.io/my-project`, or
 - `KO_DOCKER_REPO=my-dockerhub-user`
 
-If both the flag and the environment variable are set, the flag value takes
-precedence.
-
 # Build an Image
 
 `ko publish ./cmd/app` builds and pushes a container image, and prints the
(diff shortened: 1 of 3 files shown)


imjasonh:
Thanks for making this change. I think I'm convinced that grabbing the env var doesn't belong in `makePublisher`, but instead in an option like you're adding with this PR. 👍 

I'm still a bit less convinced that it therefore needs to be a separate flag. Can we compromise and make `AddPublishArg` grab the env var and set the option for now?

Then, to embed this in skaffold (or elsewhere) we can make `makePublisher` and any other method you need exported so you can call them from your code. That way you can still embed ko logic, and ko doesn't have to add a flag (...yet).

wdyt?

halvards:
Thanks for the quick reply! Yes, that SGTM, I'll go ahead and make the change.

The flag isn't required for embedding, I just added it for consistency, since all the other fields in `PublishOptions` are exposed as flags.

README.md
```diff
@@ -68,6 +68,9 @@ e.g.:
 - `KO_DOCKER_REPO=gcr.io/my-project`, or
 - `KO_DOCKER_REPO=my-dockerhub-user`
 
+If both the flag and the environment variable are set, the flag value takes
```
imjasonh, line 71: This change can be removed now.

pkg/commands/options/publish.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Google LLC All Rights Reserved.
+// Copyright 2021 Google LLC All Rights Reserved.
```
imjasonh, line 1: I'm not sure why the boilerplate check is complaining about this, but just to appease it can you use the `/* */` comment form? 🙄

pkg/commands/options/publish.go
```diff
@@ -25,6 +26,9 @@ import (
 
 // PublishOptions encapsulates options when publishing.
 type PublishOptions struct {
+	// DockerRepo configures the destination image repository
```
imjasonh, line 29: Could you add a reference to the KO_DOCKER_REPO env var in this comment? Something like, "In normal ko usage, this is populated with the value of $KO_DOCKER_REPO".

imjasonh (commented):
Lookin' good, just a few small things.

pkg/commands/options/publish.go
halvards: Done

pkg/commands/options/publish.go
halvards: Done. It doesn't match the other copyright headers, but 🤷

README.md
halvards: Done

imjasonh (approved):

jonjohnsonjr (approved):

PR #354, opened 2021-05-05
Annotate images with base image information

imjasonh:
~~Depends on https://github.com/google/go-containerregistry/pull/960~~

This annotates images (and images within indexes) built by `ko` with base image (or base image index) information necessary to be able to perform automatic image rebases, using nothing but information available in the image itself.


pkg/build/gobuild.go
```diff
@@ -596,6 +602,21 @@ func (g *gobuild) buildOne(ctx context.Context, s string, base v1.Image, platfor
...
+	if err != nil {
+		return nil, err
+	}
+
+	m := map[string]string{
+		baseRefAnnotation:    baseRef.Name(),
+		baseDigestAnnotation: baseDigest.String(),
+	}
+	log.Println("DEBUGGING ANNOTATIONS:", m) // TODO remove
+	withApp, err = mutate.Annotations(withApp, m)
```
jonjohnsonjr, line 615: Would you look at that!

pkg/build/gobuild.go
```diff
@@ -762,7 +778,7 @@ func (g *gobuild) buildAll(ctx context.Context, s string, base v1.ImageIndex) (v
 		if err != nil {
 			return nil, err
 		}
-		img, err := g.buildOne(ctx, s, base, desc.Platform)
+		img, err := g.buildOne(ctx, ref, baseRef, base, desc.Platform)
```
jonjohnsonjr, line 781: I don't think this is right.

`base.Digest()` will point to the image, but `baseRef` will point to the index, so any naive code that tries to diff for updates will always think it's changed.

imjasonh:
An example image built with the latest version of this code that annotates a base _index_ digest, if the base was an index: https://explore.ggcr.dev/?image=gcr.io%2Fimjasonh%2Ftest-46c4b272b3716c422d5ff6dfc7547fa9%40sha256%3Af6bd96de09aa407d8f922fefa85f8f775eafd6896677b737b8ea0b98a3f0c0e7

jonjohnsonjr:
> An example image

This is still using the old proposed key:

```
org.opencontainers.image.base.ref.name
```

New one is:

```
org.opencontainers.image.base.name
```

imjasonh:
You mentioned you'd like to also annotate the index in a multi-platform image so that it can also be rebased. That'll depend on https://github.com/google/go-containerregistry/pull/1082, and effectively on ggcr/`crane` gaining further support for rebasing multi-platform images in general, and then in detecting rebase annotations in indexes.

This works for now, for `--platform=<one-platform>` single-platform images, and I think it's a step forward in a good direction. We'll build upon it to layer on more support for multi-platform rebaseable images as part of the larger work.

jonjohnsonjr (approved):

PR #365, opened 2021-05-18
Enable setting the working dir for the go tool

halvards:
This change adds a `Dir` field to `options.BuildOptions`, but doesn't
expose this as a CLI flag. The default zero value means the current
working directory. The value is used as the working directory for
executing `go` tool commands.

When embedding ko in other tools, it is sometimes necessary to set the
working directory for executing the `go` tool, instead of assuming the
current process working directory.

An example of where this is required from Skaffold:
https://github.com/GoogleContainerTools/skaffold/tree/master/examples/microservices

In this example, the working directory doesn't contain either `go.mod`
or any Go files. The `skaffold.yaml` configuration file specifies
a `context` field for each image, which is the directory where the `go`
tool can find package information.

I considered updating the `buildContext` interface to wrap `go/build.Context`.
That'd avoid duplicating the `dir` field in `gobuild` and `buildContext`.
Instead of add that to this pull request, I made that change in another branch,
mainly to explore what that could look like:
https://github.com/halvards/ko/blob/base-dir-buildcontext/pkg/build/context.go#L17
PLMKWYT


jonjohnsonjr:
Can you explain the `QualifyImport` addition? Do we need to expose this? Or can we teach `IsSupportedReference` and `Build` to do this automagically?

halvards:
> Can you explain the `QualifyImport` addition?

`QualifyImport` wraps this logic in a function:
https://github.com/google/ko/blob/522c37c4e0ec9537068f16e8f1652f969377b62e/pkg/commands/publisher.go#L47-L56

And to resolve local "import paths" correctly, the function needs to know the current directory, from the new `dir` field on `gobuild`:
https://github.com/google/ko/blob/c9883379f5e168368aaa3416c27531a18e0260ea/pkg/build/gobuild.go#L257

> Do we need to expose this? Or can we teach `IsSupportedReference` and `Build` to do this automagically?

Not without changing the method signatures AFAICT. We could call a non-exported `qualifyImport()` from both `IsSupportedReference` and `Build`, but `Publish` also needs the resolved/qualified import path: https://github.com/google/ko/blob/c9883379f5e168368aaa3416c27531a18e0260ea/pkg/commands/publisher.go#L43

pkg/commands/options/build.go
```diff
@@ -24,7 +26,12 @@ import (
 type BuildOptions struct {
 	// BaseImage enables setting the default base image programmatically.
 	// If non-empty, this takes precedence over the value in `.ko.yaml`.
-	BaseImage            string
+	BaseImage string
+
+	// Dir allows for setting the working directory for invocations of the `go` tool.
+	// Empty string means the current working directory.
+	Dir string
```
jonjohnsonjr, line 33: Can we rename this to `WorkingDir` or `WorkingDirectory`? I think that would help with the name collision for https://github.com/google/ko/pull/365

halvards: Done. Also squashed the commits.

jonjohnsonjr (approved):

Issue #368, opened 2021-05-27
Feature Request: custom base image from local docker daemon


misterikkit:
It would be convenient to specify a `customBaseImage` from my local docker daemon. If I tag an image as `foo:1` then ko will search index.docker.io. If I tag as `ko.local/foo:1` then ko will try to dial `ko.local`.

I have a project in which several containers are built from a common base image, and the base image is also built in the same project. The current build flow is something like,
```sh
$ docker build -t base:0.1 -f base/Dockerfile .
$ docker build -t foo:0.1 -f foo/Dockerfile .
$ cat foo/Dockerfile
FROM base:0.1
...
```

jonjohnsonjr:
https://github.com/google/ko/pull/371 just assumes `ko.local` things are in the daemon, not any other domain (or images without a domain), would that work for you?

misterikkit:
@jonjohnsonjr Yeah I think that would solve the use case I was hitting. At this point, I am just experimenting with ko. Integrating it into my CI/build pipelines would involve propagating that `ko.local` prefix to many places.

PR #371, opened 2021-05-28
Look for ko.local in the daemon

jonjohnsonjr:
Fixes https://github.com/google/ko/issues/368

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,12 @@ identified by `KO_DOCKER_REPO`.
 `ko` can also publish images to a local Docker daemon, if available, by setting
 `KO_DOCKER_REPO=ko.local`, or by passing the `--local` (`-L`) flag.
 
+Locally-published images can be used as a base image for other `ko` images:
+
+```yaml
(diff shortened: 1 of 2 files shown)


imjasonh:
Can you also add this to the docs? We might want a quick example of building a base image locally then ko-publishing on top of it, and the `ko.local` convention would be useful to spell out.

jonjohnsonjr:
> a quick example of ...

We might want a directory with examples to avoid polluting the README? Would also be nice for getting started. Too lazy to do that for this PR :)

misterikkit (approved):
Are there other places where a ref to `ko.local` might not resolve to the daemon? Eventually, it would be nice to have a consistent "ko.local means daemon" behavior across all surfaces.

jonjohnsonjr:
I think this covers everything. We pull base images and push built images, and the base image side probably doesn't receive as much love because (usually) the distroless default base is good enough.

jonjohnsonjr:
@imjasonh ptal

imjasonh (approved):

PR #374, opened 2021-06-09
Build working Windows container images

### LINKED ISSUE #361: Ko-built Windows images aren't pullable/runnable
Reported in Tekton's #windows Slack channel: https://tektoncd.slack.com/archives/C020SPF0B7Y/p1620277036006500 -- copied here for posterity and to aid future discussion.

tl;dr: Images that Ko builds for Windows don't appear to be pullable, possibly due to `KO_DATA_PATH` and the `/var/run/ko` directory in the tar file.

It sounds like we could benefit from some Windows CI at least, to ensure that Windows images built with `ko` are runnable. We could have a GitHub Actions Windows runner pull and run a ko-built image, or after we cut a release with #339 we could build and run ko within CI to detect breakages before merging pull requests.

The specific issue might require us to have Windows-specific logic in gobuild.go, to change where kodata ends up in the image.

---

@aiden-deloryn

I have been doing some testing with a Windows node and have encountered a problem which prevents the entrypoint image from running (and as a result the TaskRun's pod won't initialise). Describing the pod reveals the following error under Events:

```
Failed to pull image "10.115.11.41:30002/tekton/entrypoint-bff0a22da108bc2f16c818c97641a296@sha256:1f900110f0103c208200d0be7e5614836d036845ac934d093192826dd2b94d84": rpc error: code = Unknown desc = failed to register layer: re-exec error: exit status 1: output: mkdir \\?\C:\ProgramData\docker\tmp\hcs523070999\var\run\ko: expected relative path
```

After inspecting the container image, I can see that the path `/var/run/ko` is injected into the image as an environment variable called `KO_DATA_PATH`. According to the Ko documentation this path is used to bundle static assets from a `kodata` sub-directory into the image's `KO_DATA_PATH` directory. I'm not sure why this extra path is being appended to the beginning - `\\?\C:\ProgramData\docker\tmp\hcs523070999`.

As this problem occurs when processing an image layer it can also be reproduced by running `docker pull <image>`, after building the Windows image with `ko publish ./cmd/entrypoint --platform=windows/amd64`. Note: for testing purposes, I'm using a Ko base image override of `github.com/tektoncd/pipeline/cmd/entrypoint: mcr.microsoft.com/windows/servercore:ltsc2019`

I haven't used Ko previously so I'm not all that familiar with it. I'm also not sure if this is a Docker issue or a Ko issue. My Ko version is 0.8.2 and Windows Docker runtime version 20.10.4 (Mirantis). I can't inspect the image using Docker because I cannot pull it. It seems either Ko is producing an incorrect image layer, or Docker is interpreting/converting the path incorrectly. Is there any way I can troubleshoot Ko further to see what's going on?

---

@aiden-deloryn

Your suggestion does sound plausible to me. Perhaps a relative path could be the solution (as the error seems to suggest). Maybe just removing the root path separator would be enough? (i.e.  `var/run/ko`)
If we need a more specific location, I've found some directories in nanoserver (smallest image) which might be suitable:
C:\ProgramData\ >>> for application data that is not user-specific
C:\Windows\Temp\ >>> for temporary files
C:\Users\ContainerUser\ >>> user's home directory
C:\Users\ContainerUser\AppData\Local\ >>> user-specific application data
C:\Users\ContainerUser\AppData\Local\Temp\ >>> user's temp directory

---

@adamrehn:

Looking at the implementation of `tarAddDirectories()`, it looks like the [intended file permissions](https://github.com/google/ko/blob/v0.8.2/pkg/build/gobuild.go#385) for the kodata directory are world readable and executable, but not writable? If static data files don't need to be writable then C:\ProgramData\ko might be the ideal candidate, since the primary benefit of opting for a temporary or user-specific directory would be that the resulting files can be modified. (Avoiding user-specific directories also sidesteps any issues around determining whether a given Windows container image uses the `ContainerUser` or `ContainerAdministrator` user by default.)

imjasonh:
Fixes #361 

The main discoveries while implementing this have been:

- Windows containers expect contents to be in a `Files/` root directory in the tar archive
    - ...and for a `Hives/` root directory to be present, though it can be empty
- Executable entries in the tar archive must have a `PAXRecords` value to be executable at runtime

Even after this work lands, users building images will have to make sure they use a base image that provides the exact Windows OS version that they intend to run the image on. Users wishing to produce a multi-OS image with `ko` will probably also have to stitch together a suitable base image from nanoserver etc., and whatever Linux base image they want to use (distroless, alpine, etc.), since AFAIK suitable multi-OS base images aren't exactly prevalent in the wild.


pkg/build/gobuild.go
```diff
@@ -639,14 +662,14 @@ func updatePath(cf *v1.ConfigFile) {
 		}
 		key, value := parts[0], parts[1]
 		if key == "PATH" {
-			value = fmt.Sprintf("%s:%s", value, appDir)
+			value = fmt.Sprintf("%s:%s", value, appPath)
```
micahyoung, line 665: This would need a Windows-specific implementation to convert `appPath` to be suitable for `PATH`. In a quick test, generating a Windows image as described in the new README changes, I saw the generated `PATH` as this:

```bash
crane config <image> | jq .config.Env
## output
[
  "PATH=Files/ko-app/ko.exe",
  "KO_DATA_PATH=/var/run/ko"
]
```
Instead it would need to be something like:
```
[
  "PATH=c:\\ko-app\\ko.exe",
  "KO_DATA_PATH=/var/run/ko"
]
```
Note: While `PATH` would have to be Windows syntax, `KO_DATA_PATH` may not *need* to change since Go-generated Windows binaries seem to accept most POSIX-style paths, (though it would probably be more valid to do so).

pkg/build/gobuild.go
```diff
@@ -604,8 +623,12 @@ func (g *gobuild) buildOne(ctx context.Context, s string, base v1.Image, platfor
 	}
 
 	cfg = cfg.DeepCopy()
+	cfg.Config.Cmd = nil
 	cfg.Config.Entrypoint = []string{appPath}
-	updatePath(cfg)
+	if platform.OS == "windows" {
+		cfg.Config.Entrypoint = []string{`C:\ko-app\` + appFilename(ref.Path()) + ".exe"}
+	}
```
micahyoung, line 630: I'm wondering if this could be omitted if the `PATH` for Windows is fixed in the comment below?

micahyoung (changes requested):
This is great! There's just saw a small issue with the Windows `PATH` value that might simply some things.

README.md
```diff
@@ -410,20 +410,21 @@ export KO_DATA_DATE_EPOCH=$(git log -1 --format='%ct')
...
 
 For example, to build a Windows container image for `ko`, from within this repo:
 
 ```
 KO_DEFAULTBASEIMAGE=mcr.microsoft.com/windows/nanoserver:1809 ko publish ./ --platform=windows/amd64
 ```
 
-## Can I optimize images for [eStargz support](https://github.com/containerd/stargz-snapshotter/blob/v0.2.0/docs/stargz-estargz.md)?
+### Known issues 🐛
+- Symlinks in `kodata` are ignored when building Windows images; only regular files and directories will be included in the Windows image.
```
jonjohnsonjr, line 425: Nit: newline after header.

pkg/publish/daemon.go
```diff
@@ -145,7 +145,8 @@ func (d *demon) Publish(ctx context.Context, br build.Result, s string) (name.Re
 	}
 
 	log.Printf("Loading %v", digestTag)
-	if _, err := daemon.Write(digestTag, img, d.getOpts(ctx)...); err != nil {
+	if resp, err := daemon.Write(digestTag, img, d.getOpts(ctx)...); err != nil {
+		log.Println("daemon.Write response: ", resp)
```
jonjohnsonjr, line 149: Even if `err == nil`, there's sometimes interesting stuff in resp. Should we care? Wasn't your debugging purgatory one such situation?

imjasonh: Yeah 😒 

The problem is that resp is pretty noisy in the common case where it actually works and doesn't return an err (it logs write progress bars and some other stuff), so I wouldn't want to log it all the time. Maybe only if `--verbose`, but then I'd have to add a `--verbose` flag.

I'm going to tempt fate and say that hopefully this fairly arcane situation is one of the only ones that cause resp to be useful in the absence of err, and try to remember that sometimes it's the case if I'm unlucky enough to find myself debugging arcane `daemon.Write` issues in the future.

jonjohnsonjr (approved):

PR #377, opened 2021-06-28
Update "go install" instructions

xuhdev:
I got the following error when following the original instructions:

    go install: version is required when current directory is not in a module
            Try 'go install github.com/google/ko@latest' to install the latest version

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ brew install ko
 With Go 1.16+, build and install the latest released version:
 
 ```
-go install github.com/google/ko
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

imjasonh:
Sorry for the long delay on this PR. We had an issue with kind e2e tests that was resolved in #184 -- if you rebase you should pick it up, and tests should pass, and we can merge this. Otherwise I'd be happy to make the change myself.

imjasonh (approved):

PR #379, opened 2021-07-04
make ko environment variables aware while resolving YAML files

developer-guy:

Let's assume that I have a YAML file that includes some environment variables, and I want to resolve my YAML file using ko's resolve sub-command.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: falco-event-listener
  namespace: falco
  labels:
    app: falco-event-listener
spec:
  containers:
  - image: ko://github.com/developer-guy/falco-event-listener
    name: falco-event-listener
    args:
      - "--notify-url"
      - "http://receiver.flux-system/$(WEBHOOK_URL)"
    env:
      - name: WEBHOOK_URL
        value: $WEBHOOK_URL
```
As you can see, I have some environment variables defined in the YAML file. Before running the following command I defined the `WEBHOOK_URL` environment variable in my shell session:

```shell
$ export WEBHOOK_URL="somerandomurl"
$ ko resolve -f listener.yaml
2021/07/04 18:24:59 NOTICE!
ko and go have mismatched GOROOT:
    go/build.Default.GOROOT = "go"
    $(go env GOROOT) = "/usr/local/Cellar/go/1.16.5/libexec"

Inferring GOROOT="/usr/local/Cellar/go/1.16.5/libexec"

Run this to remove this warning:
    export GOROOT=$(go env GOROOT)

For more information see:
    https://github.com/google/ko/issues/106
apiVersion: v1
kind: Pod
metadata:
  name: falco-event-listener
  namespace: falco
  labels:
    app: falco-event-listener
spec:
  containers:
    - image: gcr.io/developerguy-311909/falco-event-listener-355c83b3413baa4dc0f62f21c3a83388@sha256:a21798867709e562b969ea16307c03e4a7d1a45f2dc58e9a32e4732984653a10
      name: falco-event-listener
      args:
        - "--notify-url"
        - "http://receiver.flux-system/$(WEBHOOK_URL)"
      env:
        - name: WEBHOOK_URL
          value: $WEBHOOK_URL
```
As you can see from the above output, the variable remains the same, so nothing changed. This PR aims to solve this problem.

> ko version: 0.83
> environment: Darwin xxxx 19.6.0 Darwin Kernel Version 19.6.0: Thu May  6 00:48:39 PDT 2021; root:xnu-6153.141.33~1/RELEASE_X86_64 x86_64

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -432,7 +432,7 @@ func resolveFile(
 	// The loop is to support multi-document yaml files.
 	// This is handled by using a yaml.Decoder and reading objects until io.EOF, see:
 	// https://godoc.org/gopkg.in/yaml.v3#Decoder.Decode
-	decoder := yaml.NewDecoder(bytes.NewBuffer(b))
+	decoder := yaml.NewDecoder(bytes.NewBuffer([]byte(os.ExpandEnv(string(b)))))
 	for {
 		var doc yaml.Node
 		if err := decoder.Decode(&doc); err != nil {


imjasonh:
Thanks for proposing this change! I'm frankly surprised it's as easy to support as it is in this PR! 😮

However, I'm a little worried that this is an initial step toward `ko` being used for general-purpose YAML templating, which is not in its scope today, and which can become a fairly complex task.

Detecting and replacing the string `$WEBHOOK_URL` with the value of the environment variable `WEBHOOK_URL` isn't terribly difficult, as you've seen, but when `ko` supports that case, I worry a user will inevitably want to process strings like `${WEBHOOK_URL}` too (which this PR would support), or `${WEBHOOK_URL:-default}` (i.e., templates with default values), or express loops to consolidate repeated blocks of config, or conditional expressions, and so on. All of these are worthwhile features for a general-purpose templating engine, but could greatly complicate `ko`, which today is relatively single-purpose.

The good news is that `ko` [already works](https://github.com/google/ko#does-ko-work-with-kustomize) with other templating solutions like [Kustomize](https://kustomize.io/), because it can process YAML from stdin, or emit YAML to stdout to be processed by other tools.

So a user who wants to replace `$WEBHOOK_URL` in their YAML and have `ko` process it can do so already with something like [`envsubst`](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html), for example:

```
envsubst < input.yaml | ko resolve -f - > output.yaml
```

Or, `ko` can process the YAML first, and pass it to envsubst:

```
ko resolve -f input.yaml | envsubst > output.yaml
```

developer-guy:
Omg @imjasonh thank you for such an amazing explanation, yeah I got you and you are right. It would better to combine ko with some kind of templating tools such as `envsubst` and `Kustomize` like you said to do such things. 

I'm closing this PR. 🙋🏻‍♂️🙏

PR #397, opened 2021-07-26
Silence usage report on internal errors

evankanderson:
See https://github.com/spf13/cobra/issues/340 for why this isn't the default (answer: historically, it wasn't the default, and spf13 is trying not to break people).


jonjohnsonjr (approved):

imjasonh:
Thank you! 🙏

PR #398, opened 2021-07-26
Make --insecure-registry work with TLS registries whose certs we can't verify.

### LINKED ISSUE #396: Can't figure out flags for insecure registry
I'm using `ko` version `v0.8.3`, and I'm probably an idiot, but I can't seem to get either `--insecure-registry` flag from `ko resolve` or the `--insecure-registry` and `--insecure-skip-tls-verify` flags working against a private harbor registry.

The harbor registry has a SNI cert for `192.168.10.12`, and I've set `KO_DOCKER_REPO=192.168.10.12`, and I've set up authorization via `docker login` (and added the registry to `insecure-registries` in the docker JSON config).

It looks like the following _should_ be the correct usage:

```shell
$ ko apply --insecure-registry -f config/contour
...
Publishing 192.168.10.12/dev/contour-ff924afcef2cd0177014d053ee4b167d:latest
....
error during command execution:error processing import paths in "config/contour/external.yaml": error resolving image references: Get "https://192.168.10.12/v2/": x509: certificate signed by unknown authority; Get "https://192.168.10.12:443/v2/": x509: certificate signed by unknown authority
```

This just produces an error:

```shell
$ ko --insecure-registry apply -f config/contour
Error: unknown flag: --insecure-registry
```

The `--insecure-skip-tls-verify` flag might be a kubectl flag, though it's not documented _what_ it connects to:

```shell
$ ko apply -f config/contour --insecure-skip-tls-verify
error: Unexpected args: [true]
...
Error: error executing 'kubectl apply': exit status 1
...
      --insecure-skip-tls-verify       If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure
```
(This command doesn't seem to end up trying to publish an image before it fails)

evankanderson:
Fixes #396

Note that this is technically breaking if people are using --insecure-registry and replacing the RoundTripper with something that's not an `http.Transport`. I decided to blow up loudly, rather than let it pass, but I could be convinced to go the other way.

Verified on harbor 2.2.3 with a self-signed cert with a SAN of 192.168.10.12.

### DIFF
--- a/pkg/publish/options.go
+++ b/pkg/publish/options.go
@@ -16,7 +16,6 @@ package publish
 
 import (
 	"crypto/tls"
-	"fmt"
 	"log"
 	"net/http"
 	"path"
@@ -109,7 +108,7 @@ func Insecure(b bool) Option {
 		i.insecure = b
 		t, ok := i.t.(*http.Transport)
 		if !ok {
-			return fmt.Errorf("unable to configure insecure roundtripper (not HTTP)")
+			return nil
 		}
 		t = t.Clone()
 		if t.TLSClientConfig == nil {
--- a/pkg/publish/options.go
+++ b/pkg/publish/options.go
@@ -15,6 +15,8 @@
 package publish
 
 import (
+	"crypto/tls"
+	"fmt"
 	"log"
 	"net/http"
 	"path"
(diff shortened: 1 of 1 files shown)


pkg/publish/options.go
```diff
@@ -105,6 +107,17 @@ func WithTagOnly(tagOnly bool) Option {
 func Insecure(b bool) Option {
 	return func(i *defaultOpener) error {
 		i.insecure = b
+		t, ok := i.t.(*http.Transport)
+		if !ok {
+			return fmt.Errorf("unable to configure insecure roundtripper (not HTTP)")
```
jonjohnsonjr, line 112: I don't think it makes sense to error here.

If we can't override the TLSClientConfig, then in the case where we needed to, it will fail anyway, and in the case where we _didn't_ need to, it fails even when it shouldn't.

evankanderson: I've switched this -- do you think I should log anything here for the debugging case?

jonjohnsonjr: Logging sounds reasonable, but also the only time anyone is going to hit this is if they are using `ko` as a library _and_ supplying their own transport, so it might not really matter.

jonjohnsonjr (approved):

PR #405, opened 2021-07-30
Run integration_test.sh with Go 1.16

imjasonh:
I think this might break for 1.16...

(Also simplify the workflow yaml; the only relevant change is to the `go-version` matrix values)

### DIFF
--- a/.github/workflows/modules-integration-test.yaml
+++ b/.github/workflows/modules-integration-test.yaml
@@ -5,28 +5,19 @@ on:
     branches: ['main']
 
 jobs:
-
   test:
     name: Module Tests
     strategy:
       matrix:
-        go-version: [1.14.x, 1.15.x]
-        platform: [ubuntu-latest]
-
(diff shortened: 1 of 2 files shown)


halvards:
I have to set `export GO111MODULE=off` before I can run `integration_test.sh` locally. I was using Go 1.16, now Go 1.17.

The default changed from `auto` to `on` in 1.16. 1.17 was supposed to ignore `GO111MODULE`, but it still seems to respect it, and I didn't see any mention in the 1.17 release notes.

https://go.dev/blog/go116-module-changes

imjasonh:
Yeah I think we'll have to have the test detect the go version and change it's behavior to cover all cases.

We can probably drop 1.14, I don't think any relevant behavior changed between 1.14 and 1.15.

mattmoor (approved):

imjasonh:
Closing in favor of #435

PR #408, opened 2021-08-04
Don't set image.base.name if base is specified by digest

imjasonh:
Based on conversation in https://github.com/google/go-containerregistry/pull/1098

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -702,7 +702,7 @@ func (g *gobuild) configForImportPath(ip string) Config {
 	return config
 }
 
-func (g *gobuild) buildOne(ctx context.Context, refStr string, baseRef name.Reference, baseDigest v1.Hash, base v1.Image, platform *v1.Platform) (v1.Image, error) {
+func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, platform *v1.Platform) (v1.Image, error) {
 	ref := newRef(refStr)
 
 	cf, err := base.ConfigFile()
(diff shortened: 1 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -928,8 +932,12 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseRef name.Referen
 	// Annotate the index with base image information, if the index is an OCI image index.
 	// (Docker manifest lists don't support annotations)
 	if baseType == types.OCIImageIndex {
+		var baseName string
+		if _, ok := baseRef.(name.Tag); ok {
+			baseName = baseRef.Name()
+		}
 		idx = mutate.Annotations(idx, map[string]string{
-			specsv1.AnnotationBaseImageName:   baseRef.Name(),
+			specsv1.AnnotationBaseImageName:   baseName,
```
jonjohnsonjr, line 940: Won't this put an empty string in the map?

imjasonh: Yeah, I'm doing the same thing in https://github.com/google/go-containerregistry/pull/1098 but I can make both just omit the key entirely.

imjasonh: done

jonjohnsonjr:
Instead of doing this in two places, you could capture the resulting `build.Result` from this block (as a separate function): https://github.com/google/ko/blob/4ff83080866f8acd0aba662c628eb69b8f9b940d/pkg/build/gobuild.go#L866-L881

And then just `mutate.Annotations` it in one place. 

This is possible because `build.Result`: https://github.com/google/ko/blob/a85cfeb2041e2635735bf35eca59ab82eed8ad14/pkg/build/build.go#L47-L53

Is compatible with `mutate.Annotatable`: https://github.com/google/go-containerregistry/blob/a65a0a616b77e0460c5b4b999d3afc346e2d21af/pkg/v1/mutate/mutate.go#L119-L121

And publishers will have to do the casting themselves downstream.

imjasonh:
Man that's nice. Good call 👍

jonjohnsonjr (approved):

PR #413, opened 2021-08-10
Remove dependency on cli-runtime

### LINKED ISSUE #317: Remove k8s deps if possible
I added these [forever ago](https://github.com/google/go-containerregistry/pull/297) to get kubectl flags, but it's been a source of endless trouble...

See https://github.com/google/ko/issues/315#issuecomment-783718564 for a recent issue.

imjasonh:
Fixes #317 

This change effectively forks the parts of genericclioptions that we use (i.e., global kubectl flag definitions) into an `internal` package, where flags are documented as deprecated. It severs our dependency on cli-runtime, and promotes an indirect dep on client-go to a direct one, for now. In a future change, we can remove this package too.

```
$ go run ./ create -f test --namespace=foo
2021/08/09 21:16:06 NOTICE!
Passing kubectl global flags to ko directly is deprecated.

Instead of passing:
    ko create ... --namespace=foo

Pass kubectl global flags separated by "--":
    ko create ... -- --namespace=foo

For more information see:
    https://github.com/google/ko/issues/317
```

If flags are passed before and after `--`, the one after the `--` is used:
```
$ go run ./ create -f test --namespace=foo -- --namespace=bar
...
Error from server (NotFound): error when creating "STDIN": namespaces "bar" not found
```

Flags after `--` aren't parsed at all by `ko`, and are just passed directly to `kubectl`:

```
$ go run ./ create -f test -- --unknown=flag
...
Error: unknown flag: --unknown
See 'kubectl apply --help' for usage.
```


internal/k8sflags.go
```diff
@@ -0,0 +1,129 @@
...
+
+import (
+	"path/filepath"
+	"strings"
+
+	"github.com/spf13/pflag"
+	"k8s.io/client-go/util/homedir"
+)
+
+var defaultCacheDir = filepath.Join(homedir.HomeDir(), ".kube", "cache")
```
jonjohnsonjr, line 25: If we don't have any of our own defaults, and only pass in non-zero flags, won't kubectl infer the default anyway? This would let us drop the client-go dep.

imjasonh: Excellent idea, done.

jonjohnsonjr (approved):

PR #414, opened 2021-08-10
Generate Markdown docs

imjasonh:
This is largely copied from similar work in go-containerregistry

This required moving the Root command definition out of main() into a
place where it could be referenced from the gendoc tooling.

Rendered: https://github.com/imjasonh/ko/blob/docs/doc/ko.md


imjasonh:
Two docgen diff issues:

1. `--jobs` defaults to `GOMAXPROCS`, which will be different for different users

```
-  -j, --jobs int                       The maximum number of concurrent builds (default 2)
+  -j, --jobs int                       The maximum number of concurrent builds (default 8)
```

2. `--cache-dir` (inherited from `kubectl`) will differ based on the user's homedir:

```
-      --cache-dir string               Default cache directory (default "/home/runner/.kube/cache")
+      --cache-dir string               Default cache directory (default "/Users/jasonhall/.kube/cache")
```

Sort of inclined to make both of these defined as zero values for the flag default, then interpret `--jobs=0` as `--jobs=GOMAXPROCS` and `--cache-dir` as `--cache-dir=$HOME/.kube/cache` only at runtime.

jonjohnsonjr:
> --jobs defaults to GOMAXPROCS, which will be different for different users

What if we did something stupid like replace `The maximum number of concurrent builds (default 2)` with `The maximum number of concurrent builds (default GOMAXPROCS)` via sed?

imjasonh:
> What if we did something stupid like replace `The maximum number of concurrent builds (default 2)` with `The maximum number of concurrent builds (default GOMAXPROCS)` via sed?

Defaulting -j at runtime seems more robust than relying on sed. See https://github.com/google/ko/pull/414/commits/903d3f4cdedf7fc63436a4e3bef60c34ff812da1

PR #419, opened 2021-08-18
Enable setting environment variables in .ko.yaml

halvards:
Matches the GoReleaser format.

Related: #340

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,8 @@ configuration section in your `.ko.yaml`.
 builds:
 - id: foo
   main: ./foobar/foo
+  env:
+  - GOPRIVATE=git.internal.example.com,source.developers.google.com
   flags:
   - -tags
   - netgo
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -144,6 +146,8 @@ builds:
   - -X main.version={{.Env.VERSION}}
 - id: bar
   main: ./foobar/bar/main.go
+  env:
+  - CGO_ENABLED=1
```
imjasonh, line 150: We don't tend to support cgo well, can we use another example here instead?

halvards: Good point, changed to `GOCACHE`.

halvards:
Should I commit the changes in `doc/` too? The only changes are the dates in the `Auto generated` comments on each file, there are no changes to the content.

jonjohnsonjr (dismissed):

jonjohnsonjr:
I think this just needs a rebase.

halvards:
Rebased, but Travis CI complains about the s390x build. Other archs seem fine, and the other PR has the same problem.

jonjohnsonjr:
> Rebased, but Travis CI complains about the s390x build. Other archs seem fine, and the other PR has the same problem.

Frustrating :/ just going to merge this anyway but we should probably just drop travis.

PR #420, opened 2021-08-19
Ignore cobra auto-generated comments with dates in doc diff

halvards:
spf13/cobra inserts comments with dates in generated Markdown files.

This change makes the presubmit check ignore those comments when verifying that ko's Markdown docs are up-to-date.

Also fixes some `shellcheck` warnings.


imjasonh (approved):
Thanks!

Issue #357, opened 2021-05-07
Signing built images


imjasonh:
Users can sign images produced with `ko publish` using tools like [`cosign`](https://github.com/sigstore/cosign).

For example:

```
$ cosign sign -key cosign.key $(ko publish ./)
```

`ko resolve` produces potentially many images, which makes this a bit harder. You could `ko resolve` then scan the resulting YAML for `ko`-built image references and sign all of those with some bash magic, but 🤮 .

Would it be useful to have a `ko resolve --sign cosign.key` flag that used the provided key to sign all images built during `ko resolve`?

`ko publish --sign cosign.key` could also be a convenience alias for, effectively, `cosign sign $(ko publish)`, which wouldn't require users to have `cosign` installed.

@dlorenc good idea? bad idea?

dlorenc:
> @dlorenc good idea? bad idea?

Great idea!

mattmoor:
+1 to native integration here.

Generally my preference would be to bias towards the ephemeral key route and use the same ephemeral key to sign all of the images produced (we should probably also start to think about incorporating a variety of opinionated attributes based on things we know about the build).

re: many images

something like `mink` produced ~25 images at its peak, for ~5 architectures, so 1 key vs ~150 keys (25*5 + 25) is probably a better way to go 🤣

mattmoor:
This PR includes a technique I have been using to incorporate keyless signing into some ko-like tools: https://github.com/sigstore/cosign/pull/647

Basically it only kicks in when `COSIGN_EXPERIMENTAL=true` for now, which is similar to how we've gated certain features with `GGCR_EXPERIMENTAL_*`, and this feels like an extremely lightweight way to integrate things.

We sacrifice some amount of configurability, but my inclination here would be to probably adjust how `cosign` itself works to use `envconfig` as a way to support a standard scheme for env-var fallback for configuration (no cobra, so no viper).

imjasonh:
+1 to `KO_EXPERIMENTAL_SOMETHING` to iterate on this. How do we imagine this looking when it graduates from an experiment? We don't have a good example of doing that in GGCR as far as I know.

Ideally we'd end in a state where things are signed by default using ambient tokens if available, which can be disabled with a flag or config. Do we want to support signing with explicit keys? Or just leave that for cosign?

mattmoor:
I think the first order of business will be to get the distroless images signed against Fulcio.

[...]

Issue #356, opened 2021-05-07
Verify base image signatures


imjasonh:
If `ko` is instructed to build an image based on another image, it would be nice to verify signatures on those base images.

It just so happens the default base image, [distroless](https://github.com/googlecontainertools/distroless), is signed. 🎉 

- What should the CLI surface look like for enabling/disabling signing base images?
- Should we start by only signing the default distroless base image?
- How should we get the keys to verify the signatures?
- How should users configure the keys to verify arbitrary images' signatures?

cc @dlorenc

mattmoor:
I'd propose a similar two-release strategy to things we've done previously (e.g. requiring `ko://`):
1. Start to verify things by default, be noisy when they are NOT (link to how), and have an option to fail,
2. Require verification by default, fail closed, and have an option to allow unsigned images.

imjasonh:
I don't think there are enough signed base images in the wild yet to justify failing closed. A big scary warning sgtm though.

Since we control the default and the default is signed, I do think it's safe to fail closed if we detect distroless isn't signed.

mattmoor:
@imjasonh Yeah 2 doesn't have to be the next release, but we should indicate in our scary warning that this is the direction we intend to head.  We can always wake up smarter and walk that back, but the more people we can scare into taking action the better IMO

mattmoor:
To pontificate a bit.  At least for me, `ko` and other "last mile" build tools exist to try to realize the best practices (e.g. nonroot, multi-arch, minimal base).

While the breaking nature of this sort of against the idea of back-compatibility, I think the north star we want to remain compatible with is that we leave folks with containers that are as good as we can possibly produce given the evolution of best practices over time.

I wonder if it would be worth including a bit about that as a sort of "mission statement" or "philosophy" for the tool in the main README.md?  Sort of like we added the principles section to GGCR.

`</pontification>`

mattmoor:
This applies similarly to: https://github.com/google/ko/issues/357

imjasonh:
Taking a stab at the latter two points:

.ko.yaml:
```yaml
verifySignatures:
  "alpine:1.23":
  - publicKey: https://some.url/path/to/key.pub
    annotations:
      key: value
      another: value2
```

If any image is based on `alpine:1.23` then `ko` will verify that it has a signature signed using the private key it can find at the URL, and has those annotations.

[...]

PR #433, opened 2021-09-12
[WIP] Start to sketch out integration with `cosign`.

mattmoor:
When `COSIGN_EXPERIMENTAL=true`, this will verify base images, and sign produced images using the KEYLESS flow.

Fixes: https://github.com/google/ko/issues/357
Fixes: https://github.com/google/ko/issues/356


mattmoor:
`cosign` uses Go 1.16 features, so we should discuss what to do with some of the older tests 🤔

pkg/commands/config.go
```diff
@@ -124,6 +127,18 @@ func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// With the resolved digest in hand (in descriptor), we verify the
+		// signature of the base image.
+		if cli.EnableExperimental() {
+			digest := ref.Context().String() + "@" + desc.Digest.String()
```
imjasonh, line 134: `ref.Context().Digest(desc.Digest.String())`

pkg/commands/config.go
```diff
@@ -139,6 +154,52 @@ func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
...
+}
+
+// entry is used to hold the sync.Once used to execute the digest verification,
+// and the resulting error (possibly nil), that results from verification.
+type entry struct {
+	once   sync.Once
+	result error
+}
+
+func (v *verifier) verify(ctx context.Context, digest string) error {
```
imjasonh, line 171: Can `digest` be a `name.Digest` instead of a string?

pkg/commands/config.go
```diff
@@ -139,6 +154,52 @@ func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
...
+	e := v.getEntry(digest)
+	e.once.Do(func() {
+		// cosign emits a lot of stuff to STDOUT that we don't want
```
imjasonh, line 174: I know this is just WIP but this is gross, we should fix this upstream before papering over it here.

imjasonh (commented):
I'm honestly pleaseantly surprised how relatively small this diff is, in terms of ko lines changed. However, we should try to refactor cosign's packages to make them easier for us (and others, hopefully) to consume in this way, so this diff isn't so omghuge.

pkg/commands/config.go
mattmoor: neat, TIL

pkg/commands/config.go
mattmoor: yup

pkg/commands/config.go
mattmoor: I did, it broke and got rolled back.  I need to sort out what needs to be fixed in my fix :)

mattmoor:
> I'm honestly pleaseantly surprised how relatively small this diff is

Yeah, I've been working to minimize it 😉 

Really aside from the output garbage and the logic to eliminate redundant verifications it's like 5 LoC each.

> we should try to refactor cosign's packages to make them easier for us (and others, hopefully) to consume in this way

Part of me wonders if we'd have even noticed if we didn't check in `vendor`.  I agree, and would certainly appreciate any help in that department.  🙏 😇

jonjohnsonjr (changes requested):
ಠ_ಠ

mattmoor:
@jonjohnsonjr @imjasonh https://github.com/sigstore/cosign/issues/657 👀

mattmoor:
@dlorenc thinks I should play with the stuff in `pkg/cosign` instead.  I'll see how much leaner that is from a `vendor` perspective, and how much more code it requires.

pkg/commands/config.go
mattmoor: This is gone from the latest update.  Inlining stuff FTW, I guess.

imjasonh:
Needs rebase.

PR #435, opened 2021-09-13
Bump to Go 1.16.x

mattmoor:
Go 1.16 introduces "embed", which folks are starting to adopt.

Folks are also starting to use things from `io` instead of `ioutil`, which is another 1.16 change I've seen folks picking up without even realizing it.

### DIFF
--- a/.github/workflows/modules-integration-test.yaml
+++ b/.github/workflows/modules-integration-test.yaml
@@ -10,7 +10,7 @@ jobs:
     name: Module Tests
     strategy:
       matrix:
-        go-version: [1.14.x, 1.15.x]
+        go-version: [1.16.x]
         platform: [ubuntu-latest]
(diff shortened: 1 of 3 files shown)


.github/workflows/modules-integration-test.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
     name: Module Tests
     strategy:
       matrix:
-        go-version: [1.14.x, 1.15.x]
+        go-version: [1.16.x]
```
imjasonh, line 13: See https://github.com/google/ko/pull/405

mattmoor: Do you need to steal the change I made to the script's `go get`?

imjasonh (approved):

PR #436, opened 2021-09-14
Incorporate "keyless" verification of base images.

mattmoor:
This change enables folks to start verifying their base images are signed against Fulcio by specifying `COSIGN_EXPERIMENTAL=true`, which will ensure each digest we intend to base an image on is verified.

Unless `COSIGN_EXPERIMENTAL=true` is enabled, this change will have no effect.

When `COSIGN_EXPERIMENTAL=true` is enabled, this change will verify base image digests, and emit a `WARNING:` when they are not signed (the `gcr.io/distroless` images are now signed!).

When `COSIGN_EXPERIMENTAL=true` is enabled *AND* `--strict` is passed, then failure to verify a base image digest will fail the `ko` command.

Related: https://github.com/google/ko/issues/356
Related: https://github.com/google/ko/pull/433


jonjohnsonjr:
I would like to take a chainsaw to the cosign dependency graph.

mattmoor:
Ok cool, as expected with `--strict` the windows leg fails because the `mcr.....` image for Windows isn't signed, duh...

mattmoor:
> I would like to take a chainsaw to the cosign dependency graph.

Do it!

mattmoor:
... and https://github.com/sigstore/cosign/issues/666 which seems very popular 😅

imjasonh:
Needs rebase.

mattmoor:
> Needs rebase.

Done

pkg/commands/config.go
```diff
@@ -55,6 +58,7 @@ var (
 // getBaseImage returns a function that determines the base image for a given import path.
 // If the `bo.BaseImage` parameter is non-empty, it overrides base image configuration from `.ko.yaml`.
 func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
+	v := &verifier{entries: make(map[name.Digest]*entry, 10)}
```
imjasonh, line 61: Why `10` here?

pkg/commands/config.go
```diff
@@ -139,6 +159,55 @@ func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
...
+}
+
+// entry is used to hold the sync.Once used to execute the digest verification,
+// and the resulting error (possibly nil), that results from verification.
+type entry struct {
+	once   sync.Once
+	result error
+}
+
+func (v *verifier) verify(ctx context.Context, digest name.Digest) error {
```
imjasonh, line 176: I'd like to have a better idea how verification will look in the fullness of time. For now, it sounds like it's just "verify this image has any signature" -- does it also look it up in Rekor? Eventually folks might want to disable checking Rekor, or express other signature requirements.

Even just a sketch of how we envision that looking (in `.ko.yaml`?) would be helpful to me.

pkg/commands/config.go
```diff
@@ -139,6 +159,55 @@ func getBaseImage(platform string, bo *options.BuildOptions) build.GetBase {
...
+		}
+	})
+	return e.result
+}
+
+func (v *verifier) getEntry(digest name.Digest) *entry {
+	v.m.Lock()
+	defer v.m.Unlock()
+	// See if there's already an entry for this digest.
+	if e, ok := v.entries[digest]; ok {
```
imjasonh, line 202: nit: You could hide `v.entries` initialization in here, instead of making it the responsibility of the `verifier` instantiator, above.

imjasonh (commented):
I think the main remaining hurdle from last week was excising unnecessary cosign deps so this doesn't explode the graph. Is that mainly done at this point, or is there more we should wait for?

pkg/commands/config.go
mattmoor: Seemed like a small enough number to be reasonable, yet avoid resizing.  I don't particularly care.

imjasonh: Same. I don't think there's any reason to care about the initial size of the map, so let's just YOLO-`make` it, or `map[name.Digest]*entry{}`.

Issue #406, opened 2021-08-02
Build summary for `ko resolve`


HeavyWombat:
We do builds with `ko resolve` to have images and also a deployment YAML. However, we were wondering whether there would be an easy way to get a list of all images (and their digest) that are created as part of the `resolve` run (without parsing the output). If not, would a command line flag, e.g. `--bom` (bill of materials, or similar) an option? Thanks in advance for feedback.

imjasonh:
In theory all the information you want is available somewhere in the mass of output that `ko` spams at you. Lines like:

```
2021/08/02 08:25:07 gcr.io/imjasonh/windows/github.com/tektoncd/pipeline/cmd/pullrequest-init@sha256:60ddf2f4cf2392f95da20ec6895b89143a6299bb8331898f58af9de81dcfa6e6: digest: sha256:60ddf2f4cf2392f95da20ec6895b89143a6299bb8331898f58af9de81dcfa6e6 size: 1328
```

...but I agree it's not terribly useful.

I'm a bit hesitant to call this better output a "BOM", because that term is quickly expanding to mean anything and everything anybody wants it to mean. 🙃 

Are you envisioning some secondary yaml/json output in some file somewhere that you can post-process? It should live outside the current working directory, since that will likely pollute your Git workspace.

Another similar idea is something like `ko extract` described in https://github.com/google/ko/issues/358#issuecomment-834599338 -- something that could crawl a YAML document looking for strings that look like `ko`-produced images, and output them. I swear I wrote code for this recently but I can't seem to find it now. 🤔  Anyway, that could be useful for lots of stuff, and it would compose well too, like `ko resolve | ko extract`, or even `cat release.yaml | ko extract`, or other bash monstrosities 😈

HeavyWombat:
I am not big fan of the term BOM myself, so I would be more than happy to have some secondary YAML/JSON that contains the container images - brief and simple.

jonjohnsonjr:
You can kinda do that today:

```
$ ko resolve -f ./test/test.yaml --oci-layout-path /tmp/ko --push
2021/08/02 08:22:08 Using base gcr.io/distroless/static:nonroot for github.com/google/ko/test
2021/08/02 08:22:09 Building github.com/google/ko/test for linux/amd64
2021/08/02 08:22:10 Saving ko://github.com/google/ko/test
2021/08/02 08:22:10 Saved ko://github.com/google/ko/test
[...]

Issue #358, opened 2021-05-07
Idea: ko rebase


imjasonh:
When a newer version of a base image is available, in some cases you can pick up those updates without having to completely rebuild your image from scratch on top of the new base image. Instead, you can [rebase](https://github.com/google/go-containerregistry/blob/main/cmd/crane/rebase.md). It just so happens `ko` builds images that are amenable to rebasing. 👍 

With [future improvements to the OCI spec](https://github.com/opencontainers/image-spec/pull/822) to declare standard annotations to signal base image information, [`crane` will be able to take advantage of these annotations](https://github.com/google/go-containerregistry/pull/960) to detect and rebase images without any external information required.

`ko resolve` builds a bunch of images, by default based on gcr.io/distroless/static:nonroot. If `ko` produced images that wrote base image annotations, it could also have a CLI surface to ingest YAML generated by a previous `ko resolve` to detect image references, determine base image information for those images, identify new available base image versions, perform a rebase, and replace image references in the output YAML.

The result would be a workflow like: `ko resolve > release.yaml`, wait for a distroless release, `ko rebase release.yaml > release.yaml`, and release.yaml would now include images based on the new distroless release, without having to rebuild anything at all, or even have access to the original source.

jonjohnsonjr:
> it could also have a CLI surface to ingest YAML generated by a previous ko resolve to detect image references

Detecting which strings are actually images is a bit tricky. With the original ko yaml, we have `ko://` prefixes, which makes this easy, but we will have to hit the internet to detect which strings are definitely images _and_ we'll need some way to distinguish `ko`-built images that we want to rebase.

I think it would be fine to have `ko rebase` as a command, but this makes me think about how we might decouple the operation from the source and sink.

Imagine:

```
# Extract images from release.yaml
ko extract -f release.yaml
example.com/foo:bar
example.com/baz@sha256:abc
```

```
# Drop into a streaming mode that reads from stdin and writes to stdout
ko extract -f release.yaml | crane rebase -
example.com/foo:bar@sha256:aaa
example.com/baz@sha256:def
```

```
# Equivalent to `ko rebase -f release.yaml`
ko extract -f release.yaml | crane rebase - | ko rewrite -f release.yaml --images - > rebased.yaml
```

Assuming that `ko extract` and `ko rewrite` use the same logic for finding images within `release.yaml` in the exact same order, `ko rewrite` can know which image references to replace. A little bit of care around deduplication, and we've got a lovely little tool :P

[...]

PR #443, opened 2021-09-20
Add `ko extract` to extract ko-built image references from rendered YAML

imjasonh:
@mattmoor @jonjohnsonjr 

Tekton currently does this with a custom-built Python script, [`koparse`](https://github.com/tektoncd/plumbing/blob/bd1811ea7d20fe8a9ff41bdbf02ea8583c586e97/tekton/images/koparse/koparse/koparse.py), and first-party support in `ko` itself could be useful.


imjasonh:
https://github.com/google/ko/issues/406
https://github.com/google/ko/issues/358#issuecomment-834599338

jonjohnsonjr:
I feel like it might be more useful to have a flag where `ko` can spit out anything it builds? I think `skaffold` does something similar with `--file-output`.

Doing this with a resolved yaml file means we'll pick up anything in `KO_DOCKER_REPO`, not just things that we built.

imjasonh:
> I feel like it might be more useful to have a flag where `ko` can spit out anything it builds? I think `skaffold` does something similar with `--file-output`.
> 
> Doing this with a resolved yaml file means we'll pick up anything in `KO_DOCKER_REPO`, not just things that we built.

That would be fine too, we'd just need to settle on an output format. OCI layout? Newline-delimited image refs@digest?

`ko extract` as it is in this PR is imperfect, sure, but it would do everything we need for Tekton and Knative (I think) -- I can't think of a use case where someone would have other `KO_DOCKER_REPO`-prefixed strings in their YAML that we shouldn't report. Maybe we could have `ko resolve` emit fully-qualified registry URLs to prevent the `KO_DOCKER_REPO=username` case, which could lead to trouble here.

We could document its limitations a bit better, if that's a concern.

jonjohnsonjr:
> We could document its limitations a bit better, if that's a concern.

Yeah this is my main issue. The command is fine, we should just be more transparent about what it's doing.

PR #445, opened 2021-09-21
[ci] upgrade to Go 1.17 everywhere, clean up configs

imjasonh:
- module integration test now runs with 1.16 and 1.17, and drops an unnecessary `matrix` value.
- drop unnecessary step `name`s (e.g., "Set up Go 1.17.x", which is redundant with the step config, and will only fall out of date and confuse people), and `shell:bash` (`bash` is the default value)

### DIFF
--- a/.github/workflows/boilerplate.yaml
+++ b/.github/workflows/boilerplate.yaml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - uses: actions/setup-go@v2
         with:
-          go-version: 1.16.x
+          go-version: 1.17.x
 
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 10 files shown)


imjasonh:
`gofmt` in Go 1.17+ adds `//go:build` lines corresponding to `// +build` lines it sees, which is causing the diff. This change is backward compatible for pre-1.17, so we could just `gofmt` with Go 1.17 to get those lines.

This could be annoying if users are adding new `// +build` lines and gofmt'ing with Go <1.17, but I don't think that's too likely, and the mitigation is simple -- upgrade your Go version.

wdyt?

jonjohnsonjr:
I'm ok with deleting the 1.13 thing?

> This could be annoying if users 

This is for contributors to `ko`, right? I think that's fine.

imjasonh:
https://github.com/google/ko/pull/446 drops <1.13 support, which is nice because I can't get the boilerplate check to pass with it anyway. 🤷

jonjohnsonjr (approved):

PR #449, opened 2021-09-28
Always produce OCI images and indexes

imjasonh:
Followup from #354 

Before this change, if the base image was a Docker manifest list, it could not be annotated with base image information. With this change, `ko` always produces OCI images and indexes, which can be annotated.

Verified manually as well:

```
$ crane manifest $(go run ./ publish ./test --platform=all) | jq '.annotations'
...
{
  "org.opencontainers.image.base.digest": "sha256:7cb5539ebb7b99352d736ed97668060cee123285f01705b910891acdf7d945e3",
  "org.opencontainers.image.base.name": "gcr.io/distroless/static:nonroot"
}                                                                                                                                                      
```

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -923,7 +923,6 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIn
 			Add: img,
 			Descriptor: v1.Descriptor{
 				URLs:        desc.URLs,
-				MediaType:   types.OCIManifestSchema1,
 				Annotations: desc.Annotations,
 				Platform:    desc.Platform,
 			},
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
My only reluctance here is breaking folks that push to old registries that won't support OCI. I guess they could just use an old version of `ko`... wdyt?

imjasonh:
> My only reluctance here is breaking folks that push to old registries that won't support OCI. I guess they could just use an old version of `ko`... wdyt?

Yeah... we could have an `--oci` flag that defaults true?

jonjohnsonjr:
> Yeah... we could have an --oci flag that defaults true?

I'd be ok with deferring this until someone complains.

mattmoor (approved):

PR #457, opened 2021-10-01
Re-enable codeql analysis

imjasonh:
These were dropped in https://github.com/google/ko/pull/364 and I don't remember why. The result is that code scanning alerts have been disabled since May: https://github.com/google/ko/security/code-scanning

### DIFF
--- a/.github/workflows/analyze.yaml
+++ b/.github/workflows/analyze.yaml
@@ -1,0 +1,25 @@
+name: Analyze
+
+on:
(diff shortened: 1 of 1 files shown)


imjasonh:
This seems to have found [two genuine issues](https://github.com/google/ko/security/code-scanning?query=is%3Aopen++pr%3A457+).

jonjohnsonjr (approved):

PR #477, opened 2021-10-13
Allow to customize `publish.Namer`

### LINKED ISSUE #476: Allow to customize `publish.Namer`
### Problem

Ko has a number of built-in [`publish.Namer`'s](https://pkg.go.dev/github.com/google/ko/pkg/publish#Namer), but despite its public type, the code doesn't allow to influence it. Also, the built-in namers, all, use the [`path.Join`](https://pkg.go.dev/path#Join) which is uses `/` char, which often prevents using _google/ko_ with container registries that don't allow nested directories - a majority.

https://github.com/google/ko/blob/6447264ff8b5d48aff64000f81bb0847aefc7bac/pkg/commands/options/publish.go#L119-L128

### Possible solutions

I think a specific function should be used instead of `path.Join`, and maybe it should use some environmental variable to allow overriding default delimiter (`/` char). It also seems to be good idea to allow providing completely custom implementation for users.

cardil:
Fixes #476

- Adds ability to pass custom `publish.Namer`
- Adds ability to override default path separator, allowing use with container registries that don't allow nested repositories (Quay, Docker Hub etc.)

### DIFF
--- a/pkg/commands/options/publish.go
+++ b/pkg/commands/options/publish.go
@@ -21,7 +21,6 @@ import (
 	"encoding/hex"
 	"os"
 	"path"
-	"strings"
 
 	"github.com/google/go-containerregistry/pkg/v1/daemon"
 	"github.com/google/ko/pkg/publish"
@@ -66,12 +65,8 @@ type PublishOptions struct {
 	BaseImportPaths bool
 	// Bare uses a tag on the KO_DOCKER_REPO without anything additional.
 	Bare bool
-	// ImageNameSeparator is used to create the image name from base repo and
-	// specific import path. If not set a default of "/" will be used.
-	ImageNameSeparator string
 	// ImageNamer can be used to pass a custom image name function. When given
-	// PreserveImportPaths, BaseImportPaths, Bare, and ImageNameSeparator has
-	// no effect.
+	// PreserveImportPaths, BaseImportPaths, Bare has no effect.
 	ImageNamer publish.Namer
 }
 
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
Exposing this as an option seems fine, but I'm less excited about another flag for this... want to split this into 2 PRs so we can land the namer change?

cardil:
Okay. That seems reasonable to me. I would only need the code change, as I'm embedding ko in my build tools, so I think I'll just drop the flag out in this PR.

pkg/commands/options/publish.go
```diff
@@ -65,6 +66,13 @@ type PublishOptions struct {
 	BaseImportPaths bool
 	// Bare uses a tag on the KO_DOCKER_REPO without anything additional.
 	Bare bool
+	// ImageNameSeparator is used to create the image name from base repo and
+	// specific import path. If not set a default of "/" will be used.
+	ImageNameSeparator string
```
jonjohnsonjr, line 71: Do we still need this now that the flag is dropped and you can just override `ImageNamer`?

cardil: Hmm. I could drop this too.

But, I thought it would be easier to just allow to change separator and still use built in algorithms.

cardil: I did remove the `ImageNameSeparator` option

pkg/commands/options/publish.go
```diff
@@ -98,31 +106,43 @@ func AddPublishArg(cmd *cobra.Command, po *PublishOptions) {
 		"Whether to just use KO_DOCKER_REPO without additional context (may not work properly with --tags).")
 }
 
-func packageWithMD5(base, importpath string) string {
-	hasher := md5.New() //nolint: gosec // No strong cryptography needed.
+func pathJoin(paths ...string) string {
```
jonjohnsonjr, line 109: This seems more complicated than just using `path.Join` -- is there a reason for it?

cardil: That's a left over from removed option. On it.

jonjohnsonjr (approved):

jonjohnsonjr:
Thanks!

Issue #479, opened 2021-10-15
Enable optionally using `gotip`


imjasonh:
[`gotip`](https://pkg.go.dev/golang.org/dl/gotip) is a tool provided by the Go team that lets you try out more easily unreleased Go features.

You install `gotip` (`go install golang.org/dl/gotip@latest`), then `gotip download` will update `gotip` to a Go toolchain built from the tip of the Go repo. In every other way it's interchangeable with the regular `go` binary -- running `gotip build` will build with the Go toolchain built from head.

I propose some env var or (global?) flag to `ko` to use `gotip` instead of `go`. Something like:

```
KO_EXPERIMENTAL_GOTIP=true ko build ./
```

This would switch invocations of `go` in [`gobuild.go`](https://github.com/google/ko/blob/main/pkg/build/gobuild.go) and used by `ko deps`, etc., to invoke `gotip` instead.

PR #481, opened 2021-10-18
Allow users to use gotip

imjasonh:
Fixes https://github.com/google/ko/issues/479

Would we prefer this to be a `build.Option`, which would make it easier for packages that use `ko` as a library?

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -33,6 +33,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
> ..or just `alias go=gotip`

I think if we're going to allow substitution of the `go` command, we should talk about how this might work for non-go images. If aliasing like this is sufficient, maybe this PR could just be a docs change instead for now?

imjasonh:
I'm fine just closing it for now. If anybody wants gotip besides me I can add docs for it.

PR #492, opened 2021-11-05
Enforce more lint checks, fix findings

imjasonh:


pkg/commands/options/build.go
```diff
@@ -91,15 +92,15 @@ func (bo *BuildOptions) LoadConfig() error {
 	v.AddConfigPath(bo.WorkingDirectory)
 
 	if err := v.ReadInConfig(); err != nil {
-		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-			return fmt.Errorf("error reading config file: %v", err)
+		if !errors.As(err, &viper.ConfigFileNotFoundError{}) {
```
mattmoor, line 95: Can this be `Is`?

imjasonh: Maybe. I couldn't figure out the correct invocation to get it to pass both lint and tests.

mattmoor (approved):
A question, and there's a CodeQL thing we should look at, but otherwise 👍 

thanks for the cleanup 🙏

pkg/commands/options/build.go
mattmoor: Ok, then the only think I had was the CodeQL bit, which I think is a problem at HEAD.

imjasonh: Yeah I'll tackle that one separately, I don't think it's a genuine issue at least, but convincing codeql might take some work.

imjasonh:
> A question, and there's a CodeQL thing we should look at, but otherwise 👍

The CodeQL alert was supposed to be fixed by https://github.com/google/ko/pull/459, I think this is some regression in CodeQL that it considers this an error again. 😕

PR #494, opened 2021-11-09
Fix issue where kubectl apply is run in place of kubectl create

sbwsg:
I think this use of `kubectl create` was accidentally replaced by `kubectl apply` (assumption based on the fact that the comments and error messages surrounding this change weren't updated to reflect new use of `apply`).

We've started seeing errors from `ko` in Tekton Pipelines' integration test runs with the following format today, which I believe are related:

```
        error: from default-cluster-admin-: cannot use generate name with apply
        Error: error executing 'kubectl create': exit status 1
        2021/11/09 16:40:53 error during command execution:error executing 'kubectl create': exit status 1
```


sbwsg:
@googlebot I fixed it.

PR #498, opened 2021-11-10
feat: sign checksum using cosign

### LINKED ISSUE #491: Sign released ko binaries
Let's consider adding a signing step for ko binaries and container images with cosign? Or we should sign a checksum file as `GoReleaser` did [^1]

[^1]: https://carlosbecker.com/posts/goreleaser-cosign/

cc: @dentrax @imjasonh @jonjohnsonjr

Dentrax:
Signed-off-by: Furkan <furkan.turkal@trendyol.com>

Fixes #491

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,9 @@ on:
 
 jobs:
   goreleaser:
+    permissions:
+      id-token: write
+      contents: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -15,6 +18,7 @@ jobs:
         with:
           go-version: 1.18
           check-latest: true
+      - uses: sigstore/cosign-installer@581838fbedd492d2350a9ecd427a95d6de1e5d01 # v2.0.1
       - uses: goreleaser/goreleaser-action@v2.9.1
         with:
           version: latest
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -34,6 +34,19 @@ archives:
     amd64: x86_64
 checksum:
   name_template: 'checksums.txt'
+signs:
+  - cmd: cosign
+    env:
+      - COSIGN_EXPERIMENTAL=1
+    signature: "${artifact}.sig"
+    certificate: "${artifact}.pem"
+    args:
+      - sign-blob
+      - '--oidc-issuer={{if index .Env "CI"}}https://token.actions.githubusercontent.com{{else}}https://oauth2.sigstore.dev/auth{{end}}'
+      - '--output-certificate=${certificate}'
+      - '--output-signature=${signature}'
+      - '${artifact}'
+    artifacts: checksum
 snapshot:
   name_template: "{{ .Tag }}-next"
 changelog:


Dentrax:
We used `cosign-release: 'v1.3.0'` here, but we can not do verify for now since we have to wait the next cosign release. Here is the related PR: https://github.com/sigstore/cosign/pull/991 by @shibumi. This feature scheduled in the next [v1.4.0](https://github.com/sigstore/cosign/milestone/12?closed=1) release. If we want to export the cert in the release, then probably we have to wait one of these PRs to get merged: https://github.com/sigstore/cosign/pull/1016 and https://github.com/sigstore/cosign/pull/1021 by @developer-guy

.goreleaser.yml
```diff
@@ -34,6 +34,10 @@ archives:
     amd64: x86_64
 checksum:
   name_template: 'checksums.txt'
+signs:
+  - cmd: cosign
+    args: ["sign-blob", "--oidc-issuer=https://token.actions.githubusercontent.com", "--output=${signature}", "${artifact}"]
```
shibumi, line 39: This part is missing the `signature: ${artifact}.sig` value.. also it might makes sense to wait for the goreleaser support for releasing certificates, too :)

developer-guy: IMHO we don't need this one because we're using `artifacts: checksum,` so it creates `checksums.txt.sig` by default, as you can see from [here](https://github.com/Dentrax/ko/releases/download/v0.9.3-signoidc/checksums.txt.sig). Also, we don't need to wait to release the certificates feature because in cosign with https://github.com/sigstore/cosign/pull/991, we can verify blobs in keyless mode.  The problem here is that these certificates are short-lived IIRC, so for users who might want to use these to verify the ko binary might be problematic because these certificates will be expired in a short time, WDYT @shibumi @imjasonh?

shibumi: Just for the record (@developer-guy  know this already). Short-Lived certificates are not problematic, because it just refers to the signature generation, not the signature verification. See also: https://github.com/goreleaser/goreleaser/issues/2659#issuecomment-965756308

.goreleaser.yml
```diff
@@ -34,6 +34,14 @@ archives:
     amd64: x86_64
 checksum:
   name_template: 'checksums.txt'
+signs:
+  - cmd: cosign
+    env:
+    - COSIGN_EXPERIMENTAL=1
+    certificate: '{{ trimsuffix .Env.artifact ".txt" }}.pem'
+    signature: "${artifact}.sig"
+    args: ["sign-blob", "--oidc-issuer=https://token.actions.githubusercontent.com", "--output-signature=${signature}", "--output-certificate=${certificate}",  "${artifact}"]
```
shibumi, line 43: Note: this will not work with cosign v1.3.1, because the implemented feature is not part of that release.
CC: @dlorenc can we get a new cosign release, soon? :D

shibumi: I ran into the same issue with my new kubectl plugin... cosign will fail with "unknown flag --output-signature" with version v1.3.1

developer-guy:
@shibumi cosign v1.4.1 with some bunch of fixes is released today as you might know, would you like to give this issue a hand? 🤩 Or I can do it if you want if you don't have time 🤝

shibumi:
@developer-guy  I can confirm this works: https://github.com/shibumi/kubectl-htpasswd/releases/tag/v0.1.7

But you *must* set the experimental flag with cosign v1.4.1

.goreleaser.yml
shibumi: this is solved with cosign 1.4.1

.github/workflows/release.yml
```diff
@@ -7,14 +7,18 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow
-      - uses: actions/setup-go@v1
+      - uses: actions/setup-go@v2
         with:
           go-version: 1.17.x
-      - uses: goreleaser/goreleaser-action@v1
+      - uses: sigstore/cosign-installer@v1.3.1
```
shibumi, line 20: Please update to cosign-installer@v1.4.1

.goreleaser.yml
```diff
@@ -34,6 +34,14 @@ archives:
...
+    artifacts: checksum
```
shibumi, line 44: Do you really only want to sign the checksums?

shibumi (changes requested):

.goreleaser.yml
developer-guy: thank you 🙋🏻‍♂️

.github/workflows/release.yml
developer-guy: thank you 🙋🏻‍♂️

.goreleaser.yml
developer-guy: I left it empty. 
https://github.com/goreleaser/goreleaser/pull/2716/files#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826R208-R218

developer-guy:
kindly ping @mattmoor @imjasonh @shibumi, seems everything works fine.
👉 https://github.com/developer-guy/ko/releases/tag/v0.9.3-signchecksum

```shell
$ httpie --download https://github.com/developer-guy/ko/releases/download/v0.9.3-signchecksum/checksums.txt.sig
$ httpie --download https://github.com/developer-guy/ko/releases/download/v0.9.3-signchecksum/checksums.txt.pem
$ httpie --download https://github.com/developer-guy/ko/releases/download/v0.9.3-signchecksum/checksums.txt

$ cosign verify-blob --signature checksums.txt.sig --cert checksums.txt.pem checksums.txt
Verified OK
```

👉 https://github.com/developer-guy/ko/runs/4592518036?check_suite_focus=true

.goreleaser.yml
```diff
@@ -34,6 +34,19 @@ archives:
...
   name_template: 'checksums.txt'
+signs:
+  - cmd: cosign
+    env:
+      - COSIGN_EXPERIMENTAL=1
+    signature: "${artifact}.sig"
+    certificate: "${artifact}.pem"
+    args:
+      - sign-blob
+      - '--oidc-issuer={{if index .Env "CI"}}https://token.actions.githubusercontent.com{{else}}https://oauth2.sigstore.dev/auth{{end}}'
```
imjasonh, line 45: Is the `CI` env var the best way we have to detect we're running in CI? Maybe the presence of the `GITHUB_SHA` env var?

(edit: Oh, this is copied from goreleaser/goreleaser's similar config. Does the goreleaser action set the `CI` env var? 🤔)

.github/workflows/release.yml
```diff
@@ -7,14 +7,18 @@ on:
...
     steps:
       - uses: actions/checkout@v2
       - run: git fetch --prune --unshallow
-      - uses: actions/setup-go@v1
+      - uses: actions/setup-go@v2
         with:
           go-version: 1.17.x
-      - uses: goreleaser/goreleaser-action@v1
+      - uses: sigstore/cosign-installer@v1.4.1
+      - uses: goreleaser/goreleaser-action@v2
```
imjasonh, line 21: Could we add a step after goreleaser that runs `cosign verify <checksum>` to check that this worked as expected?

shibumi: Question: Is cosign validating this already? I like the idea of validating the signature after signature creation, but I am not sure if validating makes sense if cosign is validating it already during creation.

shibumi:
@Dentrax can you add  these two lines to the goreleaser  configuration? They ensure that goreleaser will produce a source tarball and signs this tarball as well. It provides an easy way to download a signed tarball via curl
```yaml
source:
  enabled: true
```

.goreleaser.yml
developer-guy: I think this approach might help for local development, so, to ensure that someone who might want to run the goreleaser commands locally still can do that If I'm not mistaken.

.github/workflows/release.yml
developer-guy: nice let's do this.

developer-guy:
we only sign the chekcsums.txt file right now, I think it is enough for verifying sha256 of the files has not been tampered with but, the final decision belongs to @imjasonh, let's ask that to him and wait for his response, then according to his response, maybe we can start signing tar files instead of just checksums.txt file. WDYT @shibumi?

developer-guy:
Btw @imjasonh GoReleaser is now capable of generating SBOMs by using the Syft tool under the hood, so, that we can add that support to the ko project🙋🏻‍♂️

imjasonh:
> Btw @imjasonh GoReleaser is now capable of generating SBOMs by using the Syft tool under the hood, so, that we can add that support to the ko project🙋🏻‍♂️

I'm not opposed, but I think that's a separate issue. I'd like to make sure that Syft's SBOM of `ko` and `ko`'s SBOM of `ko` don't disagree. 🙃

.goreleaser.yml
imjasonh: nvm, GitHub Actions says `CI` is always set during GHA workflow execution, so this is right.

developer-guy:
I removed the --oidc-issuer flag to enable ambient credential detection support in cosign

shibumi:
> we only sign the chekcsums.txt file right now, I think it is enough for verifying sha256 of the files has not been tampered with but, the final decision belongs to @imjasonh, let's ask that to him and wait for his response, then according to his response, maybe we can start signing tar files instead of just checksums.txt file. WDYT @shibumi?

You misunderstood me. The problem is: If you do not generate source tarballs with goreleaser, goreleaser will not create a source tarball for the project. Instead, Github will create the source tarball, but there will be `no` checksum file for the source tarball.

developer-guy:
kindly ping @imjasonh 🙋🏻‍♂️

developer-guy:
kindly ping @imjasonh

Dentrax:
Rebased, updated. https://github.com/Dentrax/ko/releases/tag/v0.0.9

developer-guy:
kindly ping @imjasonh

.github/workflows/release.yml
```diff
@@ -7,6 +7,9 @@ on:
 
 jobs:
   goreleaser:
+    permissions:
+      id-token: write
+      contents: write
```
imjasonh, line 12: Just to confirm: This needs `contents: write` to be able to add things to the release right?

imjasonh (commented):
Sorry it's taken so long to get back to this. 😞 

Some time shortly before or after we cut a `ko` release we should include some docs about how to verify the releases, and what verifying proves. But I don't think we should add docs to README.md yet that won't be useful for current releases.

imjasonh:
I think with #730 we have a fair assurance that the artifacts that we've released have come from reliable sources and are tamper-evident. I'll close this unless folks think there's more we should do to gain more assurance.

Issue #500, opened 2021-11-12
Consider not adding `-trimpath` flag by default when disabling compiler optimizations


halvards:
`ko` adds the `-trimpath` flag if no build config is specified: https://github.com/google/ko/blob/5617d1ebf8560df778aa884fa328fe05f62bdb5e/pkg/build/gobuild.go#L562

This removes information necessary for remote debugging, see: https://github.com/GoogleContainerTools/skaffold/issues/6843#issuecomment-966758261

Users who want to debug would typically also use the [`--disable-optimizations`](https://github.com/google/ko/blob/5617d1ebf8560df778aa884fa328fe05f62bdb5e/pkg/build/gobuild.go#L565) flag.

Should we default to _not_ add `-trimpath` if `disableOptimizations` is true?

In other words, change [this conditional](https://github.com/google/ko/blob/5617d1ebf8560df778aa884fa328fe05f62bdb5e/pkg/build/gobuild.go#L560) from `!ok` to `!ok && !g.disableOptimizations`?

A workaround is to ask users to specify a build config if they want to control whether the `-trimpath` flag is added.

halvards:
(By the way, I'd be happy to implement it.)

imjasonh:
Not making a judgement either way for now, but this issue explains why we added `-trimpath` in the first place: https://github.com/google/ko/issues/71

halvards:
Thanks for the context, I :heart: reproducible builds too.

But when interactively debugging it's not as important IMHO. Or is `--disable-optimizations` used for other purposes too?

imjasonh:
I don't know enough about Skaffold, but does it have some "run this in debug mode" option that could disable `-trimpath` for `ko` builds? Or is it always "run this code; I may want to debug it later"?

If Skaffold's use case is iterative (local?) development and not releasing to prod, I care a lot less about the reproducibility of Skaffold's `ko` builds, you could maybe even just have Skaffold always unset `-trimpath` and get everything you want. I'm not sure this even needs changes in `ko` to enable that.

halvards:
Thanks for the reply, I'm definitely happy to discuss other solutions!

To your questions:

> does it have some "run this in debug mode"

Yes, that's exactly the use case. Users run `skaffold debug` and attach a debugger, or perform an equivalent action using the Cloud Code IDE extensions. Reproducibility is not important for this action.

> you could maybe even just have Skaffold always unset `-trimpath`

(warning: long answer!)

[...]

Issue #78, opened 2019-09-02
Document how to debug applications


slinkydeveloper:
Hi,
I would like to know how to debug applications deployed with ko with a remote debugger (like https://squash.solo.io/)

jonjohnsonjr:
cc @Harwayne @ilackarms 
xref: https://github.com/knative/eventing/issues/1733

I think the short version is that you need a base image that contains `ls` and use the `--disable-optimizations` flag. We should definitely make this easier and document it, though.

slinkydeveloper:
Can we enable remote debugging just packaging dlv and running the application with dlv? In a similar fashion to this: https://mikemadisonweb.github.io/2018/06/14/go-remote-debug/

jonjohnsonjr:
I think it's a good idea -- I [prototyped](https://github.com/google/go-containerregistry/compare/master...jonjohnsonjr:ko-debug) something like this that would rewrite the base image and entrypoint to automagically drop you into a delve debugging session, but ultimately my lack of k8s/delve knowledge made me give up. It only worked ~30% of the time and leaked a bunch of connections/resources that were tough to clean up.

Getting the user experience correct here is tough, I think. We could make things incrementally better pretty easily, e.g. some way to just include delve in the produced binary is a good start.

jonjohnsonjr:
The skaffold folks are doing some interesting work that we might be able to steal:

https://github.com/GoogleContainerTools/container-debug-support
https://github.com/GoogleContainerTools/skaffold/pull/2306

PR #505, opened 2021-11-17
Add `Trimpath` build option

halvards:
Enables programmatic control of adding the `-trimpath` flag to `go build`.

The `-trimpath` flag removes file system paths from the resulting binary. `ko` adds `-trimpath` by default as it aids in achieving reproducible builds.

However, removing file system paths makes interactive debugging more challenging, in particular in mapping source file locations in the IDE to debug information in the binary.

If you set `Trimpath` to `false` to enable interactive debugging, you likely also want to set `DisableOptimizations` to `true` to disable compiler optimizations and inlining.

Reference for `-trimpath`: https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies

Resolves: #500
Related: #71, #78, https://github.com/GoogleContainerTools/skaffold/issues/6843

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -65,6 +65,7 @@ type gobuild struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	disableOptimizations bool
+	trimpath             bool
 	buildConfigs         map[string]Config
 	platformMatcher      *platformMatcher
 	dir                  string
@@ -80,6 +81,7 @@ type gobuildOpener struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	disableOptimizations bool
+	trimpath             bool
 	buildConfigs         map[string]Config
 	platform             string
 	labels               map[string]string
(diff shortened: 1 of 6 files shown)


pkg/build/gobuild.go
```diff
@@ -556,9 +559,10 @@ func createBuildArgs(buildCfg Config) ([]string, error) {
 }
 
 func (g *gobuild) configForImportPath(ip string) Config {
-	config, ok := g.buildConfigs[ip]
-	if !ok {
-		// Apply default build flags in case none were supplied
+	config := g.buildConfigs[ip]
+	if g.trimpath {
```
thielepaul, line 563: I am trying to disable `-trimpath` for local debugging. After reading https://github.com/ko-build/ko/issues/500#issuecomment-970695696 I hoped that setting `flags: []` should be sufficient to do so.
Maybe, we could change the check in this line in the following way:
```go
	if g.trimpath && config.Flags == nil {
```
Or is there another way to make a build without trimpath that I am currently missing?

halvards: `flags: []` should be sufficient I believe, or if not then you could set some other flag, such as `["-v"]`.
Did you verify that `ko` picked up the build config correctly from `.ko.yaml`?

thielepaul: Yes, I verified that the .ko.yaml is used by adding an invalid flag that made the build fail. I also did a local build with a print statement and the trimpath flag was always appended no matter if other flags were set or not.

PR #507, opened 2021-11-20
Generate ko deps in SPDX format

imjasonh:
- copy out a bunch of BuildInfo stuff that will land in 1.18

The intention is that this should marry up with Matt's #506 so that SBOMs can be generated at image-build time.


pkg/commands/deps.go
```diff
@@ -112,10 +115,24 @@ If the image was not built using ko, or if it was built without embedding depend
 				cmd := exec.CommandContext(ctx, "go", "version", "-m", n)
 				cmd.Stdout = os.Stdout
 				cmd.Stderr = os.Stderr
+				if spdx {
+					var buf bytes.Buffer
+					cmd.Stdout = &buf
+					if err := cmd.Run(); err != nil {
+						return err
+					}
+					b, err := sbom.GenerateSPDX(Version, h.ModTime, buf.Bytes())
```
mattmoor, line 124: I might replace `h.ModTime` with `cfg.Created`

pkg/commands/deps.go
```diff
@@ -112,10 +115,24 @@ If the image was not built using ko, or if it was built without embedding depend
 				cmd := exec.CommandContext(ctx, "go", "version", "-m", n)
 				cmd.Stdout = os.Stdout
 				cmd.Stderr = os.Stderr
+				if spdx {
```
mattmoor, line 118: Consider normalizing the vanilla output of `ko deps` so instead of `/tmp/garbage/` is prints `/ko-app/blah`?

pkg/commands/deps.go
imjasonh: Done.

pkg/commands/deps.go
imjasonh: Done.

pkg/commands/deps.go
```diff
@@ -110,12 +120,34 @@ If the image was not built using ko, or if it was built without embedding depend
...
+				if err := cmd.Run(); err != nil {
+					return err
+				}
+				// In order to get deterministics SBOMs replace
+				// our randomized file name with the path the
+				// app will get inside of the container.
+				mod := bytes.Replace(buf.Bytes(),
+					[]byte(n),
+					[]byte(path.Join("/ko-app", filepath.Base(filepath.Clean(h.Name)))),
+					1)
```
mattmoor, line 135: You should be able to remove the `sed -1d` in the actions flow with this 🤩

mattmoor (commented):
We should hook this up to the other stuff now too 🎉

pkg/commands/deps.go
mattmoor: https://github.com/google/ko/blob/6d0691330a6df9407a1232e8e8835c6860387b22/.github/workflows/kind-e2e.yaml#L89-L90

mattmoor (approved):

PR #511, opened 2021-11-22
Connect SBOMs with SPDX support.

mattmoor:
This combines Jason's SPDX stuff and my SBOM stuff to support
SPDX-based SBOMs by default instead of our `go version -m`
invention.

To simplify things, I got rid of `disableSBOM` in favor of supporting
`nil` and simply eliding the SBOM call.

### DIFF
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -86,7 +86,7 @@ jobs:
 
         IMAGE=$(ko build ./test)
         SBOM=$(cosign download sbom ${IMAGE})
-        KO_DEPS=$(ko deps --sbom=spdx ${IMAGE})
+        KO_DEPS=$(ko deps ${IMAGE})
 
         echo '::group:: SBOM'
         echo "${SBOM}"
--- a/.github/workflows/kind-e2e.yaml
+++ b/.github/workflows/kind-e2e.yaml
@@ -84,9 +84,9 @@ jobs:
       run: |
         set -o pipefail
 
-        IMAGE=$(ko publish ./test)
+        IMAGE=$(ko build ./test)
         SBOM=$(cosign download sbom ${IMAGE})
-        KO_DEPS=$(ko deps ${IMAGE})
+        KO_DEPS=$(ko deps --sbom=spdx ${IMAGE})
 
         echo '::group:: SBOM'
         echo "${SBOM}"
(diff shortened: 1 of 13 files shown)


pkg/build/gobuild.go
```diff
@@ -130,8 +129,8 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 func NewGo(ctx context.Context, dir string, options ...Option) (Interface, error) {
 	gbo := &gobuildOpener{
 		build: build,
-		sbom:  sbom,
 		dir:   dir,
+		sbom:  spdx("(none)"),
```
imjasonh, line 133: Should this pass a real version value?

pkg/build/options.go
```diff
@@ -116,6 +116,24 @@ func withBuilder(b builder) Option {
 	}
 }
 
+// WithGoVersionSBOM is a functional option to direct ko to use
+// go version -m for SBOM format.
+func WithGoVersionSBOM() Option {
```
imjasonh, line 121: I'll probably make this `WithSBOM(impl)` but that can happen later.

.github/workflows/kind-e2e.yaml
```diff
@@ -84,9 +84,9 @@ jobs:
       run: |
         set -o pipefail
 
-        IMAGE=$(ko publish ./test)
+        IMAGE=$(ko build ./test)
         SBOM=$(cosign download sbom ${IMAGE})
-        KO_DEPS=$(ko deps ${IMAGE})
+        KO_DEPS=$(ko deps --sbom=spdx ${IMAGE})
```
imjasonh, line 89: Can you make this the default for `ko deps` too?

imjasonh (approved):
🎉🎉🎉

pkg/build/options.go
mattmoor: I hate the signature for `sbomber`, so I didn't go that route (I started to), but I'd love to head that direction if you have ideas.

.github/workflows/kind-e2e.yaml
mattmoor: Sure

pkg/build/gobuild.go
mattmoor: So this is something I spent a bit of time futzing around with, so LMK if you have idea wrt plumbing, but basically the place we inject `Version` today is in the consuming `pkg/commands` package.  The way the CLI defaults are set up, this is always overridden with `WithSPDX(Version)` so we get `(devel)` at HEAD today, and presumably even better with releases.  However, I wanted the library default to be SPDX as well, and used `(none)` to distinguish.

imjasonh (approved):

PR #512, opened 2021-11-22
Normalize the package hash to hex.

mattmoor:
We were emitting package checksum hashes as `h1:{base64}`.  `h1:` is a prefix that indicates "Hash 1", which is a SHA-256 based hash of the files, which is then base64 encoded as the suffix.

This change detects/strips the `h1:` prefix and re-encodes the base64 data as hex.

I also combined the two loops over `.Deps` since I noticed `bom` doing things in this order. 🤷

### DIFF
--- a/internal/sbom/spdx.go
+++ b/internal/sbom/spdx.go
@@ -16,6 +16,9 @@ package sbom
 
 import (
 	"bytes"
+	"encoding/base64"
+	"encoding/hex"
+	"fmt"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


mattmoor:
So it looks like for whatever reason locally I get:
```
dep	github.com/google/go-containerregistry	v0.7.0	h1:u0onUUOcyoCDHEiJoyR1R1gx5er1+r06V5DBhUU5ndk=
```

However, in CI we're seeing:
```
dep	github.com/google/go-containerregistry	v0.7.0	
```

Not sure what's going on here, but seems like the `Sum` field should be considered optional in our SPDX conversion.

cc @imjasonh

mattmoor:
For now I've predicated the `PackageChecksum` on `.Sum` being present, but we should figure out if there's a way to ensure this comes out.

PR #514, opened 2021-11-24
Support Openshift's internal registry with ocp.local.

markusthoemmes:
This adds support for Openshift's internal registry if the user specifies `ocp.local` as their docker repo. It logs into the registry, sets up a tunnel, connects to the correct port and then pushes the images "as usual" to the cluster.

This hasn't been widely discussed yet, so I don't expect it to merge timely if at all. This also serves as a prototype for more generalized tunneling support in an issue I'm just writing up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,8 +513,8 @@ Oh, you betcha. Here's a partial list:
 
 ## Does `ko` work with [OpenShift Internal Registry](https://docs.openshift.com/container-platform/latest/registry/registry-options.html#registry-integrated-openshift-registry_registry-options)?
 
-We've got you covered! Make sure you're
-[logged into `oc`](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html#cli-logging-in_cli-developer-commands)
+We've got you covered! Make sure 
+[you've installed `oc` and you're logged into your cluster](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html#cli-logging-in_cli-developer-commands)
 with a user that's allowed to push to the internal registry and use `ocp.local` as your
 `KO_DOCKER_REPO`. The images will be published into the currently active project.
 
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ Oh, you betcha. Here's a partial list:
 
 ## Does `ko` work with [OpenShift Internal Registry](https://docs.openshift.com/container-platform/latest/registry/registry-options.html#registry-integrated-openshift-registry_registry-options)?
 
-We've got you covered! Make sure 
+We've got you covered! Make sure
 [you've installed `oc` and you're logged into your cluster](https://docs.openshift.com/container-platform/latest/cli_reference/openshift_cli/getting-started-cli.html#cli-logging-in_cli-developer-commands)
 with a user that's allowed to push to the internal registry and use `ocp.local` as your
 `KO_DOCKER_REPO`. The images will be published into the currently active project.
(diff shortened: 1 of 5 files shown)


pkg/publish/openshift.go
```diff
@@ -0,0 +1,154 @@
...
+type ocpPublisher struct {
+	inner  Interface
+	tunnel *os.Process
+}
+
+// NewOpenshiftPublisher returns a new publish.Interface that loads images into
+// Openshift's internal registry.
+func NewOpenshiftPublisher(namer Namer, tags []string) (Interface, error) {
+	// Login to the registry.
+	login := exec.Command("oc", "registry", "login")
```
imjasonh, line 52: Can this be a `runOc`?

pkg/publish/openshift.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+type ocpPublisher struct {
+	inner  Interface
+	tunnel *os.Process
+}
+
+// NewOpenshiftPublisher returns a new publish.Interface that loads images into
+// Openshift's internal registry.
+func NewOpenshiftPublisher(namer Namer, tags []string) (Interface, error) {
```
imjasonh, line 50: supernit: I think the trademark is technically "OpenShift" with a capital S, I'm ashamed to even be this pedantic.

pkg/publish/openshift.go
```diff
@@ -0,0 +1,154 @@
...
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/ko/pkg/build"
+)
+
+const (
+	// OpenshiftDomain is a sentinel "registry" that represents loading images into
+	// Openshift's internal registry.
+	OpenshiftDomain = "ocp.local"
```
imjasonh, line 37: It's not really "local" is it? Maybe `ocp.internal`?

README.md
```diff
@@ -513,25 +513,10 @@ Oh, you betcha. Here's a partial list:
...
-```
-
-- Create a namespace where you will push your images, i.e: `ko-images`
-- Execute this command to set `KO_DOCKER_REPO` to publish images to the internal
-  registry.
-
-```sh
-   export KO_DOCKER_REPO=$(oc registry info --public)/ko-images
-```
+We've got you covered! Make sure you're
```
imjasonh, line 516: Also make sure you have `oc` installed in the first place 😄

README.md
```diff
@@ -513,25 +513,10 @@ Oh, you betcha. Here's a partial list:
 
 ## Does `ko` work with [OpenShift Internal Registry](https://docs.openshift.com/container-platform/latest/registry/registry-options.html#registry-integrated-openshift-registry_registry-options)?
```
imjasonh, line 514: I think once we get this and the other any-k8s-internal-registry publishers in, we can just move these up to the [Naming Images](https://github.com/google/ko#naming-images) section above, where `ko.local` and `kind.local` are documented.

imjasonh (commented):
This is the one rare occasion where deleting documentation is allowed -- this is so much easier! 💯

README.md
markusthoemmes: I should actually implement a check for that with a nice error message 😂

pkg/publish/openshift.go
markusthoemmes: Depends on how you look at it. It's local in the sense of "local to the cluster". In the same way, `kind.local` may or may not really be local.

I'm not married to it, I just stuck with what was there already. I'm fine with `ocp.internal` as well.

pkg/publish/openshift.go
markusthoemmes: ack!

pkg/publish/openshift.go
markusthoemmes: Yep!

README.md
markusthoemmes: Yeah, I guess at some point we should just list the specialcases we integrate with.

Issue #515, opened 2021-11-24
Support registries available via a tunnel


markusthoemmes:
Sometimes, registries are not readily available via "normal" networking means (i.e. the internet). That is true for registries that run inside of a cluster for example (e.g. [Openshift's internal registry](https://docs.openshift.com/container-platform/latest/registry/registry-options.html#registry-integrated-openshift-registry_registry-options)).

In order to use `ko` with such registries, one usually has 2 options:
1. Expose them to the internet (as we [document today for Openshift for instance](https://github.com/google/ko#does-ko-work-with-openshift-internal-registry)) and then use `ko` with the externally available registry.
2. Create a tunnel to the registry (`kubectl tunnel ...`), create a DNS entry to match up the internal hostname (i.e. `$svc.$ns.svc:$port`) to the local tunnel and then use `ko` with the internal name. The DNS entry is necessary as the cluster internally won't know what to do with `localhost:61237` for example.

Both are IMO lacking quite a bit and are more complex than they need to be. Exposing a registry also might not always be an options.

Soooo, it would be much easier if we provided a way for us to tell `ko`, that it should connect to a specific address in order to talk to the registry, while keeping the registry information as-is. I could imagine a new environment variable `KO_DOCKER_REPO_HOST` for example to supplement this information. Super open to alternative suggestions though.

### Bonus step 2: Auto-tunnel

In addition to manually setting up the tunnel and providing the necessary information, I could also see us adding an option that creates the tunnel for the user, i.e. if a `KO_DOCKER_REPO` of `$svc.$ns.svc:$port` is set we run `kubectl port-forward -n $ns svc/$svc :$port` for the user and make all the settings for them to be able to then talk to the registry through this tunnel.  The tunnel is then automatically shut down after `ko` is done.

What do you think? Happy to send a PR with `KO_DOCKER_REPO_HOST` support for a starter. Individual pieces of how this **can** look like can be seen in my Openshift Internal Registry prototype here: https://github.com/google/ko/pull/514

PR #516, opened 2021-11-24
Enable talking to tunneled registries via KO_DOCKER_REPO_ADDRESS

markusthoemmes:
**Note:** The issue hasn't been discussed and/or accepted yet, so I'm happy to bin this if not deemed useful. I just had a good flow 😂 
**Note 2:** Will think about tests if this is deemed useful and we want to move forward with it.

Fixes https://github.com/google/ko/issues/515 (without the Bonus)

This adds a new `KO_DOCKER_REPO_HOST` environment variable that, if set, forces the publisher to connect to that exact host. This can be used to connect to `localhost:5000` (for example to a tunnel) even if the registry is something like `my-registry.my-namespace.svc:8888` (in a K8s cluster for example). The env vars used in this example would be

```
export KO_DOCKER_REPO=my-registry.my-namespace.svc:8888
export KO_DOCKER_REPO_HOST=localhost:5000
```

### DIFF
--- a/pkg/commands/options/publish.go
+++ b/pkg/commands/options/publish.go
@@ -33,10 +33,10 @@ type PublishOptions struct {
 	// In normal ko usage, this is populated with the value of $KO_DOCKER_REPO.
 	DockerRepo string
 
-	// DockerRepoHost configures the host at which the registry can be reached. This can
+	// DockerRepoAddress configures the host at which the registry can be reached. This can
 	// differ from DockerRepo if the registry is talked to through a tunnel for example.
 	// If empty, this will be inferred from DockerRepo.
-	DockerRepoHost string
+	DockerRepoAddress string
 
 	// LocalDomain overrides the default domain for images loaded into the local Docker daemon. Use with Local=true.
 	LocalDomain string
@@ -78,8 +78,8 @@ func AddPublishArg(cmd *cobra.Command, po *PublishOptions) {
 	if dockerRepo, exists := os.LookupEnv("KO_DOCKER_REPO"); exists {
 		po.DockerRepo = dockerRepo
 	}
-	if dockerRepoHost, exists := os.LookupEnv("KO_DOCKER_REPO_HOST"); exists {
-		po.DockerRepoHost = dockerRepoHost
+	if dockerRepoHost, exists := os.LookupEnv("KO_DOCKER_REPO_ADDRESS"); exists {
+		po.DockerRepoAddress = dockerRepoHost
 	}
 
 	cmd.Flags().StringSliceVarP(&po.Tags, "tags", "t", []string{"latest"},
--- a/pkg/commands/options/publish.go
+++ b/pkg/commands/options/publish.go
@@ -33,6 +33,11 @@ type PublishOptions struct {
 	// In normal ko usage, this is populated with the value of $KO_DOCKER_REPO.
 	DockerRepo string
 
+	// DockerRepoHost configures the host at which the registry can be reached. This can
+	// differ from DockerRepo if the registry is talked to through a tunnel for example.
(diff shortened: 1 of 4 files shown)


imjasonh:
Nice.

> **Note 2:** Will think about tests if this is deemed useful and we want to move forward with it.

This seems like a good candidate for a test in [kind-e2e.yaml](https://github.com/google/ko/blob/main/.github/workflows/kind-e2e.yaml) that runs an in-registry repo and tunnels to it. This also makes it a useful runnable example we can point to from the README.

markusthoemmes:
> This seems like a good candidate for a test in kind-e2e.yaml that runs an in-registry repo and tunnels to it. This also makes it a useful runnable example we can point to from the README.

Done that!

pkg/commands/options/publish.go
```diff
@@ -73,6 +78,9 @@ func AddPublishArg(cmd *cobra.Command, po *PublishOptions) {
 	if dockerRepo, exists := os.LookupEnv("KO_DOCKER_REPO"); exists {
 		po.DockerRepo = dockerRepo
 	}
+	if dockerRepoHost, exists := os.LookupEnv("KO_DOCKER_REPO_HOST"); exists {
```
jonjohnsonjr, line 81: `KO_DOCKER_REPO_HOST` doesn't feel like the right name for this, but I don't think I understand this well enough to suggest a better name.

pkg/publish/options.go
```diff
@@ -120,3 +122,24 @@ func Insecure(b bool) Option {
 		return nil
 	}
 }
+
+// WithExplicitRepoHost is a functional option for pinning the repository to a specific
+// host, for example if the repository is talked to through a local tunnel.
+func WithExplicitRepoHost(host string) Option {
```
jonjohnsonjr, line 128: This feels like a very niche option -- can you use `WithTransport` to do this instead?

pkg/commands/options/publish.go
markusthoemmes: Totally open to other possibilities. I even thought about using some weird syntax for `KO_DOCKER_REPO`, but that seemed more dangerous.

halvards: Is this like `curl --resolve`? If so, how about something with `RESOLVE` in the name?

And since the `DialContext` argument allows for host+port, how about `ADDRESS` instead of `HOST`?

E.g., `KO_REPO_RESOLVE_ADDRESS`

markusthoemmes: @halvards yeah indeed, it's like `--resolve`

pkg/publish/options.go
markusthoemmes: Done.

pkg/commands/options/publish.go
```diff
@@ -73,6 +78,9 @@ func AddPublishArg(cmd *cobra.Command, po *PublishOptions) {
 	if dockerRepo, exists := os.LookupEnv("KO_DOCKER_REPO"); exists {
 		po.DockerRepo = dockerRepo
 	}
+	if dockerRepoHost, exists := os.LookupEnv("KO_DOCKER_REPO_ADDRESS"); exists {
```
jonjohnsonjr, line 81: Can the same thing be achieved with `HTTP_PROXY`?

markusthoemmes: I don't **think** so, since HTTP_PROXY should be a completely different machinery. I can try it out tho. Do you have something specific in mind, or just setting HTTP_PROXY to the "actual" location of the registry (the value I'd set `KO_DOCKER_REPO_ADDRESS` to in my world)?

markusthoemmes: Tried it, doesn't work as expected.

PR #519, opened 2021-11-25
Update golangci-lint and fix issues

markusthoemmes:
The version currently used is kinda old, so this bumps it and fixes an existing warning that was annoying me locally, while developing with a newer version 😅


imjasonh (approved):
Thanks!

PR #520, opened 2021-12-02
Remove special-case logic for setting Windows entrypoint paths

imjasonh:
This is apparently not needed, as Windows container runtimes will also
support values using standard Linux paths.

Passing Windows e2e tests should ensure this isn't a regression.


imjasonh:
Glad I added the e2e test, seems like we'd still need a special-case. Dropping for now.

PR #522, opened 2021-12-05
Use default `completion` command provided by cobra.

darklore:
`cobra` provides default `completion` subcommand.
It supports bash, zsh, fish, and powershell by default.


imjasonh (approved):
Nice, thanks for this!

imjasonh:
In addition to the CLA, I think you'll also have to regenerate the docs, `./hack/update-codegen.sh`

darklore:
Thank you for review.
I updated docs by `./hack/update-cocegen.sh`

imjasonh (approved):

PR #524, opened 2021-12-06
Set config creation time without mutating

imjasonh:
This also sets layers' history creation times to the relevant times set in the tar files.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -714,6 +714,7 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 		History: v1.History{
 			Author:    "ko",
 			CreatedBy: "ko build " + ref.String(),
+			Created:   g.kodataCreationTime,
 			Comment:   "kodata contents, at $KO_DATA_PATH",
 		},
 	})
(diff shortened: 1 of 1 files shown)


pkg/build/gobuild.go
```diff
@@ -727,19 +729,13 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 		cfg.Config.Labels[k] = v
 	}
 
+	cfg.Created = g.creationTime
```
imjasonh, line 732: Should this only be if `g.creationTime != empty` like before?

jonjohnsonjr: I wanna say yes.

pkg/build/gobuild.go
```diff
@@ -803,9 +799,9 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 	return si, nil
 }
 
-func buildLayer(appPath, file string, platform *v1.Platform) (v1.Layer, error) {
+func buildLayer(appPath, file string, creationTime v1.Time, platform *v1.Platform) (v1.Layer, error) {
 	// Construct a tarball with the binary and produce a layer.
-	binaryLayerBuf, err := tarBinary(appPath, file, v1.Time{}, platform)
+	binaryLayerBuf, err := tarBinary(appPath, file, creationTime, platform)
```
jonjohnsonjr, line 804: Was this always `v1.Time{}`?

imjasonh: Looks like it?

https://github.com/google/ko/commit/ee23538378722e060a2f7c7800f226e0b82e09e7#diff-1122f9e78a67758fe70c136cfe87ae6ea4a93ec4dc81a422bfbe57d2321ba868L619

jonjohnsonjr: Seems like a mistake, honestly. That PR was for adding kodata timestamps, and the binary timestamp wouldn't benefit from that change at all.

Want to just drop the creationTime param from `tarBinary`? Otherwise this is a behavior change.

jonjohnsonjr (approved):

Issue #192, opened 2020-09-24
Investigate fanning out multi-platform builds


jonjohnsonjr:
I have a sneaking suspicion that it will be slower to try to do multiple `go build`s for the same target with different platforms due to build cache invalidation, but the go compiler might have improved since I last experimented with this.

PR #527, opened 2021-12-08
Build each platform concurrently

jonjohnsonjr:
At least on my machine, this is slightly faster on a cold build and
saves significant time on a warm build.

Fixes https://github.com/google/ko/issues/192

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -48,6 +48,7 @@ import (
 	"github.com/sigstore/cosign/pkg/oci/signed"
 	"github.com/sigstore/cosign/pkg/oci/static"
 	ctypes "github.com/sigstore/cosign/pkg/types"
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/tools/go/packages"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/build/gobuild.go
```diff
@@ -872,9 +873,12 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIn
 		return nil, err
 	}
 
+	errg, ctx := errgroup.WithContext(ctx)
```
mattmoor, line 876: I think we want this to share it's threadpool with the `--jobs` pool, right?

jonjohnsonjr: Plumbed this down into `buildOne`

pkg/commands/resolver.go
```diff
@@ -141,11 +141,6 @@ func makeBuilder(ctx context.Context, bo *options.BuildOptions) (*build.Caching,
 		return nil, err
 	}
 
-	if bo.ConcurrentBuilds == 0 {
-		bo.ConcurrentBuilds = runtime.GOMAXPROCS(0)
-	}
-	innerBuilder = build.NewLimiter(innerBuilder, bo.ConcurrentBuilds)
```
mattmoor, line 147: is this now dead?

jonjohnsonjr: It is, but `build.NewLimiter` is public so I'm somewhat reluctant to just delete it.

pkg/build/gobuild.go
```diff
@@ -123,6 +131,7 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 			buildToDiff: map[string]buildIDToDiffID{},
 			diffToDesc:  map[string]diffIDToDescriptor{},
 		},
+		semaphore: semaphore.NewWeighted(int64(gbo.jobs)),
```
mattmoor, line 134: Hmm, do we `Open()` once per `ko` invocation or once per build? 🤔

mattmoor (dismissed):

pkg/build/gobuild.go
jonjohnsonjr: It's once per "builder", which means the `--jobs` isn't shared across build _configs_... which is annoying, but I'm happy to fix that if someone actually complains...

imjasonh (approved):

PR #536, opened 2021-12-13
Support osversion when selecting base images

### LINKED ISSUE #534: ko needs an --osversion flag
Discovered while testing https://github.com/google/ko/pull/533

The `golang:1.17` base image provides platform-specific Linux and Windows images, and even a number of different Windows images for different Windows versions:

```
$ crane manifest golang:1.17 | jq '.manifests[].platform'                                                                                                                                                                                 
...
{
  "architecture": "amd64",
  "os": "windows",
  "os.version": "10.0.20348.350"
}
{
  "architecture": "amd64",
  "os": "windows",
  "os.version": "10.0.17763.2300"
}
{
  "architecture": "amd64",
  "os": "windows",
  "os.version": "10.0.14393.4770"
}
```

For GitHub Actions to work, `ko` needs to select the _middle_ one, for `os.version: 10.0.17763.2300` (or any `10.0.17763`, probably). But `--platform=amd64/windows` will naively select the first one, and produce an image that doesn't run on GitHub Actions' `windows-latest`. `--platform=all` would work, it would simply build more images than it needs, including one for the correct `10.0.17763` image. Even if the correct image was listed first, there's no guarantee that `golang` won't change the order and break us, so we should be explicit about which one we match and build on.

This is similar in spirit to https://github.com/google/go-containerregistry/pull/1173 which added an `--osversion` flag to `crane` to take the `os.version` into account when matching a platform. `ko`'s support should work in much the same way.

@jonjohnsonjr

imjasonh:
Fixes #534 

This builds on #533 to add an option to the `--platform` flag to explicitly select a platform by osversion, which happens to be necessary to select the correct `golang:1.17` base image for the GitHub Actions runner environment.

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -61,20 +61,19 @@ func gobuildOptions(bo *options.BuildOptions) ([]build.Option, error) {
 		return nil, err
 	}
 
-	platform := bo.Platform
-	if platform == "" {
-		platform = "linux/amd64"
+	if bo.Platform == "" {
+		bo.Platform = "linux/amd64"
 
 		goos, goarch, goarm := os.Getenv("GOOS"), os.Getenv("GOARCH"), os.Getenv("GOARM")
 
 		// Default to linux/amd64 unless GOOS and GOARCH are set.
 		if goos != "" && goarch != "" {
-			platform = path.Join(goos, goarch)
+			bo.Platform = path.Join(goos, goarch)
 		}
 
 		// Use GOARM for variant if it's set and GOARCH is arm.
 		if strings.Contains(goarch, "arm") && goarm != "" {
-			platform = path.Join(platform, "v"+goarm)
+			bo.Platform = path.Join(bo.Platform, "v"+goarm)
 		}
 	} else {
 		// Make sure these are all unset
@@ -86,8 +85,8 @@ func gobuildOptions(bo *options.BuildOptions) ([]build.Option, error) {
 	}
 
 	opts := []build.Option{
-		build.WithBaseImages(getBaseImage(platform, bo)),
-		build.WithPlatforms(platform),
+		build.WithBaseImages(getBaseImage(bo)),
+		build.WithPlatforms(bo.Platform),
 		build.WithJobs(bo.ConcurrentBuilds),
 	}
 	if creationTime != nil {
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -35,7 +35,7 @@ jobs:
         export PLATFORM=${GOOS}/${GOARCH}
 
         if [[ "${{ matrix.platform }}" == "windows-latest" ]]; then
-          OSVERSION="10.0.17763.2300"
+          OSVERSION="10.0.17763.2366"
           PLATFORM=${PLATFORM}:${OSVERSION}
           export KO_DEFAULTBASEIMAGE=mcr.microsoft.com/windows/nanoserver:1809
         else
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,7 +1,7 @@
 name: Basic e2e test
 
 on:
-  pull_request: 
+  pull_request:
     branches: ['main']
 
 jobs:
@@ -35,9 +35,8 @@ jobs:
         export PLATFORM=${GOOS}/${GOARCH}
 
         if [[ "${{ matrix.platform }}" == "windows-latest" ]]; then
-          # Always use the nanoserver base image, which matches the Windows
-          # version used by the GitHub Actions Windows runner.
-          rm .ko.yaml
+          OSVERSION="10.0.17763.2300"
+          PLATFORM=${PLATFORM}:${OSVERSION}
           export KO_DEFAULTBASEIMAGE=mcr.microsoft.com/windows/nanoserver:1809
         else
           # Explicitly test multiple platform builds (a subset of what's in the base!)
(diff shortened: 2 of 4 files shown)


pkg/commands/resolver.go
```diff
@@ -85,9 +84,13 @@ func gobuildOptions(bo *options.BuildOptions) ([]build.Option, error) {
 		}
 	}
 
+	if multiplatform := bo.Platform == "all" || strings.Contains(bo.Platform, ","); multiplatform && bo.OSVersion != "" {
+		return nil, fmt.Errorf("cannot use --osversion with --platform=all or multiple platforms")
```
jonjohnsonjr, line 88: :/

imjasonh: Yeah, I don't love it. If we wanted to support this I think we should do it in `--platform` somehow, but I don't have a great idea for how to express that. I don't know if anybody has a need for it right now, I only need it to build a single platform+osversion.

imjasonh:
Experimenting with the platform syntax used in https://github.com/estesp/mquery where platform is `os[/arch[/variant]][:osversion]`, which would allows us to 1.) get away without a new flag 🎉 , and 2.) allow specifying multiple different osversions in `--platform` 🎉 🎉 

This works:

```
$ go run ./ build ./ --platform=windows/amd64:10.0.17763.2300
...
gcr.io/imjasonh/ko-98b8c7facdad74510a7cae0cd368eb4e@sha256:65977075f677a3ead974261f523d8ffff894074695e5b65b13ff9d19a820cc66
```

This currently panicks:

```
$ go run ./ build ./ --platform=windows/amd64:10.0.17763.2300,windows/amd64:10.0.14393.4770
...
goroutine 25 [running]:
github.com/google/go-containerregistry/pkg/v1/partial.Descriptor({0x0, 0x0})
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/partial/with.go:310 +0x7f
github.com/google/go-containerregistry/pkg/v1/mutate.computeDescriptor({{0x0, 0x0}, {{0x0, 0x0}, 0x0, {{0x0, 0x0}, {0x0, 0x0}}, {0x0, ...}, ...}})
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:30 +0x33
github.com/google/go-containerregistry/pkg/v1/mutate.(*index).compute(0xc000190850)
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:115 +0x265
github.com/google/go-containerregistry/pkg/v1/mutate.(*index).Digest(0x0)
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:189 +0x1e
github.com/google/ko/pkg/publish.pushResult.func1({0x1a23b08, 0xc0001df260}, {0x3805b28, 0xc00028c580})
	/Users/jasonhall/git/ko/pkg/publish/default.go:139 +0x383
github.com/sigstore/cosign/pkg/oci/walk.SignedEntity.func1({0x1a23b08, 0xc0001df260}, {0x3805b28, 0xc00028c580})
	/Users/jasonhall/git/ko/vendor/github.com/sigstore/cosign/pkg/oci/walk/walk.go:35 +0x31
github.com/sigstore/cosign/pkg/oci/mutate.Map({0x1a24320, 0xc000528a40}, {0x3805b28, 0xc00028c580}, 0xc0000d5988)
	/Users/jasonhall/git/ko/vendor/github.com/sigstore/cosign/pkg/oci/mutate/map.go:48 +0x90
github.com/sigstore/cosign/pkg/oci/walk.SignedEntity({0x1a24320, 0xc000528a40}, {0x3805b28, 0xc00028c580}, 0x10ea44a)
	/Users/jasonhall/git/ko/vendor/github.com/sigstore/cosign/pkg/oci/walk/walk.go:34 +0x49
github.com/google/ko/pkg/publish.pushResult({0x1a24320, 0xc000528a40}, {{{0x0, {0xc0000fc840, 0x6}}, {0xc0000fc847, 0x2c}}, {0xc0000fc874, 0x6}, {0xc0000fc840, ...}}, ...)
	/Users/jasonhall/git/ko/pkg/publish/default.go:179 +0x2a8
github.com/google/ko/pkg/publish.(*defalt).Publish(0xc0001c0070, {0x1a24320, 0xc000528a40}, {0x3805aa0, 0xc00028c580}, {0xc0000341c0, 0x19})
	/Users/jasonhall/git/ko/pkg/publish/default.go:220 +0xfec
github.com/google/ko/pkg/publish.(*multiPublisher).Publish(0x1334be0, {0x1a24320, 0xc000528a40}, {0x3805aa0, 0xc00028c580}, {0xc0000341c0, 0x19})
	/Users/jasonhall/git/ko/pkg/publish/multi.go:45 +0xff
github.com/google/ko/pkg/publish.(*caching).Publish.func1.1()
	/Users/jasonhall/git/ko/pkg/publish/shared.go:70 +0x3d
github.com/google/ko/pkg/publish.newFuture.func1()
	/Users/jasonhall/git/ko/pkg/publish/future.go:29 +0x2a
created by github.com/google/ko/pkg/publish.newFuture
	/Users/jasonhall/git/ko/pkg/publish/future.go:28 +0x89
exit status 2
```

I'll work on figuring that out while we decide if we like this syntax.

If we go with this format we should probably also get `crane` to use it, and drop its `--osversion` flag too.

imjasonh:
Hey, good news, this isn't a new panic 🌮  https://github.com/google/ko/issues/544

imjasonh:
Any more concerns with this?

jonjohnsonjr (approved):

PR #540, opened 2021-12-14
cleanup: Remove publisher.Close method

imjasonh:
This was nearly always only implemented `return nil`, and the one case it wasn't (tarball publishing), it didn't do anything that needed to be done in `Close` anyway. This simplifies the interface, making it harder to accidentally forget to call `defer publisher.Close()` and wonder whether you're leaking resources.

This also removes a `bad practice` in pkg/publish/tarball.go which was only done because we were `defer`ring `Close()`.

### DIFF
--- a/pkg/publish/tarball.go
+++ b/pkg/publish/tarball.go
@@ -80,8 +80,6 @@ func (t *tar) Publish(_ context.Context, br build.Result, s string) (name.Refere
 
 	log.Printf("Saving %v", t.file)
 	if err := tarball.MultiRefWriteToFile(t.file, t.refs); err != nil {
-		// Bad practice, but we log  this here because right now we just defer the Close.
-		log.Printf("failed to save %q: %v", t.file, err)
 		return nil, err
 	}
 	log.Printf("Saved %v", t.file)
--- a/pkg/publish/tarball.go
+++ b/pkg/publish/tarball.go
@@ -78,6 +78,14 @@ func (t *tar) Publish(_ context.Context, br build.Result, s string) (name.Refere
 		t.refs[ref] = img
 	}
 
+	log.Printf("Saving %v", t.file)
+	if err := tarball.MultiRefWriteToFile(t.file, t.refs); err != nil {
+		// Bad practice, but we log  this here because right now we just defer the Close.
+		log.Printf("failed to save %q: %v", t.file, err)
+		return nil, err
+	}
+	log.Printf("Saved %v", t.file)
+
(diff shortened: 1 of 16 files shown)


imjasonh:
@halvards Is there any concern with this breaking anything in Skaffold?

pkg/publish/tarball.go
```diff
@@ -78,6 +78,12 @@ func (t *tar) Publish(_ context.Context, br build.Result, s string) (name.Refere
 		t.refs[ref] = img
 	}
 
+	log.Printf("Saving %v", t.file)
+	if err := tarball.MultiRefWriteToFile(t.file, t.refs); err != nil {
```
jonjohnsonjr, line 82: This makes writing to a tarball super slow :/

imjasonh: This is the same as what was in `Close`, how is this any different?

jonjohnsonjr: `ko resolve` et al will call `Publish` multiple times, so we would end up serializing and writing the tarball once for every image that gets published instead of just once per invocation.

jonjohnsonjr: https://www.joelonsoftware.com/2001/12/11/back-to-basics/#:~:text=code%20uses%20the-,Shlemiel%20the%C2%A0painter%E2%80%99s%20algorithm,-.%20Who%20is%20Shlemiel

imjasonh: Ahh that makes sense.

In that case, should we also have the defalt publisher enqueue images to publish, and only actually write them once the publisher is `Close`d? This implementation seems to be unique to the tarball publisher, but it requires changes to all the other implementations to account for it.

How does `ko resolve --tarball` even work? What ends up in the YAML? 🤔

jonjohnsonjr: > This implementation seems to be unique to the tarball publisher

Tarball is unique as a publisher in that it's impossible to do incremental writes to it because we have to serialize everything :/

> What ends up in the YAML?

What you'd expect.

Another option here would be to construct the tarball contents in a tempdir and tar it up once we're done, but we'd still need to know when we're done :/

imjasonh:
I'm fine closing this, but I think what I've learned is that `--tarball` mode is potentially under-utilized in general, and almost definitely under-exercised by any of us maintainers. Not supporting multi-platform images seems like a real gap (not to mention SBOMs, signatures, etc.). If it gives us any trouble in the future I'd like to consider deprecating and removing it in favor of `--oci-layout-path`.

Issue #535, opened 2021-12-13
SBOM generation fails on Windows


imjasonh:
https://github.com/google/ko/runs/4508967288?check_suite_focus=true#step:4:91

```
++ go run ./ publish ./ --platform=windows/amd64 --preserve-import-paths
2021/12/13 16:28:50 Using base mcr.microsoft.com/windows/nanoserver:1809 for github.com/google/ko
2021/12/13 16:28:50 Building github.com/google/ko for windows/amd64
C:\Users\RUNNER~1\AppData\Local\Temp\ko2792353245\out: not executable file
2021/12/13 16:29:33 Loading ko.local/github.com/google/ko:f16f6b389d4d4ccfc1db0524e5f8ec6b315fc6a4e080f0ffe17f5074aa4a6d2a
2021/12/13 16:29:34 Loaded ko.local/github.com/google/ko:f16f6b389d4d4ccfc1db0524e5f8ec6b315fc6a4e080f0ffe17f5074aa4a6d2a
2021/12/13 16:29:34 Adding tag latest
2021/12/13 16:29:34 Added tag latest
```

The binary built by `go build` on Windows doesn't seem to be executable when we run `go version -m` on it, which causes SBOM generation to fail. Interestingly, this doesn't seem to block pushing the rest of the image.

cc @mattmoor

mattmoor:
Why didn't CI catch this? 🤔

imjasonh:
`ko publish` doesn't fail, so e2e tests on Windows didn't care.

This sounds like it may also be a gap in e2e tests, we should consider adding a `cosign download sbom` to the mix. That would probably require us to also push to a local registry, instead of just to `ko.local` like we do today.

mattn:
Container image does not exists in the registory at the first time. So cosigin download sbom will fail.

mattn:
Ah, sorry. It is registory of base image. However I don't know why this fail always.

```
C:\>cosign download sbom gcr.io/distroless/static:latest
Error: image not found in registry
main.go:46: error during command execution: image not found in registry
```

PR #543, opened 2021-12-14
chmod the Go binary before SBOMing it

imjasonh:
Fixes https://github.com/google/ko/issues/535 (🤞)

I don't know why this would be necessary, `go build` should produce an executable binary already, but `go version -m` on Windows seems to disagree.

Let's see what the e2e test on Windows says.


imjasonh:
Nope: https://github.com/google/ko/runs/4521735128?check_suite_focus=true#step:4:99

```
C:\Users\RUNNER~1\AppData\Local\Temp\ko4218645900\out: not executable file
```

PR #545, opened 2021-12-14
Fix tarball publisher ctx handling

jonjohnsonjr:
Previously, we'd use the ctx from the Build invocation, but because
remote stashes the remote.WithContext option, when we later go to fetch
the actual base image to write it out to a tarball, it will fail because
the Build ctx was cancelled.

For whatever reason, NewGo takes a ctx, so this change propagates the
badness from remote so we can use the NewGo ctx when we fetch base
images.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -70,6 +70,7 @@ type platformMatcher struct {
 }
 
 type gobuild struct {
+	ctx                  context.Context
 	getBase              GetBase
 	creationTime         v1.Time
 	kodataCreationTime   v1.Time
@@ -90,6 +91,7 @@ type gobuild struct {
 type Option func(*gobuildOpener) error
 
 type gobuildOpener struct {
+	ctx                  context.Context
 	getBase              GetBase
 	creationTime         v1.Time
 	kodataCreationTime   v1.Time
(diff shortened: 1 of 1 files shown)


pkg/build/gobuild.go
```diff
@@ -826,7 +830,9 @@ func updatePath(cf *v1.ConfigFile, appPath string) {
 // Build implements build.Interface
 func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 	// Determine the appropriate base image for this import path.
-	baseRef, base, err := g.getBase(ctx, s)
+	// We use the overall gobuild.ctx because the Build ctx gets cancelled
```
imjasonh, line 833: Why does the Build ctx get cancelled early?

jonjohnsonjr: https://github.com/google/ko/blob/a56047a2a1d500a646393b4a62d1d509bd5575e4/pkg/commands/resolver.go#L318

https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext

We want to cancel outstanding builds if we encounter an error, so we use `errgroup.WithContext`, but the ctx is also cancelled once this returns: https://github.com/google/ko/blob/a56047a2a1d500a646393b4a62d1d509bd5575e4/pkg/commands/resolver.go#L415

imjasonh (approved):

Issue #544, opened 2021-12-14
Panic building for multiple explicit platforms


imjasonh:
Running from HEAD at 2ba70fc75bd94147b03e9c24d3e4c5f56d02edfc:

```
$ go run ./ build ./ --platform=linux/amd64,linux/arm64
...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x14a1c1f]

goroutine 93 [running]:
github.com/google/go-containerregistry/pkg/v1/partial.Descriptor({0x0, 0x0})
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/partial/with.go:310 +0x7f
github.com/google/go-containerregistry/pkg/v1/mutate.computeDescriptor({{0x0, 0x0}, {{0x0, 0x0}, 0x0, {{0x0, 0x0}, {0x0, 0x0}}, {0x0, ...}, ...}})
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:30 +0x33
github.com/google/go-containerregistry/pkg/v1/mutate.(*index).compute(0xc0001c4460)
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:115 +0x265
github.com/google/go-containerregistry/pkg/v1/mutate.(*index).Digest(0xc0005a0000)
	/Users/jasonhall/git/ko/vendor/github.com/google/go-containerregistry/pkg/v1/mutate/index.go:189 +0x1e
github.com/google/ko/pkg/publish.pushResult.func1({0x1a23a88, 0xc0002b4e70}, {0x9ced0a0, 0xc0000bf5c0})
	/Users/jasonhall/git/ko/pkg/publish/default.go:139 +0x383
github.com/sigstore/cosign/pkg/oci/walk.SignedEntity.func1({0x1a23a88, 0xc0002b4e70}, {0x9ced0a0, 0xc0000bf5c0})
	/Users/jasonhall/git/ko/vendor/github.com/sigstore/cosign/pkg/oci/walk/walk.go:35 +0x31
github.com/sigstore/cosign/pkg/oci/mutate.Map({0x1a242a0, 0xc00048b2c0}, {0x9ced0a0, 0xc0000bf5c0}, 0xc0004e9988)
[...]

PR #546, opened 2021-12-14
Filter out nil addendum before constructing index.

mattmoor:
In the PR where we added concurrency we used a fixed length array to store
addendum to preserve the ordering from the base image when constructing the
final index.  However, with `--platform=...` this list may be filtered, which
gives us `nil` entries in our addendum.

This filters `nil` entries prior to constructing the index.

Fixes: https://github.com/google/ko/issues/544

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -879,7 +879,6 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 	return res, nil
 }
 
-// TODO(#192): Do these in parallel?
 func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIndex) (oci.SignedImageIndex, error) {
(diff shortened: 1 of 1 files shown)


mattmoor:
If you have opinions on where I can easily add test coverage for this, I'm happy to, but this fixes @imjasonh repro for me

imjasonh (approved):
LGTM, but how much do we care about maintaining the original ordering? If we didn't we could just `append` and then sort for determinism. I guess there's still annoying selection behavior in runtimes if there are multiple matching platforms 🤮

imjasonh:
> If you have opinions on where I can easily add test coverage for this, I'm happy to, but this fixes @imjasonh repro for me

Adding something to e2e.yaml to just build a two-platform image would be fine by me. We don't even have to run it.

mattmoor:
Ok added something here: https://github.com/google/ko/pull/547

PR #548, opened 2021-12-15
Fix #530 Hanging behavior with ASDF installed Go on Darwin

### LINKED ISSUE #530: ko hangs when used with ASDF installed Go
I am able to reproduce either a bug, or a misconfiguration that could perhaps use a better error message of some sort.

The reproduction is here at [StevenACoffman/repro](https://github.com/StevenACoffman/repro).

### Repro 

If you clone that repo and run:

```
export GOPRIVATE=github.com/StevenACoffman
cd sub
ko publish -B --bare --local --platform=linux/amd64 --push=false .
```
The terminal will just hang there until you hit Ctrl-c.

At which point, it will then say:
```
^CError: error creating builder: 'builds': entry #0 does not contain a valid local import path (./.) for directory (.): err: signal: interrupt: stderr:
2021/12/12 21:00:19 error during command execution:error creating builder: 'builds': entry #0 does not contain a valid local import path (./.) for directory (.): err: signal: interrupt: stderr:
```

However, if you delete the `.ko.yaml` file in that directory and run it again, things work as you'd expect.
```
rm .ko.yaml
ko publish -B --bare --local --platform=linux/amd64 --push=false .
```

I have tried a number of different possible values for "main" and "dir" and "id" in the `.ko.yaml` file, including omitting them altogether. For instance, setting the `main:` to the values `github.com/StevenACoffman/repro/sub`, `.`, and `main` don't seem to affect this hanging behavior.

[`.ko.yaml`](https://github.com/StevenACoffman/repro/blob/main/sub/.ko.yaml)  contents:
```
# KO_DOCKER_REPO should be either kind.local
# (for KinD) or ko.local

defaultBaseImage: gcr.io/distroless/static
builds:
  # You can have multiple builds defined as a yaml list
  -
    # ID of the build.
    # Defaults to the project name.
    id: "my-build"

    # Path to project's (sub)directory containing Go code.
    # This is the working directory for the Go build command(s).
    # Default is `.`.
    dir: .

    # Path to main.go file or main package.
    # Notice: when used with `gomod.proxy`, this must be a package.
    #
    # Default is `.`.
    main: github.com/StevenACoffman/repro/sub


    ldflags:
      - -s 
      - -w
      - -X main.AppName=Working

    # Custom environment variables to be set during the builds.
    # Default is empty.
    env:
      - CGO_ENABLED=0
``` 

Any help or tips would be greatly appreciated!

StevenACoffman:
So this resolves the issue with ko hanging  on Darwin when Go is installed using ASDF, however, I'm not sure if this is the final fix that still works for everyone else.

@jonjohnsonjr 
Signed-off-by: Steve Coffman <steve@khanacademy.org>

### DIFF
--- a/pkg/commands/options/build.go
+++ b/pkg/commands/options/build.go
@@ -180,8 +180,11 @@ func createBuildConfigMap(workingDirectory string, configs []build.Config) (map[
 		// local import paths, therefore add a "./" equivalent as a prefix to
 		// the constructured import path
 		localImportPath := fmt.Sprint(".", string(filepath.Separator), path)
-
-		pkgs, err := packages.Load(&packages.Config{Mode: packages.NeedName, Dir: baseDir}, localImportPath)
+		dir := baseDir
+		if filepath.Clean(dir) == "." {
+			dir = ""
+		}
+		pkgs, err := packages.Load(&packages.Config{Mode: packages.NeedName, Dir: dir}, localImportPath)
 		if err != nil {
 			return nil, fmt.Errorf("'builds': entry #%d does not contain a valid local import path (%s) for directory (%s): %w", i, localImportPath, baseDir, err)
 		}
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -161,9 +161,13 @@ func NewGo(ctx context.Context, dir string, options ...Option) (Interface, error
 }
 
 func (g *gobuild) qualifyLocalImport(importpath string) (string, error) {
+	dir := g.dir
+	if filepath.Clean(g.dir) == "." {
+		dir = ""
+	}
 	cfg := &packages.Config{
 		Mode: packages.NeedName,
-		Dir:  g.dir,
+		Dir:  dir,
 	}
 	pkgs, err := packages.Load(cfg, importpath)
 	if err != nil {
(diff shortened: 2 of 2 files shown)


pkg/commands/options/build.go
```diff
@@ -181,7 +181,7 @@ func createBuildConfigMap(workingDirectory string, configs []build.Config) (map[
 		// the constructured import path
 		localImportPath := fmt.Sprint(".", string(filepath.Separator), path)
 
-		pkgs, err := packages.Load(&packages.Config{Mode: packages.NeedName, Dir: baseDir}, localImportPath)
+		pkgs, err := packages.Load(&packages.Config{Mode: packages.NeedName, Dir: ""}, localImportPath)
```
jonjohnsonjr, line 184: We'll want to do the same dance here as everywhere else, I think.

StevenACoffman: Done.

StevenACoffman (commented):
Are there places where we actually want `dir` to be `.` instead of `""`? 

There are still a few more references to this `g.dir` where we don't do the song and dance like:
https://github.com/google/ko/blob/1ec35c80cf6964d950868a626374743d75cae169/pkg/build/gobuild.go#L690

Also, I wonder about just lifting the song and dance to where the creation of the struct in the first place:
https://github.com/google/ko/blob/1ec35c80cf6964d950868a626374743d75cae169/pkg/build/gobuild.go#L131

Or here:
https://github.com/google/ko/blob/1ec35c80cf6964d950868a626374743d75cae169/pkg/build/gobuild.go#L151

pkg/commands/options/build.go
jonjohnsonjr: I think you need to push still?

StevenACoffman: 🤦 😅 Riiiiight.

jonjohnsonjr:
> There are still a few more references to this g.dir where we don't do the song and dance 

I think it's fine to leave those alone for now. This seems like a weird interaction with `packages.Load` that we need to workaround, but just execing go directly works as expected. I'll want to refactor some of this stuff regardless, so we can just merge this to fix things until I can look at it more closely.

jonjohnsonjr (approved):

Issue #542, opened 2021-12-14
Improve errors/warnings for unnecessary .ko.yaml settings


imjasonh:
From https://github.com/google/ko/issues/530#issuecomment-992110575

> A few points to note:
> 
> - The `-B` (`--base-import-paths`) and `--bare` flags are not designed to be used together, and in this case, `-B` takes precedence: https://github.com/google/ko#naming-images
> 
> - The `main` field in `.ko.yaml` is a directory, not a full Go import path.
> 
> - If your `.ko.yaml` and `go.mod` are in the same directory as your main package, and if this directory is also your working directory, then you can leave out `dir` and `main`.
> 
> - `ko` sets `CGO_ENABLED=0` by default.

Each of these could be warnings or errors to help make `.ko.yaml` easier to use and less needlessly verbose.

imjasonh:
Another one:

- `ko publish --local` is confusing if `KO_DOCKER_REPO` is set to anything besides `ko.local`, and even then it'd be nice not to have both options. Maybe we should just deprecate `--local`, but even if we don't we should at least log a warning about it.

imjasonh:
For all of these I think we should log a warning, and note that they might become failures in a future release.

Then, after we cut a release ([v0.10](https://github.com/google/ko/milestone/2)), we can make them failures.

imjasonh:
Another:

- `ko build ./ --platform=all,linux/arm64` only builds for `linux/arm64` and doesn't even log a warning. This should probably just fail if you have `all` and anything else.

PR #551, opened 2021-12-16
Pre-parse platform string with StringSliceVar

### LINKED ISSUE #537: --platform should be a StringSliceVar flag
Today, `ko build ./ --platform=linux/amd64,linux/arm64` means "build on these two platforms", and `ko` splits on the `,` character to determine this.

If `--platform` was a [`StringSliceVar`](https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar), the existing behavior would be retained, but it would also allow users to specify `ko build ./ --platform=linux/amd64 --platform=linux/arm64`

The existing semantics should be retained:
- if any value is `all`, there should only be that one value
- if any value is `linux`, all `linux/*` platforms should be matched
- if any value is `linux/arm64/v7`, the `v7` CPU variant should be matched

wilsonehusin:
Hello! First time contributing to `ko` here 👋🏼 

- The variable `BuildOptions.Platform` was renamed to `BuildOptions.Platforms` to reflect the actual data type, though I'm not sure how comfortable are we in breaking changes of (supposedly) public structs.
- I took somewhat of a "persisting structure all the way down" approach which has a side effect of `platformMatcher.spec` and `gobuildOpener.platforms` are now of type `[]string` (instead of `string`).

Happy to discuss and make changes as desired by maintainers!

Closes #537

### DIFF
--- a/pkg/build/options.go
+++ b/pkg/build/options.go
@@ -89,9 +89,9 @@ func WithConfig(buildConfigs map[string]Config) Option {
 //
 // platform = <os>[/<arch>[/<variant>]]
 // allowed = all | platform[,platform]*
-func WithPlatforms(platforms string) Option {
+func WithPlatforms(platforms []string) Option {
 	return func(gbo *gobuildOpener) error {
-		gbo.platform = platforms
+		gbo.platforms = platforms
 		return nil
 	}
 }
--- a/pkg/build/options.go
+++ b/pkg/build/options.go
@@ -15,6 +15,8 @@
 package build
 
 import (
+	"strings"
+
 	v1 "github.com/google/go-containerregistry/pkg/v1"
 )
 
@@ -85,12 +87,20 @@ func WithConfig(buildConfigs map[string]Config) Option {
 
 // WithPlatforms is a functional option for building certain platforms for
 // multi-platform base images. To build everything from the base, use "all",
-// otherwise use a comma-separated list of platform specs, i.e.:
+// otherwise use a list of platform specs, i.e.:
 //
 // platform = <os>[/<arch>[/<variant>]]
-// allowed = all | platform[,platform]*
-func WithPlatforms(platforms []string) Option {
+// allowed = "all" | []string{platform[,platform]*}
+//
+// Note: a string of comma-separated platforms (i.e. "platform[,platform]*")
+// has been deprecated and only exist for backwards compatibility reasons,
+// which will be removed in the future.
+func WithPlatforms(platforms ...string) Option {
 	return func(gbo *gobuildOpener) error {
+		if len(platforms) == 1 {
+			// TODO: inform users that they are using deprecated flow?
+			platforms = strings.Split(platforms[0], ",")
+		}
 		gbo.platforms = platforms
 		return nil
 	}
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -61,11 +61,13 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 		//
 		// Platforms can be comma-separated if we only want a subset of the base
 		// image.
-		multiplatform := bo.Platform == "all" || strings.Contains(bo.Platform, ",")
-		if bo.Platform != "" && !multiplatform {
+		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
+		selectiveMultiplatform := len(bo.Platforms) > 1
+		multiplatform := allPlatforms || selectiveMultiplatform
+		if len(bo.Platforms) > 0 && !multiplatform {
 			var p v1.Platform
 
-			parts := strings.Split(bo.Platform, ":")
+			parts := strings.Split(bo.Platforms[0], ":")
 			if len(parts) == 2 {
 				p.OSVersion = parts[1]
 			}
@@ -81,7 +83,7 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 				p.Variant = parts[2]
 			}
 			if len(parts) > 3 {
-				return nil, fmt.Errorf("too many slashes in platform spec: %s", bo.Platform)
+				return nil, fmt.Errorf("too many slashes in platform spec: %s", bo.Platforms[0])
 			}
 			ropt = append(ropt, remote.WithPlatform(p))
 		}
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -60,8 +60,7 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 		// otherwise we'll resolve it to the appropriate platform.
 		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
 
-		// Platforms can be comma-separated if we only want a subset of the base
-		// image.
+		// Platforms can be listed in a slice if we only want a subset of the base image.
 		selectiveMultiplatform := len(bo.Platforms) > 1
 
 		multiplatform := allPlatforms || selectiveMultiplatform
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -58,15 +58,18 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 
 		// Using --platform=all will use an image index for the base,
 		// otherwise we'll resolve it to the appropriate platform.
-		//
+		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
+
 		// Platforms can be comma-separated if we only want a subset of the base
 		// image.
-		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
 		selectiveMultiplatform := len(bo.Platforms) > 1
+
 		multiplatform := allPlatforms || selectiveMultiplatform
-		if len(bo.Platforms) > 0 && !multiplatform {
+		if !multiplatform {
 			var p v1.Platform
 
+			// There is _at least_ one platform specified at this point,
+			// because receiving "" means we would infer from GOOS/GOARCH.
 			parts := strings.Split(bo.Platforms[0], ":")
 			if len(parts) == 2 {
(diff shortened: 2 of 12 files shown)


pkg/commands/resolver.go
```diff
@@ -61,20 +61,22 @@ func gobuildOptions(bo *options.BuildOptions) ([]build.Option, error) {
...
+			envPlatform = path.Join(goos, goarch)
 		}
 
 		// Use GOARM for variant if it's set and GOARCH is arm.
 		if strings.Contains(goarch, "arm") && goarm != "" {
-			bo.Platform = path.Join(bo.Platform, "v"+goarm)
+			envPlatform = path.Join(envPlatform, "v"+goarm)
 		}
+
+		bo.Platforms = append(bo.Platforms, envPlatform)
```
imjasonh, line 79: We know `bo.Platforms` is empty at this point, I think it's a bit clearer to say `bo.Platforms = []string{envPlatform}`

pkg/commands/config.go
```diff
@@ -61,11 +61,13 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 		//
 		// Platforms can be comma-separated if we only want a subset of the base
 		// image.
-		multiplatform := bo.Platform == "all" || strings.Contains(bo.Platform, ",")
-		if bo.Platform != "" && !multiplatform {
+		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
+		selectiveMultiplatform := len(bo.Platforms) > 1
+		multiplatform := allPlatforms || selectiveMultiplatform
+		if len(bo.Platforms) > 0 && !multiplatform {
```
imjasonh, line 67: I like the new descriptive variable names!

I think `bo.Platforms` is always non-empty at this point, because we set it to some default if unset. Can this just be `if !multiplatform`? That makes the `bo.Platforms[0]` easier to understand, since we'll know there's only one value.

pkg/commands/config.go
```diff
@@ -61,11 +61,13 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 		//
 		// Platforms can be comma-separated if we only want a subset of the base
 		// image.
-		multiplatform := bo.Platform == "all" || strings.Contains(bo.Platform, ",")
-		if bo.Platform != "" && !multiplatform {
+		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
```
imjasonh, line 64: Do we already have code that fails if you pass `--platform=all,linux/arm64`? Or at least handles it by reducing it to `--platform=all`?

imjasonh (commented):
Thanks for this contribution, and sorry for the delay in reviewing. This looks mostly good, just a few smallish comments/questions.

pkg/commands/resolver.go
wilsonehusin: hmm yeah I think I thought too much about allocations when I wrote it 😆

pkg/commands/config.go
wilsonehusin: I'm not aware of such scenario when I was writing this PR. I thought about adding it to `parseSpec`, but it looked quite complicated, so I decided to preserve the original behavior. Happy to include that into this PR if you'd like to see it!

wilsonehusin (commented):
@imjasonh less than 24 hours is not "late" 😄 — thank you for reviewing! Revision incoming in a bit, though would like confirmation on the `--platform=all,linux/arm64` scenario.

pkg/commands/config.go
imjasonh: We don't have to include it in this PR, but I'd like to at least handle it better. Maybe another PR.

I've added it to the growing laundry list in https://github.com/google/ko/issues/542.

wilsonehusin:
The diff in Verify Codegen seems expected:

```diff
-      --platform strings         Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]][,platform]*
+      --platform string          Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]][,platform]*
```

Should I run `hack/update-codegen.sh`?

pkg/build/options.go
```diff
@@ -89,9 +89,9 @@ func WithConfig(buildConfigs map[string]Config) Option {
 //
 // platform = <os>[/<arch>[/<variant>]]
 // allowed = all | platform[,platform]*
-func WithPlatforms(platforms string) Option {
+func WithPlatforms(platforms []string) Option {
```
imjasonh, line 92: @jonjohnsonjr had the good idea that this change could be `WithPlatforms(platforms string...)` and not break existing users. (Thanks Jon!)

If you feel like doing this change in this PR that'd be great, if not I can do it in a followup before we release this.

Otherwise, lgtm!

imjasonh (approved):

pkg/build/options.go
wilsonehusin: Ah yes! I felt silly I didn't think of it in the first place. Fixing...

wilsonehusin: 🤔 actually, using `...string` would not break compile-time for existing users, but it would break the behavior, unless we do some sort of `strings.Spilt` here.

Do we want to add that logic as well? Something like:

```go
func WithPlatforms(platforms ...string) Option {
  if len(platforms) == 1 {
    platforms = strings.Split(platforms, ",")
  }
  ...
}
```

There is an itch on me that this code is waiting to be deleted, but happy to make changes if maintainers are ok with maintaining it 😄

pkg/commands/options/build.go
```diff
@@ -76,7 +76,7 @@ func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
 		"Disable optimizations when building Go code. Useful when you want to interactively debug the created container.")
 	cmd.Flags().StringVar(&bo.SBOM, "sbom", "spdx",
 		"The SBOM media type to use (none will disable SBOM synthesis and upload, also supports: spdx, go.version-m).")
-	cmd.Flags().StringVar(&bo.Platform, "platform", "",
+	cmd.Flags().StringSliceVar(&bo.Platforms, "platform", []string{},
 		"Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]][,platform]*")
```
halvards, line 80: The format here should be updated, to indicate that the value is no longer a comma-separated string.

pkg/commands/config.go
```diff
@@ -58,14 +58,19 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 
 		// Using --platform=all will use an image index for the base,
 		// otherwise we'll resolve it to the appropriate platform.
-		//
+		allPlatforms := len(bo.Platforms) == 1 && bo.Platforms[0] == "all"
+
 		// Platforms can be comma-separated if we only want a subset of the base
 		// image.
```
halvards, line 64: It'd be great to update this comment to reflect the new behavior.

pkg/commands/options/build.go
```diff
@@ -51,7 +51,7 @@ type BuildOptions struct {
 	ConcurrentBuilds     int
 	DisableOptimizations bool
 	SBOM                 string
-	Platform             string
+	Platforms            []string
```
halvards, line 54: This breaks [Skaffold's `ko` integration](https://skaffold.dev/docs/pipeline-stages/builders/ko/), so I'd be cautious about this sort of change in general - as you mentioned in the PR.

However, in this particular case I'm ok with it, as I'm closely involved with Skaffold and I'm aware of this breaking change. It also means that I can remove this hoop from the Skaffold codebase: https://github.com/GoogleContainerTools/skaffold/blob/v1.35.1/pkg/skaffold/build/ko/builder.go#L53

We should flag the breaking change somehow to anyone else who might be embedding ko though - there could be private repos/tools out there that would be impacted.

A more cautious approach would be to retain the `Platform` field with a note that it has been deprecated, and use its value if `Platforms` is empty.

pkg/commands/options/build.go
wilsonehusin: comma-separated string is still supported, so existing usage with commas _should_ still work (I tested locally, but I don't think there is bash-level test in the project?)

pkg/commands/config.go
wilsonehusin: ditto — comma-separated is still supported from CLI

pkg/commands/options/build.go
imjasonh: This PR was partially the reason I wanted #554 -- if there's a change that would make the CLI better, like this one, but that might break users of the Go package in small and easy-to-workaround ways, I think we should do it.

If you'd like more time to stage the change, we can keep `Platform` as `Deprecated` and special-case it for one release.

As an alternative, it looks like you could just change your code to `Platforms: `.KoArtifact.Platforms,` when you bump your `ko` dep to include this change. That's pretty straightforward (and cleaner anyway), and I imagine most callers would have a similar diff. We should call out the breaking change in release notes, definitely.

pkg/commands/options/build.go
halvards: TIL, thanks

pkg/commands/options/build.go
halvards: Yes, I agree. We don't need a phased approach in this case.

pkg/build/options.go
imjasonh: Yeah let's add the behavior you describe, for folks using `WithPlatforms("linux/amd64,linux/arm64")`, and document that this behavior will change in a future release. This is a fairly small concession we should make to existing library users.

wilsonehusin:
thanks for the review folks! :tada:

imjasonh:
Thanks for working on this, and sorry for the holiday-related slowness in reviewing it again. Looks like there's a build failure, otherwise LGTM.

wilsonehusin:
That was a silly mistake, apologies! Happy new year! :tada:

imjasonh (approved):

PR #568, opened 2022-01-24
Support fuzzy platform matching

imjasonh:
CI tests recently started failing because we require a Windows platform with exact osversion, and the osversion provided by `golang:1.17` no longer includes that exact value -- e.g: https://github.com/google/ko/runs/4916021150?check_suite_focus=true#step:4:98

This change makes `--platform` matching slightly fuzzy, especially in its handling of osversion, so you can specify `windows/amd64:10.0.17763` and get _any_ matching image (in `golang:1.17` there's only one).


imjasonh:
Well here's a fun one!

```console
$ crane manifest mcr.microsoft.com/windows/nanoserver:1809 | jq '.manifests[].platform'
{
  "architecture": "amd64",
  "os": "windows",
  "os.version": "10.0.17763.2458"
}
```

```console
$ crane manifest golang:1.17 | jq '.manifests[].platform'
...
{
  "architecture": "amd64",
  "os": "windows",
  "os.version": "10.0.17763.2452"
}
```

So there's no nanoserver base image that exactly matches the base image that golang:1.17 provides, as of this moment. And presumably random moments in the future.

I think I'm going to have to add the partial matching logic to get around this reliably. 🤷

imjasonh:
This now integrates https://github.com/google/go-containerregistry/pull/1261 (not yet merged), to support partial platform matching, so these all work:

```
ko build ./ --platform=windows/amd64:10.0.17763 # single matching image
ko build ./ --platform=windows/amd64:10.0.17763,windows/amd64:10.0.20348
ko build ./ --platform=windows/amd64 # two matching images
ko build ./ --platform=windows/amd64:10.0 # two matching images
```

Marking this as draft until the other PR settles, but I think this is a good improvement.

imjasonh:
The decision we came to was to keep partial platform matching inside ko's pre-existing partial platform matching, and just to teach it as little as possible about Windows osversion matching semantics to get by. Closing this.

Issue #560, opened 2021-12-22
Introduce log levels


halvards:
When using `ko` as a library, it can be helpful to have more control of the log output.

If a downstream tool enables debug or trace level logging, it could then propagate this setting to `ko`. Likewise, a downstream warning log level could translate to having `ko` only output warnings (e.g., with #542).

For instance, while developing features for #348 and https://github.com/GoogleContainerTools/skaffold/issues/6041, I often added debug log lines to `ko` that I removed before committing changes.

One approach would be to follow the approach in `ggcr`: https://github.com/google/go-containerregistry/blob/v0.7.0/pkg/logs/logs.go#L24-L31

This would also allow downstream tools to plug in their logging libraries by sending `ko` log output to another destination using [`log/Logger.SetOutput`](https://pkg.go.dev/log#Logger.SetOutput), like what `ko` does with `ggcr`: https://github.com/google/ko/blob/v0.9.3/cmd/ko/main.go#L39-L40

And just to be clear, I'm not suggesting introducing a logging library!

jonjohnsonjr:
I have some regrets about the global logging stuff in ggcr. I would be fine with adopting https://github.com/go-logr/logr if we want to expose this.

halvards:
What regrets do you have with the `ggcr` approach? I like the simplicity of it, which unsurprisingly aligns with `ko` :smile:.

On https://github.com/go-logr/logr, libraries/SDKs should log to interfaces IMHO, so that aligns with my principles, but that'll be a bigger change. Is the goal to inject logger instances rather than reaching for global variables? Or do you see benefits in the structured logging or numbered levels?

jonjohnsonjr:
> What regrets do you have with the ggcr approach? I like the simplicity of it, which unsurprisingly aligns with ko 😄.

The fact that's it's package-level globals is kinda gross. I'm not sure if adding a logger option is much better, but I do hate globals :)

> but that'll be a bigger change

Agreed... I'm assuming you'd want to expose this for use by skaffold? Otherwise we could keep it internal.

>  Is the goal to inject logger instances rather than reaching for global variables?

Yup!

PR #569, opened 2022-01-25
Enable log levels and injecting loggers, using `go-logr`

halvards:
This change adds a dependency on `go-logr/logr` and replaces all* uses of the `log` package from Go's standard library with a `logr`-based implementation.

To ease the transition, and to keep the diff small, this change introduces a package `github.com/google/ko/pkg/log` that should be imported in place of `log` from the standard library.

The new package provides some of the same functions as `log`, e.g., `Printf()` and `Fatalf()`. The replacement functions takes an extra `context.Context` argument. The `Context` is used to look up the `logr.Logger` to use for logging.

Existing log statements have been updated to use the transition functions, with `Context` plumbed through where required.

New code can use the new logging syntax, e.g.:

```go
log.L(ctx).Info("something interesting happened", "key", "value")
```

or

```go
log.L(ctx).Error(err, "uh oh", "what", "snafu")
```

This change alters the log output to `logr`'s structured output, with named fields for the log message and level. As an example, a log line that previously looked like this:

```
2022/01/25 19:44:18 Building github.com/google/ko/test for linux/amd64
```

Will look like as follows with this change:

```
2022/01/25 19:46:02 ko: "level"=1 "msg"="Building github.com/google/ko/test for linux/amd64"
```

\* The `log` package from Go's standard library is still used in the deprecated `pkg/commands/options/filestuff.go`.

Fixes: #560
Related: #542

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,8 @@ require (
 	github.com/docker/docker v20.10.12+incompatible
 	github.com/dprotaso/go-yit v0.0.0-20191028211022-135eb7262960
 	github.com/fsnotify/fsnotify v1.5.1
+	github.com/go-logr/logr v1.2.2
+	github.com/go-logr/stdr v1.2.2
 	github.com/go-training/helloworld v0.0.0-20200225145412-ba5f4379d78b
 	github.com/google/go-cmp v0.5.7
 	github.com/google/go-containerregistry v0.8.0
--- a/go.sum
+++ b/go.sum
@@ -602,8 +602,11 @@ github.com/go-logr/logr v0.1.0/go.mod h1:ixOQHD9gLJUVQQ2ZOR7zLEifBX6tGkNJF4QyIY7
 github.com/go-logr/logr v0.2.0/go.mod h1:z6/tIYblkpsD+a4lm/fGIIU9mZ+XfAiaFtq7xTgseGU=
 github.com/go-logr/logr v0.4.0/go.mod h1:z6/tIYblkpsD+a4lm/fGIIU9mZ+XfAiaFtq7xTgseGU=
 github.com/go-logr/logr v1.0.0/go.mod h1:jdQByPbusPIv2/zmleS9BjJVeZ6kBagPoEUsqbVz/1A=
-github.com/go-logr/logr v1.2.0 h1:QK40JKJyMdUDz+h+xvCsru/bJhvG0UxvePV0ufL/AcE=
 github.com/go-logr/logr v1.2.0/go.mod h1:jdQByPbusPIv2/zmleS9BjJVeZ6kBagPoEUsqbVz/1A=
+github.com/go-logr/logr v1.2.2 h1:ahHml/yUpnlb96Rp8HCvtYVPY8ZYpxq3g7UYchIYwbs=
+github.com/go-logr/logr v1.2.2/go.mod h1:jdQByPbusPIv2/zmleS9BjJVeZ6kBagPoEUsqbVz/1A=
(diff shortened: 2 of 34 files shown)


halvards:
The `e2e windows-latest` tests fail because the image index for `golang:1.17` doesn't contain a Windows platform with `os.version` matching what the action step expects. The action step hardcodes `10.0.17763.2366`, but currently `golang:1.17` only contains `10.0.17763.2452`.

We could bump the version number in the step to match, but that's just a temporary workaround. Any suggestions?

halvards:
If I read the step right, the job should actually pick up `mcr.microsoft.com/windows/nanoserver:1809` rather than `golang:1.17`, using `KO_DEFAULTBASEIMAGE`. Based on the logs, that doesn't appear to be happening though. And in any case, `os.version` in the Windows base image also doesn't match `OSVERSION` from the step.

jonjohnsonjr:
Thanks for doing this!

> The Context is used to look up the logr.Logger to use for logging.

I've done this a couple times and I've grown to really dislike this pattern. It's somewhat convenient, but IMO it's too hard to discover  via godoc or the type system and too easy to get wrong (resulting in just no logs because we get a discard logger).

I'd prefer having us "manually" inject a logger wherever we need it, even though it's kind of annoying to plumb around. Happy to discuss it though, if folks disagree.

halvards:
> #572

Thanks for the pointer. I think the other problem is that `baseImageOverrides` in `.ko.yaml` appears to takes precedence over `KO_DEFAULTBASEIMAGE`, which is not what most users would expect. That causes the failing step to use `golang` as the base image instead of `nanoserver`. I'll verify and create an issue for this.

halvards:
> > The Context is used to look up the logr.Logger to use for logging.
> 
> I've done this a couple times and I've grown to really dislike this pattern. It's somewhat convenient, but IMO it's too hard to discover via godoc or the type system and too easy to get wrong (resulting in just no logs because we get a discard logger).
> 
> I'd prefer having us "manually" inject a logger wherever we need it, even though it's kind of annoying to plumb around. Happy to discuss it though, if folks disagree.

I completely agree that discoverability suffers with this approach, and I'd be completely fine to abandon this PR if that's the priority 👍.

Two arguments in favor of this approach:

1. Plumbing through a logr.Logger significantly alters the effort-to-benefit ratio of this feature.

2. If we want to plumb through other arguments, e.g., spans for tracing, then our function arguments may start to get out of hand.

Two alternative tweaks to the context-wrapped approach that won't do much to aid discoverability, but should reduce the chances of getting it wrong:

1. Fall back to a basic [`funcr`](https://github.com/go-logr/logr/blob/master/funcr/funcr.go)-based logger that just does `fmt.Fprintln(os.Stderr, prefix, args)`, instead of the discard logger. Perhaps also with a warning message about using a non-performant fallback logger. Reason for using `funcr` rather than `stdr` is to avoid another dependency.

2. Fail if there's no logger in the context or if context is nil, using `panic()`!? That way, an accidentally re-initialized (or nil) context should be caught early. This will require some more work to ensure that unit tests all use contexts with a logger.

jonjohnsonjr:
> If we want to plumb through other arguments, e.g., spans for tracing, then our function arguments may start to get out of hand.

I think trace spans actually _are_ a good thing to pass along in a context, but the logger feels a lot more like application configuration than request-scoped data.

I do see that logr has these context facilities built in, so maybe I should just give this fight up :P but https://dave.cheney.net/2017/01/26/context-is-for-cancelation is so compelling.

halvards:
Is anyone interested in this? If not, I'm inclined to close it as I don't want to add tech debt (context misuse) for a niche requirement.

halvards:
Thanks for the feedback! Closing it now, we can revisit in the future if necessary.

Issue #412, opened 2021-08-09
RFC: deprecate and remove Watch Mode


imjasonh:
Watch Mode (i.e, `ko apply --watch`) was added about 2 1/2 years ago, in the old ggcr repo, in https://github.com/google/go-containerregistry/pull/336.

We don't collect usage metrics, but based on recommended/documented usage in projects that I know of that use `ko`, none use watch mode today. It was marked as EXPERIMENTAL from the beginning, and documentation for it was apparently accidentally removed from the README six months ago in https://github.com/google/ko/pull/318 😞.

Watch Mode adds quite a bit of complexity to what is otherwise a relatively straightforward transformation on inputs. It involves implementing non-standard concurrency primitives like [`Future`s](https://github.com/google/ko/blob/9f6e0d305b3d3512e3c11533366ba7057da298a9/pkg/build/future.go). It's not well tested, and not currently easily test-_able_. It might make future improvements to outputs like https://github.com/google/ko/issues/406 and [`ko extract`](https://github.com/google/ko/issues/358#issuecomment-834599338) more difficult, or simply incompatible with watch mode.

With support for watching and re-building images with `ko` available through Skaffold ([today unofficially](https://skaffold.dev/docs/tutorials/custom-builder/); [in the future, officially](https://github.com/GoogleContainerTools/skaffold/issues/6041)), and upcoming improvements to speed up no-op builds (#267 #269) I think we should consider deprecating and eventually removing watch mode from `ko`, simplifying the codebase and focusing efforts on better release management integration and security best practices.

If we decide to proceed, we can start by adding a warning linking this issue when using `--watch`, to collect any feedback that might encourage us to reconsider deprecating. If we do decide to keep it, we should invest more in documentation and testing, and promote it out of an undocumented experiment.

wdyt @mattmoor @jonjohnsonjr

jonjohnsonjr:
> If we decide to proceed, we can start by adding a warning linking this issue when using --watch, to collect any feedback that might encourage us to reconsider deprecating.

This seems funny to me. If `--watch` is set, print something like `Come argue in issue 412 if you use this.`

On the other hand, I _think_ I like the future stuff. It is pretty nice for streaming support (`ko resolve -f -`).

imjasonh:
Honestly I'm fine with just removing it, I just want to give people an opportunity to tell us why they think it's load-bearing for them, if it is.

[...]

PR #571, opened 2022-01-25
feat: allow symlink

### LINKED ISSUE #460: `ko apply --recursive -f` to follow symlinked directories
This will enable sharing *.yaml files between different deployment configurations, or with other tools (like a helm chart directory).

developer-guy:

Fixes #460 

Is that what you thought? @imjasonh

### DIFF
--- a/pkg/commands/options/filestuff.go
+++ b/pkg/commands/options/filestuff.go
@@ -59,12 +59,21 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
 				// processing (unless we are in recursive mode).  If we decide to process
 				// the directory, and we're in watch mode, then we set up a watch on the
 				// directory.
+				symlink, err := ResolveSymlink(path, fi)
+				if err != nil {
+					return err
+				}
+
 				if fi.IsDir() {
 					if path != paths && !fo.Recursive {
 						return filepath.SkipDir
 					}
 					// We don't stream back directories, we just decide to skip them, or not.
 					return nil
+				}
+
+				if symlink != "" {
+					files <- symlink
 				}
 
 				// Don't check extension if the filepath was passed explicitly
@@ -86,4 +95,14 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
 		}
 	}()
 	return files
+}
+
+func ResolveSymlink(path string, fi os.FileInfo) (symlink string, err error) {
+	if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
+		symlink, err = os.Readlink(path) // TODO: Handle error here
+		if !filepath.IsAbs(symlink) {    // Output of os.Readlink is OS-dependent...
+			symlink = filepath.Join(filepath.Dir(path), symlink)
+		}
+	}
+	return
 }
--- a/test/symlink.yaml
+++ b/test/symlink.yaml
@@ -1,0 +1,1 @@
+test.yaml
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -405,7 +405,25 @@ func resolveFile(
 	if f == "-" {
 		b, err = ioutil.ReadAll(os.Stdin)
 	} else {
-		b, err = ioutil.ReadFile(f)
+		fi, err := os.Lstat(f)
+		if err != nil {
+			return nil, fmt.Errorf("unable to find file %s selector: %wß", f, err)
+		}
+
+		symlink, err := options.ResolveSymlink(f, fi)
+		if err != nil {
+			return nil, fmt.Errorf("unable to resolve symlink of file %s selector: %w", f, err)
+		}
+
+		if symlink != "" {
+			b, err = ioutil.ReadFile(symlink)
+		} else {
+			b, err = ioutil.ReadFile(f)
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("unable to read file %s: %w", f, err)
+		}
 	}
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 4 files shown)


pkg/commands/options/filestuff.go
```diff
@@ -87,11 +87,20 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
...
 						return filepath.SkipDir
 					}
 					if watcher != nil {
-						watcher.Add(path)
+						if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
+							symlink, err := filepath.EvalSymlinks(path)
+							if err != nil {
+								return err
+							}
+							watcher.Add(symlink)
```
imjasonh, line 100: This seems to only affect watch mode (`-W`), which is deprecated and will eventually be removed (#412).

Does this also happen to work for non-watch-mode `ko apply -f`?

developer-guy: I did something to fix that, hope it'd work 🤦🏻‍♂️

pkg/commands/options/filestuff.go
```diff
@@ -87,11 +87,17 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
 				// the directory, and we're in watch mode, then we set up a watch on the
 				// directory.
 				if fi.IsDir() {
+					symlink, err := resolveSymlink(path, fi)
```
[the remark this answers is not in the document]
developer-guy: I've tested all of the samples and everything went well 🙋🏻‍♂️

would you mind reviewing one more time ✌️

pkg/commands/options/filestuff.go
```diff
@@ -142,3 +155,10 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
 	}()
 	return files
 }
+
+func ResolveSymlink(path string, fi os.FileInfo) (symlink string, err error) {
+	if fi.Mode()&os.ModeSymlink == os.ModeSymlink {
+		symlink, err = os.Readlink(path) // TODO: Handle error here
```
imjasonh, line 161: This TODO is done now right?

imjasonh:
I think the suggestion in https://github.com/google/ko/pull/571#discussion_r792132067 is still useful, we could add a symlink in the repo and exercise this new path in an e2e test

pkg/commands/options/filestuff.go
developer-guy: yes, right.

pkg/commands/options/filestuff.go
```diff
@@ -86,17 +86,30 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
...
+					return err
+				}
+
 				if fi.IsDir() {
 					if path != paths && !fo.Recursive {
 						return filepath.SkipDir
 					}
 					if watcher != nil {
-						watcher.Add(path)
+						watcher.Add(symlink)
```
imjasonh, line 99: Let's just remove any watcher interaction, we're going to delete it all soon anyway in #585

pkg/commands/options/filestuff.go
```diff
@@ -86,17 +86,30 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
...
+					} else {
+						if symlink != "" {
+							files <- symlink
+						}
```
imjasonh, line 103:
```suggestion
					} else if symlink != "" {
						files <- symlink
					}
```

pkg/commands/options/filestuff.go
```diff
@@ -142,3 +155,10 @@ func EnumerateFiles(fo *FilenameOptions) chan string {
 	}()
 	return files
 }
+
+func ResolveSymlink(path string, fi os.FileInfo) (symlink string, err error) {
```
imjasonh, line 159: Can this be replaced with https://pkg.go.dev/path/filepath#EvalSymlinks ?

test/symlink.yaml
```diff
@@ -0,0 +1 @@
+test/test.yaml
```
imjasonh, line 1: Can you add a test that checks that this symlink is followed if you `ko apply -f test/symlink.yaml` ?

pkg/commands/options/filestuff.go
developer-guy: thank you 🙋🏻‍♂️

pkg/commands/options/filestuff.go
developer-guy: OFC, thank you 🙋🏻‍♂️

pkg/commands/options/filestuff.go
developer-guy: it didn't work when I switched to EvalSymlink, I don't know why 🤷🏻‍♂️

test/symlink.yaml
developer-guy: I tried something to test this but couldn't make it work but anyways I wanted to share the code with you to see what I missed or did wrong 🙋🏻‍♂️ maybe you can help me to fix this 🙏

developer-guy: thanks a ton, @dentrax, he helped me to fix this. ❤️

Dentrax:
NIT: One small concern here is that we do not support the case in which symlink points itself: `too many levels of symbolic links`

A workaround: https://github.com/fkautz/gitbom-go/pull/7/files (It would be overengineered to implement that in this project I think)

pkg/commands/options/filestuff.go
imjasonh: I'd like to understand better why that's the case. This is getting fairly complicated, and I'd rather depend on stdlib methods as much as possible. If we still need to write significant code to handle symlinks, I'm a bit more inclined not to support symlinks for now.

developer-guy:
kindly ping 🙋🏻‍♂️

PR #578, opened 2022-02-01
Set GOAMD64 if variant is set

imjasonh:
This change also changes behavior of GOARM, since docs suggest that it's only used when the `GOARCH=arm`, and not `arm64`.

This change should benefit users using the upcoming Go 1.18 building on base images that provide amd64 with a variant specified.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -335,14 +335,20 @@ func buildEnv(platform v1.Platform, userEnv, configEnv []string) ([]string, erro
 		"GOOS=" + platform.OS,
 		"GOARCH=" + platform.Architecture,
 	}
-
-	if strings.HasPrefix(platform.Architecture, "arm") && platform.Variant != "" {
-		goarm, err := getGoarm(platform)
-		if err != nil {
-			return nil, fmt.Errorf("goarm failure: %w", err)
-		}
-		if goarm != "" {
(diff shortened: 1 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -335,14 +335,20 @@ func buildEnv(platform v1.Platform, userEnv, configEnv []string) ([]string, erro
...
-	if strings.HasPrefix(platform.Architecture, "arm") && platform.Variant != "" {
-		goarm, err := getGoarm(platform)
-		if err != nil {
-			return nil, fmt.Errorf("goarm failure: %w", err)
-		}
-		if goarm != "" {
-			env = append(env, "GOARM="+goarm)
+	if platform.Variant != "" {
+		switch platform.Architecture {
+		case "arm":
```
jonjohnsonjr, line 340: This drops arm64.

imjasonh: This is intentional.

> This change also changes behavior of GOARM, since docs suggest that it's only used when the GOARCH=arm, and not arm64.

The docs don't seem to mention GOARM's behavior when GOARCH=arm64, only when GOARCH=arm. Maybe I'm misreading?

jonjohnsonjr: Well there it is.

jonjohnsonjr (approved):

PR #581, opened 2022-02-03
Include cred helpers in keychain

imjasonh:
This adds implicit support for Google, Amazon, Azure and GitHub container registries if the environment provides credentials.

Binary size increases from 22 MB -> 26 MB


imjasonh:
lol: https://github.com/google/ko/runs/5045688219?check_suite_focus=true

Pulling distroless attempts to use gcloud auth, which is not configured, so it fails closed, instead of falling back to trying anonymous, which would succeed. 🤔

imjasonh:
This should be RFAL now.

mattmoor (approved):
🤩

PR #583, opened 2022-02-06
Introduce koverrides

jdolitsky:
Introduce the concept of koverrides, which provide a lightweight interface for supplying environment-specific configuration settings in Kubernetes YAML files(s).

Similar to the `ko://` prefix, YAML nodes containing a `koverride://` prefix will be preprocessed. Users can then use the `--koverrides` flag to provide configuration overrides (if necessary).

See README updates in this changeset for more details. 

*(copy pasting README changes below)*

## "koverrides"

When using `ko` in the context of Kubernetes, you may wish to reuse your YAML file(s)
across several different environments.

As a lightweight solution for environment-specific configuration, you can leverage
"koverrides" (a.k.a. `ko` overrides).

Using koverrides, you can prefix values with `koverride://` and supply a default value:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo
data:
  db-host: koverride://db.host/mysql.default.svc
  db-port: koverride://db.port/3306
  db-user: koverride://db.user
  db-pass: koverride://db.pass
```

The default value can be defined as everything after the first forward slash (`/`)
character. If there is no default, then an empty string will be used (`""`).
Nested keys are supported using dot (`.`) notation.

Consider the following koverrides file, `demo.yaml`:

```yaml
db:
  host: db.example.com
  user: secretuser
  pass: secretpass
```

Now we run `ko resolve` (or `ko apply`) using the
`--koverrides` flag, pointing to the file:

```
ko resolve -f config/ --koverrides demo.yaml
```

Which produces the following YAML:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: demo
data:
  db-host: db.example.com
  db-port: "3306"
  db-user: secretuser
  db-pass: secretpass
```

For practical purposes, all resulting values are coerced to strings
(e.g. `3306 -> "3306"`).

This feature is particularly useful for Kubernetes resource types designed
for configuration purposes, such as ConfigMaps or Secrets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Ignore GoLand (IntelliJ) files.
 .idea/
 
+coverage.txt
 ko
--- a/README.md
+++ b/README.md
@@ -421,6 +421,68 @@ ko delete -f config/
 This is purely a convenient alias for `kubectl delete`, and doesn't perform any
 builds, or delete any previously built images.
 
+## "koverrides"
+
+When using `ko` in the context of Kubernetes, you may wish to reuse your YAML file(s)
+across several different environments.
+
+As a lightweight solution for environment-specific configuration, you can leverage
+"koverrides" (a.k.a. `ko` overrides).
+
+Using koverrides, you can prefix values with `koverride://` and supply a default value:
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: demo
+data:
+  db-host: koverride://db.host/mysql.default.svc
+  db-port: koverride://db.port/3306
+  db-user: koverride://db.user
+  db-pass: koverride://db.pass
+```
+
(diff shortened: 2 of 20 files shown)


imjasonh:
Interesting!

Is there a way a user could do this with some existing YAML templating tool like `kustomize` today, that `ko`verrides supports better somehow? In general I'd prefer to have `ko` remain a relatively single-use tool, rather than embarking down the path of more complex YAML templating/overrides, especially while existing tools exist for that.

jonjohnsonjr:
Yeah I'd punt this to `kustomize` or [`skaffold`](https://skaffold.dev/docs/pipeline-stages/builders/ko/).

jdolitsky:
Perhaps you didn't notice, but I included unit tests for this feature.

PR #573, opened 2022-01-27
Generate CycloneDX SBOMs

imjasonh:
The heavy lifting is done by cyclonedx libraries recently exported by @nscuro 👏 👏👏

However... I'd love to find a way to get this same functionality without bringing in ~1M new files, which increases the total binary size by 50% (22 MB -> 33 MB).

I'll take a dive through cyclone's dependency graph and see if we can find anything to trim.

---

But, it works:

```
$ cosign download sbom $(go run ./ build ./ --sbom=cyclonedx)
2022/01/27 17:00:00 Using base golang:1.17 for github.com/google/ko
2022/01/27 17:00:01 Building github.com/google/ko for linux/amd64
2022/01/27 17:00:08 Publishing gcr.io/imjasonh/ko-98b8c7facdad74510a7cae0cd368eb4e:latest
...
2022/01/27 17:00:10 Published SBOM gcr.io/imjasonh/ko-98b8c7facdad74510a7cae0cd368eb4e:sha256-cf4e68ec7e4801d88dc1ba0092a9d20966e354c9eae337ef1f5f961dabb87a83.sbom
...
Found SBOM of media type: application/vnd.cyclonedx
{
  "bomFormat": "CycloneDX",
  "specVersion": "1.3",
  "version": 1,
  "metadata": {
...
```

Example output: https://gist.github.com/imjasonh/31002f904a28166c9c8c9b6cedd07d96

---

TODO:
- [ ] add this to `ko deps`

PR #587, opened 2022-02-09
Generate CycloneDX SBOMs using our own JSON generation

imjasonh:
Example output: https://gist.github.com/imjasonh/16185e53200bed16470e5d27e17f261e 👀 

```
cosign download sbom $(go run ./ build ./ --sbom=cyclonedx)                                                                                                                                  
```

### DIFF
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -1,0 +1,98 @@
+name: Validate SBOMs
+
+on:
+  pull_request: 
+    branches: ['main']
+
+jobs:
+  go-version-m:
+    name: Generate go version -m
+    runs-on: ubuntu-latest
+
+    env:
+      KO_DOCKER_REPO: localhost:1338
+
+    steps:
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.17.x'
+      - name: Install cmd/registry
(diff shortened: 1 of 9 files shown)


imjasonh:
Compare this output to that generated by the cyclonedx-gomod library proposed in #573: https://gist.github.com/imjasonh/31002f904a28166c9c8c9b6cedd07d96

mattmoor:
cc @puerco for thoughts

puerco:
I think it makes sense to write the small generator for both formats. Since the SBOM is only listing the dependencies for now, pulling in the full spdx/cycloneDX libraries seems like overkill. 

@imjasonh how much further do you see ko acquiring more sbom functions?  I can think of three main areas which would be good to have, but maybe are out of scope for ko to do these:

1. Licensing information
2. Describing the image(s) structure 
3. System packages from the base images

puerco:
> This would let us report this SBOM only for the ko layer(s), and depend on base images to report their own SBOMs, which we can detect and carry forward.

Awesome, I am currently working on that part. An important feature will be to be able to reference external documents from the deps SBOM generated by ko, I can add that feature for the next release.

jonjohnsonjr (approved):

PR #592, opened 2022-02-11
Log base image@digest when building

imjasonh:
Before:

```
2022/02/11 09:15:05 Using base golang:1.17 for github.com/google/ko
```

After:

```
2022/02/11 09:13:49 Using base golang:1.17 (index.docker.io/library/golang@sha256:1a35cc2c5338409227c7293add327ebe42e1ee5465049f6c57c829588e3f8a39) for github.com/google/ko
```

I couldn't tell where else to find the specific base image that was used before, without trying to guess using layer digests.

### DIFF
--- a/pkg/commands/config.go
+++ b/pkg/commands/config.go
@@ -106,11 +106,16 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 			return ref, cached, nil
 		}
 
-		log.Printf("Using base %s for %s", ref, s)
 		result, err := fetch(ctx, ref)
 		if err != nil {
 			return ref, result, err
 		}
+		dig, err := result.Digest()
+		if err != nil {
+			return ref, result, err
+		}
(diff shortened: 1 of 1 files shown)


imjasonh:
This could be shorter if the base image is _specified_ by digest in .ko.yaml, I'm not sure how common that is in practice though.

mattmoor:
That's pretty long, I wonder if we should just emit the hash part, since the rest is redundant with something we're already printing on that line.

imjasonh:
> That's pretty long, I wonder if we should just emit the hash part, since the rest is redundant with something we're already printing on that line.

Yeah, good call. Now it's just:

```
2022/02/11 09:53:26 Using base golang:1.17@sha256:1a35cc2c5338409227c7293add327ebe42e1ee5465049f6c57c829588e3f8a39 for github.com/google/ko
```

mattmoor (approved):

PR #595, opened 2022-02-13
Sign image using cosign

imjasonh:
Annotate signature with run information

edit: example successful run: https://github.com/imjasonh/ko/runs/5174710473?check_suite_focus=true


mattmoor (approved):
🤩

PR #596, opened 2022-02-16
Use configured image repository when publishing to Docker daemon

aknuds1:
Use configured image repository when publishing to Docker daemon, instead of hardcoding to ko.local.


imjasonh:
Hey @aknuds1 ! I'm not sure if you intended to send this, but just in case this is something you wanted to contribute, I think it could be useful.

I looked around in your fork and it has some work to support a `ko publish-layout` feature. I didn't look very closely, but it sounds like this would be pretty similar to the existing `ko publish --oci-layout-path` option.

There might also be some overlap with [your fork's `ko push <layout>` work](https://github.com/grafana/ko/commit/6585bd206ee6596b385b03ab1359e2d92f2fc95b) and [`crane push <layout>`](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane_push.md). All that is to say, all of this is work we're interested in supporting in general, and we've (mostly @jonjohnsonjr) has already done some work to enable it. If there's anything missing, please let us know, we'd love to help!

aknuds1:
@imjasonh thanks for the feedback :) I will look at `ko publish --oci-layout` when time permits, and get back to you. We do mean to try to upstream any changes.

aknuds1:
@imjasonh `ko publish` is just an alias for `ko build` isn't it? And `ko build --oci-layout-path` will only store an OCI image layout to disk won't it? The purpose of `ko publish-layout` is to instead publish an OCI image layout that has been previously built (so to break into two different CI steps).

It sounds as if `ko publish-layout` indeed overlaps with `crane push`. I'm wondering though if it would be beneficial to have this support in ko and not having to throw crane into the mix.

imjasonh:
Yep, `ko publish` and `ko build` are equivalent (technically `ko build` hasn't been released yet -- soon!)

`ko build --oci-layout-path=layout` will build an image _and_ write its layout to the path. You can disable the push part of that with `--push=false`.

The intention was that after that you could do whatever you want with the layout (scan it, modify it, etc.), then push it to a registry with `crane push`, or any other similar tool that can push a layout.

Off-topic, but there's also been some interest in making [go-containerregistry's registry implementation in `pkg/registry`](https://github.com/google/go-containerregistry/tree/main/pkg/registry) capable of serving contents from a layout directory. So you could (eventually) `ko publish --oci-layout-path=layout/ --push=false` then `./registry --oci-layout-path=layout/` and serve the built contents directly locally, for testing, etc. If that's interesting to you, let me know, we can put more work into that.

---

> It sounds as if ko publish-layout indeed overlaps with crane push. I'm wondering though if it would be beneficial to have this support in ko and not having to throw crane into the mix.

I've been hesitant to add arbitrary image management features to `ko` directly, since it's ideally (to me) a relatively single-use tool to build images from Go code. There's also some light YAML templating, but mostly it just builds images.

`crane` is also a (relatively) single-use tool for moving images around, between registries, via image tarballs and layouts, etc. There are other tools that can do it too, Skopeo, some others, and there could be even more in the future.

This isn't to say that we absolutely won't add things like this to `ko`, but that I'd want to see a strong signal that this helps with `ko`'s usability, and doesn't proliferate identical functionality across a bunch of similar tools (inevitably, with slight incompatibilities slipped in).

aknuds1:
> I've been hesitant to add arbitrary image management features to ko directly, since it's ideally (to me) a relatively single-use tool to build images from Go code. There's also some light YAML templating, but mostly it just builds images.
>
> crane is also a (relatively) single-use tool for moving images around, between registries, via image tarballs and layouts, etc. There are other tools that can do it too, Skopeo, some others, and there could be even more in the future.
>
> This isn't to say that we absolutely won't add things like this to ko, but that I'd want to see a strong signal that this helps with ko's usability, and doesn't proliferate identical functionality across a bunch of similar tools (inevitably, with slight incompatibilities slipped in).

@imjasonh OK that makes sense, thanks. I actually remember now that we also use `ko publish-layout` to load OCI image layouts into the Docker daemon. We cannot do that with crane, right? We'd have to eventually use `registry` to serve the layout?

imjasonh:
> @imjasonh OK that makes sense, thanks. I actually remember now that we also use `ko publish-layout` to load OCI image layouts into the Docker daemon. We cannot do that with crane, right? We'd have to eventually use `registry` to serve the layout?

Ah, yeah. `crane` doesn't do daemon operations only registry<->registry and registry<->tarball/layout.

As you say, one option would be to do layout -> layout-backed-registry -> `docker pull` -> daemon, but that's getting pretty Rube-Goldergian. Maybe there's an opportunity for a dedicated layout<->daemon tool. That should be fairly straightforward to build with go-containerregistry I think. I'd be happy to contribute/advise on something like that.

In the meantime, your thought about making `ko` be able to load images into the daemon that aren't `ko.local` sounds like a good step too, cutting out the layout steps.

aknuds1:
> Ah, yeah. `crane` doesn't do daemon operations only registry<->registry and registry<->tarball/layout.
> 
> As you say, one option would be to do layout -> layout-backed-registry -> `docker pull` -> daemon, but that's getting pretty Rube-Goldergian. Maybe there's an opportunity for a dedicated layout<->daemon tool. That should be fairly straightforward to build with go-containerregistry I think. I'd be happy to contribute/advise on something like that.

@imjasonh I am thinking now of making a dedicated little tool to implement pushing OCI image layouts either to a registry or to the daemon. I.e., instead of forking ko.

> In the meantime, your thought about making `ko` be able to load images into the daemon that aren't `ko.local` sounds like a good step too, cutting out the layout steps.

Thanks. That is useful for us, since we don't have to shoehorn in the support for ko.local prefixed images into our logic.

imjasonh:
> @imjasonh I am thinking now of making a dedicated little tool to implement pushing OCI image layouts either to a registry or to the daemon. I.e., instead of forking ko.

Sounds great! Let me know if you have any trouble, I'd be happy to help. If you find anything missing in either layout or daemon, [file a bug at go-containerregistry](https://github.com/google/go-containerregistry/issues/new/choose) and it'll find me.

PR #601, opened 2022-02-17
Update 'ko publish' -> 'ko build' in README

imjasonh:
Also tweak some wording about "publishing" images locally, since that doesn't make much sense.


mattmoor (approved):
🎉

PR #604, opened 2022-02-20
feat: not enforce KO_DOCKER_REPO with push=false

### LINKED ISSUE #590: ux: consider do not enforce `KO_DOCKER_REPO` variable if `--push=false`
Since we can pass `--push=false`, in this case `$KO_DOCKER_REPO` env flag could be useless to check whether it's set or not. It should be only checked if we're pushing image to remote. Wdyt?

```
$ ko publish --bare -B --platform=linux/amd64 --tags latest --push=false --oci-layout-path oci-dir .
Error: error creating publisher: KO_DOCKER_REPO environment variable is unset
2022/02/10 10:01:55 error during command execution:error creating publisher: KO_DOCKER_REPO environment variable is unset
```

dirien:
Hi everyone,

This PR fixes #590 and was good for me to get used with the codebase of ko.

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -186,7 +186,7 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 			return publish.NewKindPublisher(namer, po.Tags), nil
 		}
 
-		if repoName == "" {
+		if repoName == "" && po.Push {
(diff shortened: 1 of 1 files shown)


imjasonh:
Nice! Thanks for this contribution!

imjasonh (approved):

PR #607, opened 2022-02-21
fix: create dir for KOCACHE

### LINKED ISSUE #591: KOCACHE directory is not created if set
```
$ echo $KOCACHE
/Users/jasonhall/.ko
$ ls $KOCACHE
ls: /Users/jasonhall/.ko: No such file or directory
$ go run ./ build ./test/
... succeeds, takes a few seconds
$ go run ./ build ./test/
... succeeds, takes about the same amount of time
$ ls $KOCACHE
ls: /Users/jasonhall/.ko: No such file or directory
```

At this point, I would have expected for the `KOCACHE` directory to be created if I specified it, or at least complain in some way that it was specified but doesn't exist.

If I create it, I also need to `export` the env var to actually see the cache directory get populated.

developer-guy:

Fixes #591 

cc: @imjasonh

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -261,6 +261,15 @@ func build(ctx context.Context, ip string, dir string, platform v1.Platform, con
 	}
 
 	if dir := os.Getenv("KOCACHE"); dir != "" {
+		dirInfo, err := os.Stat(dir)
+		if os.IsNotExist(err) {
+			if err := os.MkdirAll(dir, os.ModePerm); err != nil && !os.IsExist(err) {
+				return "", fmt.Errorf("could not create KOCACHE dir %s: %w", dir, err)
+			}
+		} else if !dirInfo.IsDir() {
+			return "", fmt.Errorf("KOCACHE should be a directory, %s is not a directory", dir)
+		}
+
 		// TODO(#264): if KOCACHE is unset, default to filepath.Join(os.TempDir(), "ko").
 		tmpDir = filepath.Join(dir, "bin", ip, platform.String())
 		if err := os.MkdirAll(tmpDir, os.ModePerm); err != nil {
--- a/pkg/build/gobuild_test.go
+++ b/pkg/build/gobuild_test.go
@@ -23,9 +23,11 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"os"
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -255,12 +255,17 @@ func build(ctx context.Context, ip string, dir string, platform v1.Platform, con
 		return "", fmt.Errorf("could not create env for %s: %w", ip, err)
 	}
 
-	tmpDir, err := ioutil.TempDir("", "ko")
+	tmpDir, err := ioutil.TempDir(os.TempDir(), "ko")
```
imjasonh, line 258: I'm not sure this change is necessary, at least not for the goal of this PR.

pkg/build/gobuild.go
```diff
@@ -255,12 +255,17 @@ func build(ctx context.Context, ip string, dir string, platform v1.Platform, con
...
 	if err != nil {
 		return "", err
 	}
 
 	if dir := os.Getenv("KOCACHE"); dir != "" {
+		if _, err := os.Stat("dir"); os.IsNotExist(err) {
```
imjasonh, line 264: This should be `dir`, not the string `"dir"`.

imjasonh, line 264: We should check if the file exists and is a directory. If it's a file etc then KOCACHE won't work.

imjasonh (dismissed):

imjasonh:
We should think of some way to effectively test this behavior end to end. Maybe a test that sets `KOCACHE=/tmp/$(date +%s)` or something and checks that it exists after a build.

pkg/build/gobuild.go
developer-guy: thank you, good catch 🙈

pkg/build/gobuild.go
developer-guy: okay, I reverted that change.

pkg/build/gobuild.go
developer-guy: makes sense, I've updated the code according to your update, thanks a ton. 🙋🏻‍♂️

pkg/build/gobuild_test.go
```diff
@@ -772,6 +774,42 @@ func TestGoBuild(t *testing.T) {
 	})
 }
 
+func TestGoBuildWithKOCACHE(t *testing.T) {
+	now := time.Now() // current local time
+	sec := now.Unix()
+	koCacheDir := "/tmp/"
```
jonjohnsonjr, line 780: I think `/tmp/` might not always be writable. I'd set this to https://pkg.go.dev/testing#T.TempDir and ensure that the `filepath.Join(t.TempDir(), strconv.FormatInt(sec, 10))` directory exists.

developer-guy: thank you, great advice, done. 🙋🏻‍♂️

jonjohnsonjr:
Can you rebase this to pick up the new actions config?

developer-guy:
> Can you rebase this to pick up the new actions config?

OFC 🙋🏻‍♂️

jonjohnsonjr (approved):

PR #609, opened 2022-02-22
Take advantage of Chainguard maintained versions of various actions.

mattmoor:
Also adds dependabot config to bump the action definitions via PRs.


imjasonh (approved):

mattmoor:
I bumped the cosign version used here as well, since that step hit a flake with 1.3.1

imjasonh (approved):

PR #618, opened 2022-02-25
warning users when using both --base-import-paths --bare flags

### LINKED ISSUE #542: Improve errors/warnings for unnecessary .ko.yaml settings

cpanato:
warning users when using both --base-import-paths --bare flags

Partially fixes #542 

@imjasonh let me know if this is something you have in mind, otherwise, if this is not the project want we can close or improve. thanks!

```
$ ./main build --base-import-paths  --bare .
2022/02/25 17:55:08 WARNING!
Both --base-import-paths and --bare was set.

--base-import-paths will take precedence and ignore --bare flag.

In future this will be an error
2022/02/25 17:55:10 Using base golang:1.17@sha256:e06c83493ef6d69c95018da90f2887bf337470db074d3c648b8b648d8e3c441e for github.com/google/ko
2022/02/25 17:55:10 Building github.com/google/ko for linux/amd64
^C2022/02/25 17:55:11 Unexpected error running "go build": signal: killed
Error: failed to publish images: error building "ko://github.com/google/ko": signal: killed
2022/02/25 17:55:11 error during command execution:failed to publish images: error building "ko://github.com/google/ko": signal: killed
```

### DIFF
--- a/pkg/commands/apply.go
+++ b/pkg/commands/apply.go
@@ -82,6 +82,10 @@ func addApply(topLevel *cobra.Command) {
   ko apply -f config -- --namespace=foo --kubeconfig=cfg.yaml
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := options.Validate(po, bo); err != nil {
+				return fmt.Errorf("validating options: %w", err)
+			}
+
 			if !isKubectlAvailable() {
 				return errors.New("error: kubectl is not available. kubectl must be installed to use ko apply")
 			}
--- a/pkg/commands/options/validate.go
+++ b/pkg/commands/options/validate.go
@@ -1,18 +1,16 @@
-/*
-Copyright 2022 Google LLC All Rights Reserved.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+// Copyright 2022 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 package options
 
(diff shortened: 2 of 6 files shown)


pkg/commands/apply.go
```diff
@@ -82,6 +82,11 @@ func addApply(topLevel *cobra.Command) {
   ko apply -f config -- --namespace=foo --kubeconfig=cfg.yaml
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			err := options.ValidateOptions(po, bo)
```
imjasonh, line 85: style nit: to avoid repetition, how about `options.Validate(po, bo)`?

imjasonh, line 85: style nit: if the err isn't used outside of the `if` block, I tend to prefer:

```
if err := options.Validate(...); err != nil {
  return fmt.Errorf(...)
}
```

This can make it clearer to readers that nothing else uses the `err`, since it won't even be defined outside the `if` block.

pkg/commands/options/validate.go
```diff
@@ -0,0 +1,40 @@
...
+
+import "log"
+
+const bareBaseFlagsWarning = `WARNING!
+-----------------------------------------------------------------
+Both --base-import-paths and --bare was set.
+
+--base-import-paths will take precedence and ignore --bare flag.
+
+In future this will be an error
```
imjasonh, line 27:
```suggestion
Both --base-import-paths and --bare were set.

--base-import-paths will take precedence and ignore --bare flag.

In a future release this will be an error.
```

pkg/commands/options/validate.go
```diff
@@ -0,0 +1,40 @@
...
+-----------------------------------------------------------------
+`
+
+func ValidateOptions(po *PublishOptions, bo *BuildOptions) error {
+	if po.Bare && po.BaseImportPaths {
+		log.Print(bareBaseFlagsWarning)
+
+		// TODO: return error when we decided to make this an error, for now it is a warning
+		return nil
```
imjasonh, line 36: I'd just remove this and fall through to `return nil` later.

In the future there might be more validations, and we should figure out how to warn on all of them that we can then return an error if any are errors, but that can happen in a future change.

imjasonh (commented):
Looks great to me, thanks for doing this! Just a few tiny style/naming things, but otherwise 👍

pkg/commands/options/validate.go
cpanato: thanks so much for a quick review and feedback!

cpanato:
thanks @imjasonh for the quick review and feedback, I've addressed your comments PTAL when have some free cycle

imjasonh (approved):

cpanato:
@imjasonh fixed the header ready to go

imjasonh (approved):

PR #623, opened 2022-02-28
Produce OCI images by default

imjasonh:
This changes build logic to prefer to produce OCI images and indexes,
even if original base images are Docker manifests or manifest lists.

OCI indexes support annotations, while Docker manifest lists do not, and
we'd like to inject base image information in annotations wherever possible.

Since Quay.io recently added support for OCI manifests, this is no
longer a serious breaking change -- and anyway, producing SBOMs by default
already breaks Quay.io without --sbom=none.

This behavior can be disabled with --preserve-docker-media-type=true,
which will result in Docker-type manifests being produced if and only if
the base image was a Docker-typed manifest.

This partially reverts commit 42723d75e7076c4946351c9e3197ce65ff31b4ec.

### DIFF
--- a/pkg/commands/options/build.go
+++ b/pkg/commands/options/build.go
@@ -69,7 +69,7 @@ type BuildOptions struct {
 	BuildConfigs map[string]build.Config
 
 	// If true, don't convert Docker-typed base images to OCI when building.
-	PreserveDockerMediaType bool
+	PreserveMediaType bool
 }
 
 func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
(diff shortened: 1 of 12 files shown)


pkg/commands/options/build.go
```diff
@@ -67,6 +67,9 @@ type BuildOptions struct {
 
 	// BuildConfigs stores the per-image build config from `.ko.yaml`.
 	BuildConfigs map[string]build.Config
+
+	// If true, don't convert Docker-typed base images to OCI when building.
+	PreserveDockerMediaType bool
```
jonjohnsonjr, line 72: Wdyt about `PreserveMediaTypes` instead? It's not necessarily docker media types being preserved.

imjasonh: Isn't it? The conversion only goes the one way, from docker media types to OCI types. Are there other types you're thinking of?

jonjohnsonjr: Not specifically but it's a shorter flag without `docker` and the code just preserves the existing media type rather than only docker media types.

imjasonh: Fair enough. Change incoming.

jonjohnsonjr (approved):

PR #630, opened 2022-03-02
check if have all and other platforms set in the --platform flag

cpanato:
- check if have all and other platforms set in the --platform flag

```
../../google/ko/main build --base-import-paths \
--platform=all,linux/arm --tags v1.5.1-96-gd1a4d08-dirty --tags d1a4d086309e22cc85c0c59ac662606a58c81438 github.com/sigstore/cosign/cmd/cosign
Error: validating options: all or specific platforms should be used
2022/03/02 14:12:16 error during command execution:validating options: all or specific platforms should be used
make: *** [ko] Error 1
```

Address https://github.com/google/ko/issues/542#issuecomment-997014021

/cc @imjasonh

### DIFF
--- a/pkg/commands/options/validate.go
+++ b/pkg/commands/options/validate.go
@@ -40,14 +40,6 @@ The --local flag might be deprecated in the future.
 -----------------------------------------------------------------
 `
 
-const platformsFlagWarning = `ERROR!
------------------------------------------------------------------------
-The --platform was set with all and other specific platforms.
-
-Please choose either all or a specific one, ie. --platform=linux/arm64
------------------------------------------------------------------------
-`
-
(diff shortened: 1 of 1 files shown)


pkg/commands/options/validate.go
```diff
@@ -26,11 +29,34 @@ In a future release this will be an error.
 -----------------------------------------------------------------
 `
 
+const platformsFlagWarning = `ERROR!
```
jonjohnsonjr, line 32: I think we can do without the warning if we're returning an error.

cpanato: done, thanks for the review and feedback! PTAL

imjasonh (approved):

cpanato:
rebased

PR #636, opened 2022-03-05
Correctly pass through base image annotations

imjasonh:
There were some bugs here before:

- for indexes, we'd annotate the base, but then append to empty.Index
  which didn't carry those forward.
- when producing single-platform images based on multi-platform indexes
  (the default and most common scenario), we wouldn't carry forward the
  original base index's annotations to the single matching platform base
  image.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -888,16 +888,15 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 		return nil, err
 	}
 
-	// Take the digest of the base index or image, to annotate images we'll build later.
-	baseDigest, err := base.Digest()
-	if err != nil {
-		return nil, err
-	}
-
 	// Annotate the base image we pass to the build function with
 	// annotations indicating the digest (and possibly tag) of the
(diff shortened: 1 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -888,16 +888,15 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
...
-	baseDigest, err := base.Digest()
-	if err != nil {
-		return nil, err
-	}
-
 	// Annotate the base image we pass to the build function with
 	// annotations indicating the digest (and possibly tag) of the
 	// base image.  This will be inherited by the image produced.
-	if mt != types.DockerManifestList && !g.preserveMediaType {
+	if !(mt == types.DockerManifestList && g.preserveMediaType) {
```
jonjohnsonjr, line 894: This will add annotations to docker images when preserveMediaType is true, do you want that?

imjasonh: Yeah. Docker images (not manifest lists) support annotations, at least as far as GCR lets me push those.

jonjohnsonjr (approved):

Issue #643, opened 2022-03-11
Idea: build all images defined in .ko.yaml


jsok:
## Idea

Running `ko build --all` or similar will build all `builds[]` defined in `.ko.yaml`.

## Motivation

Once you have more than a few builds defined, it becomes tedious to build everything. This is particularly true in a monorepo where you may have dozens of builds to publish in a CI pipeline.

imjasonh:
Ooh, nice. This would be useful for folks who aren't using the Kubernetes integration, to just build a whole repo's worth of images.

I wonder if we should just accept `ko build ./...` semantics, like `go build` does. This could scan the path for any `package main` and `ko build` it. This way users wouldn't have to define a `builds` section for each `package main` in their tree, which could get cumbersome if you have dozens of them.

imjasonh:
You can use `go list -json` to discover packages, and `jq` to filter those for packages that are `package main`.

```
$ go list -json ./... | jq -r -s '.[] | select (.Name == "main") | .ImportPath'
github.com/google/ko
github.com/google/ko/cmd/help
github.com/google/ko/test
```

So this should `ko build` every suitable package in your tree:

```
go list -json ./... | jq -r -s '.[] | select (.Name == "main") | .ImportPath' | xargs ko build
```

That should work, but it's not a great UX. We could package the ~same logic into `ko` itself to basically just do this when you `ko build ./...` (or `ko build ./cmd/...`, or whatever)

jsok:
ko build ./... is useful in some cases, however in my use case I really don't want to build and push everything, only those builds I've defined in .ko.yaml 😀

skaffold build follows a similar behaviour in that without any args it builds everything defined in skaffold.yaml.

carlpett:
Just found `ko`, and after trying it in one of our repos, would very much like a `--all` feature that builds everything from the configuration file. In my case, at least, it would be more useful than supporting ` ./...` (although the two do not really conflict). We tend to have a number of utility programs in our repos, so just blindly building everything with `package main` would be rather wasteful.

I'd be willing to give implementation a try if this would be accepted?

jsok:
@carlpett that's the same as my use case. I think it's a very valid argument, ./... is just too inflexible.

As mentioned above I just use skaffold instead since it has this capability and I need to build a few non-Go containers too.

imjasonh:
I think we should support `ko build --all`.

cernoel:
an alternative dumb (bash) approach might be: 
```
[...]

Issue #273, opened 2020-12-19
ko publish ./...


ImJasonH:
```
ko publish ./cmd/...
2020/12/19 09:39:01 failed to publish images: found 7 local packages, expected 1
```

`ko resolve` is  capable of building and deploying multiple images, and `ko publish` is seemingly capable of identifying multiple import paths to build, but we run afoul of https://github.com/google/ko/blob/722ad3c37b2e01219aa053ba11b6b0d6d065b5d7/pkg/commands/publisher.go#L39 in `ko publish`.

ImJasonH:
FWIW `ko publish ./cmd/foo ./cmd/bar ./cmd/baz` works, it's just `./...` that doesn't. Maybe this was an explicit decision, but I don't remember the motivation.

I guess `ko publish ./...` has slightly different semantics ("publish all supported refs in `./...`") compared to `ko publish ./a ./b` ("publish `./a` and `./b`, _if they are supported_").

jonjohnsonjr:
I don't think it was an explicit decision. Though... this reminds me of how we started requiring `ko://` in references because it made things a lot easier to reason about and error report, but I think it's probably fine to support this kind of thing, if the diff is reasonable. We should be able to find and build any main packages pretty cheaply... we might have to change some things from returning a `foo` to a `[]foo`, but I think that's fine.

PR #644, opened 2022-03-11
Support `ko build ./...`

imjasonh:
This changes `ko build` to interpret an importpath string that ends in
"/..." by running `go list` and parsing its output to find `package
main`s matched by that wildcard.

Since `ko build` already supports passing multiple paths, this
additively collects uniq matched importpaths, so this works, and
produces three images for the three `package main`s in this repo:

```
ko build ./... ./cmd/... ./
```

Matched paths are still checked if they're `package main`, so this
doesn't work:

```
ko build ./... ./pkg/build
```

(because the explicit `pkg/build` isn't `package main`)

This also doesn't work, because the wildcard importpath doesn't match any `package main`s:

```
ko build ./pkg/...
```

This _does_ work, because the set of wildcard importpaths taken together matches at least one `package main`:

```
ko build ./cmd/... ./pkg/...
```

Fixes https://github.com/google/ko/issues/643
Fixes https://github.com/google/ko/issues/273

### DIFF
--- a/.github/workflows/modules-integration-test.yaml
+++ b/.github/workflows/modules-integration-test.yaml
@@ -8,6 +8,7 @@ jobs:
   test:
     name: Module Tests
     strategy:
+      fail-fast: false
       matrix:
         go-version: [1.16.x, 1.17.x]
     runs-on: 'ubuntu-latest'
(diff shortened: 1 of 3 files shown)


mattmoor (approved):
Well that seems nice an easy 👍

imjasonh:
> Well that seems nice and easy 👍

Integration tests fail while I'm using `packages.Load` instead of shelling out to `go list` -- I'll either figure it out, or just switch to `go list` until I figure it out.

imjasonh:
Maybe helpful: https://pkg.go.dev/golang.org/x/tools/go/buildutil#ExpandPatterns

Issue #642, opened 2022-03-09
Sharp increase in resource usage between v0.9 and v0.10


antoineco:
In a project with [about 60 `ko://` references](https://github.com/triggermesh/triggermesh/blob/main/config/500-controller.yaml#L59), we are seeing a sudden increase in the resource usage in our CI workers, causing all our release jobs to fail due to `go build` processed being killed.

Judging by the logs (`"Building github.com/triggermesh/triggermesh/cmd/..."`), it seems like builds are now started all at once, without respecting `nprocs`. This wasn't the in v0.9 (again, only judging by the logs).

Before

![image](https://user-images.githubusercontent.com/3299086/157438381-4caf3fc2-c20b-4600-80b1-5f73c2a43ec6.png)

After

![image](https://user-images.githubusercontent.com/3299086/157438297-b1477c54-ede5-448a-9ca1-ebf4c1e4b284.png)

antoineco:
@imjasonh No, no building of multi-platform images.
I'll try disabling SBOM generation and report back.

mattmoor:
No, but I'll ask around 🤔

antoineco:
Same issue with `--sbom=none`. The CPU is maxed out and `go build` processes are being killed by the OOM killer.

dprotaso:
@antoineco how do you get graphs like that?


I'm curious if we need something like 
https://github.com/uber-go/automaxprocs

antoineco:
@dprotaso They are from CircleCI.

I didn't have time to investigate further yet, but I have a feeling parallelism isn't regulated at all anymore. Based solely on the logs output, we used to observe `nprocs` builds concurrently. Now all builds seem to start at once. And since we have ~60 images to build during releases, it OOMs almost instantly.

dprotaso:
Playing with Circle CI - by default it reads GOMAXPROCS as 36, but using automaxprocs it calculates it to be 4 (at least for my large instance). So I think using  automaxprocs would be beneficial here to figure out the default in a container environment.

But GOMAXPROCS has been in use for quite some time in `ko` so I wonder if the prior parallelization code wasn't working as expected?

antoineco:
Oh wow 36 for 8 CPU cores, that's unexpectedly high. Thanks for looking into it!

dprotaso:
@antoineco can you test the above change in our CI?

antoineco:
@dprotaso sure thing.
[...]

PR #649, opened 2022-03-14
use automaxprocs to calculate the number of CPUs

dprotaso:
this is  useful for container environments

Addresses: https://github.com/google/ko/issues/642


dprotaso:
Related issue: https://github.com/golang/go/issues/33803

dprotaso:
Essentially GOMAXPROCS is not cgroup aware so it reports the total CPUs on the machine instead of the # of CPUs allocated to the container

dprotaso:
got the typo fixed upstream - force pushing the fix

imjasonh (approved):
This seems worth doing, and unlikely to cause any new problems. After this is merged I'll try it out and see if it changes the behavior.

Thanks for this fix!

akshayjshah:
Keep in mind that `automaxprocs` doesn't (yet) support cgroups v2 :)

PR #667, opened 2022-03-21
Bump deps to silence security alerts

imjasonh:
- ~~`go get -u github.com/sigstore/cosign` for https://github.com/google/ko/security/dependabot/7~~
- `go get -u github.com/containerd/containerd` for https://github.com/google/ko/security/dependabot/8
- `go mod tidy && go mod vendor`

We weren't affected by either issue. For some reason dependabot thought the containerd issue couldn't be automatically upgraded, but it seems to be just fine.


imjasonh:
`misspell` doesn't ignore `vendor` 😢

imjasonh:
Bumped deps again to pick up the fix for https://github.com/google/ko/security/dependabot/10 -- I'm pretty sure we're not susceptible to it, but just in case.

`go get -u ./... && go mod tidy && go mod vendor`

PR #668, opened 2022-03-22
go1.18: Preserve trailing tabs while massaging `go version -m` output

### LINKED ISSUE #666: Regression in `v0.11.1`: Fails with `could not parse Go build info`
I saw couple of issues referencing a similar error, but this should have been resolved in `v0.11.x`. However, we just noticed in our pipelines, that with `v0.11.1` the following error is thrown (again):

```console
error during command execution:error processing import paths in "-": error resolving image references: parsing build info: could not parse Go build info: line 5: expected 3 columns for replacement; got 2
```

Reverting the pipeline back to `v0.11.0` fixes it.

Go version: `v1.17.7`

Example workflow run where this is observed with `v0.11.1`: https://github.com/vmware-tanzu/sources-for-knative/runs/5621359754?check_suite_focus=true

Working run reverting back to `v0.11.0`:
https://github.com/vmware-tanzu/sources-for-knative/runs/5626066023?check_suite_focus=true

tstromberg:
Allows ko on both go v1.17 and v1.18 to parse `go version -m` output that contains a local replace directive. For example:

```shell
./main: go1.18
	path	command-line-arguments
	dep	secret.dev/api	v0.0.0-00010101000000-000000000000
	=>	../api	(devel)	
	
```

The issue is that `ParseBuildInfo` assumes there is a 3rd column that contains a checksum, which is not applicable to local replace directives. This patch inserts the extra tab, allowing it to have an empty checksum.

Fixes #666


jonjohnsonjr (approved):

imjasonh (commented):
Can we add this to massageBuildInfo?

imjasonh:
https://github.com/google/ko/blob/main/internal/sbom/sbom.go#L29

tstromberg:
Ooh, I didn't notice `massageBuildInfo` before. Not only is the perfect place to implement this, but uhh.. it apparently the source of the bug too! I'll revise the patch.

imjasonh:
> Ooh, I didn't notice `massageBuildInfo` before. Not only is the perfect place to implement this, but uhh.. it apparently the source of the bug too! I'll revise the patch.

Oh no. 😭

tstromberg:
Thanks again for the `massageBuildInfo` tip. This PR has been updated to address the root cause rather than putting lipstick on the pig. :)

PR #670, opened 2022-03-24
Add Oracle Cloud Functions

mattn:
I don't believe this PR will be merged since the docker image string that ko outputs has to be replaced using sed for OCI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,14 @@ az containerapp update \
   --image $(ko build ./cmd/app)
 ```
 
+Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
+
+```
+fn update function \
+  my-container-app my-function-name \
+  --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
+```
+
 * Note: The image must be pushed to [ACR](https://azure.microsoft.com/services/container-registry/) or other registry service.
 See [official docs](https://docs.microsoft.com/azure/container-apps/) for more information.
 
--- a/README.md
+++ b/README.md
@@ -166,12 +166,7 @@ fn update function \
   --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
 ```
 
-* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/), `KO_DOCKER_REPO` should point to the registory with following form.
-
-```
-KO_DOCKER_REPO=<region-key>.ocir.io/<tenancy-namespace>/<repo-prefix>
-```
-See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
+* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/). See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -171,6 +171,7 @@ fn update function \
 ```
 KO_DOCKER_REPO=<region-key>.ocir.io/<tenancy-namespace>/<repo-prefix>
 ```
+See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -163,7 +163,7 @@ Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
 ```
 fn update function \
   my-container-app my-function-name \
-  --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
+  --image $(ko build --bare -P . | sed 's/@sha256:/:sha256-/')
 ```
 
 * Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/). See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
--- a/README.md
+++ b/README.md
@@ -155,6 +155,9 @@ az containerapp update \
   --image $(ko build ./cmd/app)
 ```
 
+* Note: The image must be pushed to [ACR](https://azure.microsoft.com/services/container-registry/) or other registry service.
+See [official docs](https://docs.microsoft.com/azure/container-apps/) for more information.
+
 Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
 
 ```
@@ -163,8 +166,11 @@ fn update function \
   --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
 ```
 
-* Note: The image must be pushed to [ACR](https://azure.microsoft.com/services/container-registry/) or other registry service.
-See [official docs](https://docs.microsoft.com/azure/container-apps/) for more information.
+* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/), `KO_DOCKER_REPO` should point to the registory with following form.
+
+```
+KO_DOCKER_REPO=<region-key>.ocir.io/<tenancy-namespace>/<repo-prefix>
+```
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -155,6 +155,14 @@ az containerapp update \
   --image $(ko build ./cmd/app)
 ```
 
+Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
+
+```
+fn update function \
+  my-container-app my-function-name \
+  --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
```
imjasonh, line 163: I don't believe this will work, since `ko` won't push to that tag. Have you confirmed this works in your environment?

README.md
```diff
@@ -155,6 +155,14 @@ az containerapp update \
   --image $(ko build ./cmd/app)
 ```
 
+Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
```
imjasonh, line 158: This section should go below the `* Note` on line 166, which is related to the ACR functions example above this.

Also, if there are any restrictions that Oracle Cloud Functions imposes, such as a required base image or registry, that would be useful to spell out in a similar `* Note` section.

imjasonh (commented):
Thanks for this contribution! I haven't personally used Oracle Cloud myself, so I'll be relying on you to verify that these instructions work as expected.

README.md
mattn: Yes, Oracle Cloud Functions expect following form for container image.

```
<region-key>.ocir.io/<tenancy-namespace>/<repo-name>/<func-name>:<tag>
```

https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionsupdatingfunctions.htm#notesdigest

imjasonh: It sounds from reading this like OCF will accept image references by digest just fine, but that it won't automatically pull any new image with that tag until you update the function to specify the tag. Based on my reading, this is consistent with Cloud Run's behavior, and likely Lambda's, Fly.io's, etc.

Specifying images by digest is usually considered a best practice, so I'd like to recommend it here. This would also make our OCF recommendation consistent with the other deployment targets.

Can you test that this works when deploying by digest? And that the function is updated correctly when a new image by digest is specified in a subsequent call to `fn update function`? If that works, we should recommend that.

README.md
mattn: Thanks. Updated.

README.md
```diff
@@ -158,6 +158,21 @@ az containerapp update \
...
+  my-container-app my-function-name \
+  --image $(ko build -P . | sed 's/@sha256:/:sha256-/')
+```
+
+* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/), `KO_DOCKER_REPO` should point to the registory with following form.
+
+```
+KO_DOCKER_REPO=<region-key>.ocir.io/<tenancy-namespace>/<repo-prefix>
+```
+See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
```
imjasonh, line 174:
```suggestion
* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/). See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.
```

README.md
mattn:
```
fn update function my-app my-func --image $(ko bulid -P .)
```
This return error.
> Fn: Service error:InvalidParameter. Invalid Invalid Image - Image is not a valid docker image name. http status code: 400. Opc

And this too.
```
fn update function my-app my-func --image $(ko bulid .)
```

Oracle Cloud Function does not handle `registory-url/repo-name/func-name:tag-name@sha256:xxxxx` for `--image`.

imjasonh: Okay, that's unfortunate. In that case you might need to specify `--bare` so the image gets pushed to exactly `KO_DOCKER_REPO`, without any additional path components.

imjasonh: FWIW, it's also unclear from the docs whether the registry needs to be OCIR -- the docs seem to indicate that any registry can be used:

> (this documentation assumes you will be using [Oracle Cloud Infrastructure Registry](https://docs.oracle.com/en-us/iaas/Content/Registry/Concepts/registryoverview.htm#Overview_of_Registry) as your Docker registry and provides instructions accordingly).

(from https://docs.oracle.com/en-us/iaas/Content/Functions/Concepts/functionsprerequisites.htm)

mattn: Okay, updated.

mattn: If you want short history of commits, Please let me known. I will squash them.

README.md
```diff
@@ -158,6 +158,16 @@ az containerapp update \
 * Note: The image must be pushed to [ACR](https://azure.microsoft.com/services/container-registry/) or other registry service.
 See [official docs](https://docs.microsoft.com/azure/container-apps/) for more information.
 
+Or [Oracle Cloud Functions](https://www.oracle.com/cloud-native/functions/)
+
+```
+fn update function \
+  my-container-app my-function-name \
+  --image $(ko build --bare -P . | sed 's/@sha256:/:sha256-/')
```
imjasonh, line 166: Again, I'd like to avoid recommending that we reference the image by tag, for a few reasons:

1. This is not generally considered a best practice.
2. Their docs don't seem to indicate that it's required anyway. If it _is_ required to deploy by tag, I'd prefer not to recommend using it for that reason.
3. The image won't be pushed with that tag, so deploying will fail using these instructions.

Also, `--bare` and `-P` specify conflicting behavior. It should be an error to specify both, there's a bug for that somewhere. If `--bare` is required to make this work, we should add that in a `* Note` below.

We can also remove the note that OCIR is required, but we should keep the link to official docs.

mattn: >We can also remove the note that OCIR is required, but we should keep the link to official docs.

Current note is the same as that of Azure. Do you want to remove?

mattn: How about this?

>* Note: The image must be pushed to [OCIR](https://www.oracle.com/cloud-native/container-registry/) or other registry service. You need to specify `--bare` since OCI does not support image by digest. See [official docs](https://docs.oracle.com/en-us/iaas/Content/Functions/Tasks/functionscreatefncontext.htm) for more information.

imjasonh: What I mean is, the docs seem to indicate that OCIR isn't required. We should check whether that's the case before stating that it is required. It also sounds like OCF _does_ support images by digest, so we should check that too before saying it doesn't.

mattn: Okay. Could you please specify the command you want to try? (Will do that tomorrow, it is currently midnight)

imjasonh:
```
fn update function \
  my-container-app my-function-name \
  --image $(ko build ./)
```

If that doesn't work (you had a 400 before), try `--bare`. You may also want to try another registry besides OCIR, such as DockerHub or GHCR, where the image is publicly accessible. That would tell us whether OCF actually requires OCIR, or if that's just what's documented in the docs.

Thanks for sticking with this, I really appreciate the contribution. I just want to make sure it's as generally applicable and useful to anybody who finds it, and it doesn't help that I'm personally so unfamiliar with Oracle Cloud.

mattn: Unfortunately, OCF does not support external registry. When creating new function, registry and image is not possible to be registered from external. It must be provided on OCIR.

image

image

image

And also, fn cli does not treat non-OCIR image.

>Fn: Service error:InvalidParameter. Invalid Invalid image - The image must be an OCIR image in this region's registry. http status code: 400. Opc request

mattn: Additional: fn cli or OCF does not handle image by digest. So

```
fn update function \
  my-container-app my-function-name \
  --image $(ko build ./)
```
Did not work.

imjasonh: Thanks for testing that out. What was the error when you tried to deploy by digest? It might be possible that deploying by digest is allowed, but that OCIF doesn't accept image names like `ko` produces by default -- some registries don't.

If OCF _doesn't_ support deploying functions from images by digest, I'm inclined not to recommend its use, since this is generally considered a best practice.

mattn:
Now I tried this patch, below works well for me.

```
fn update function \
  my-container-app my-function-name \
  --image $(ko build ./)
```

Can I reopen this?

PR #689, opened 2022-04-21
Include tags when --push=false is set

### LINKED ISSUE #688: [Regression] --tags and --tag-only are not respected
## Bug Report

When running `build`, `resolve`, or `apply` with the `-t` or `--tags` option, the tag is not applied to the built image and the tag is not used in generated manifests. In addition, the `--tag-only` option is not respected at all in `apply` or `resolve`.

### Expected Behavior

When `-t` or `--tags` option is used, ko images have the tag applied.
When the `--tag-only` option is used, image references in `ko resolve` and `ko apply` reference images by tag.

### Actual Behavior

Images do not have tags, and only the image sha digest is used with image references for ko apply/resolve.

adambkaplan:
- Refactor noop into its own publisher.
- Rename Option -> DefaultOption.
- Use :tag@sha format in digest when --push=false is set.
- Allow --tag-only to be used with --push=false. This includes
  validations that prevent --tag-only to be used with no tags or the
  `latest` tag.

Fixes #688

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -224,10 +224,15 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 		// If not publishing, at least generate a digest to simulate
 		// publishing.
 		if len(publishers) == 0 {
-			publishers = append(publishers, nopPublisher{
-				repoName: repoName,
-				namer:    namer,
-			})
+			noop, err := publish.NewNoOp(repoName,
+				publish.NoOpWithNamer(namer),
+				publish.NoOpWithTags(po.Tags),
+				publish.NoOpWithTagOnly(po.TagOnly),
+			)
+			if err != nil {
+				return nil, err
+			}
+			publishers = append(publishers, noop)
 		}
 
 		return publish.MultiPublisher(publishers...), nil
@@ -250,24 +255,6 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 	// Wrap publisher in a memoizing publisher implementation.
 	return publish.NewCaching(innerPublisher)
 }
-
-// nopPublisher simulates publishing without actually publishing anything, to
-// provide fallback behavior when the user configures no push destinations.
-type nopPublisher struct {
-	repoName string
-	namer    publish.Namer
-}
-
-func (n nopPublisher) Publish(_ context.Context, br build.Result, s string) (name.Reference, error) {
-	s = strings.TrimPrefix(s, build.StrictScheme)
-	h, err := br.Digest()
-	if err != nil {
-		return nil, err
-	}
-	return name.NewDigest(fmt.Sprintf("%s@%s", n.namer(n.repoName, s), h))
-}
-
-func (n nopPublisher) Close() error { return nil }
 
 // resolvedFuture represents a "future" for the bytes of a resolved file.
 type resolvedFuture chan []byte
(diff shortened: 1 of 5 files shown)


pkg/publish/default.go
```diff
@@ -47,8 +47,8 @@ type defalt struct {
 	insecure  bool
 }
 
-// Option is a functional option for NewDefault.
-type Option func(*defaultOpener) error
+// DefaultOption is a functional option for NewDefault.
+type DefaultOption func(*defaultOpener) error
```
imjasonh, line 51: This change will probably cause some pain to consumers of this package as a library. Can we keep this as `Option`?

pkg/publish/noop.go
```diff
@@ -0,0 +1,99 @@
...
+
+type noOpOpener struct {
+	repoName string
+	namer    Namer
+	tags     []string
+	tagOnly  bool
+}
+
+// NoOpOption provides functional options to the NoOp publisher.
+type NoOpOption func(*noOpOpener)
```
imjasonh, line 43: You missed a perfect opportunity to name this "NoOption" or "NOption" 😆

imjasonh (changes requested):
Hey Adam, sorry I let this PR linger for so long! The change looks good, if you're still interested in pursuing it.

I think it would be helpful to have some test that checks that the original buggy behavior is fixed by this change, otherwise it's a bit hard to tell how the noop publisher refactor is the solution. Future changes to noop publishing might regress back to the buggy behavior without a backstop unit test.

If that's something you're interested in, I promise to review it more quickly this time 😞

pkg/publish/default.go
adambkaplan: Reverted.

adambkaplan:
@imjasonh finally got around to adding an e2e test. Hopefully we can get this in before the big move to ko-build!

pkg/commands/resolver.go
```diff
@@ -224,10 +224,15 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 		// If not publishing, at least generate a digest to simulate
 		// publishing.
 		if len(publishers) == 0 {
-			publishers = append(publishers, nopPublisher{
-				repoName: repoName,
-				namer:    namer,
-			})
+			noop, err := publish.NewNoOp(repoName,
```
imjasonh, line 227: Unless the intention here is to export more functionality around no-op publishing, I'd kind of like to avoid increasing the public surface area of this change. We can do that just by passing more fields to `nopPublisher` and not exporting anything new.

If the intention _is_ to export more no-op publishing functionality, I'd like to understand more about the use case so we can make sure this change helps the issue.

In any case I think we can get this in before v0.12 and the repo move, if you're interested.

adambkaplan: I don't think there is any further no-op publishing functionality to export.

imjasonh: What I'm saying is, this change exports new methods in `pkg/publish` to enable no-op publishing. But that's not necessary just to get the CLI behavior you're proposing -- we could do that just by passing more options into the (unexported) `nopPublisher` struct.

I think we should try to get this CLI behavior change done without exporting new methods, since once we export it, we're married to it forever. AFAIK there aren't a lot of use cases for no-op publishing, vs local publishing or tarball publishing, which are better supported don't-push-to-a-registry options, except in this very narrow use case.

PR #690, opened 2022-04-21
check and use the latest go available for the defined version 

cpanato:
- check and use the latest go available for the defined version 
- update cosign action and version

cc @imjasonh

### DIFF
--- a/.github/workflows/style.yaml
+++ b/.github/workflows/style.yaml
@@ -4,9 +4,6 @@ on:
   pull_request:
     branches: ['main']
 
-env:
-  GO_VERSION: '1.17'
-
 jobs:
 
   gofmt:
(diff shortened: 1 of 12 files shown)


.github/workflows/style.yaml
```diff
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: ['main']
 
+env:
+  GO_VERSION: '1.17'
```
imjasonh, line 8: I'd kind of prefer to just inline these, instead of using envs to share it.

imjasonh (changes requested):

.github/workflows/style.yaml
cpanato: reverted, thanks for the review and feedback

imjasonh (approved):

PR #698, opened 2022-05-05
Fix texts regarding SBOM

otms61:
Thank you for the great project. I am learning about SBOM and found that help texts and the error message seem to be incorrect.
`ko` can generate cyclonedx format, but help texts and the error message does not include cyclonedx.

I fixed help texts and the error message and ran `./hack/update-codegen.sh`.


imjasonh (approved):
🤦🏻‍♂️ Wow.

Anyway, thanks for fixing this!

Issue #541, opened 2021-12-14
Use Cobra flag helpers for better shell completion


imjasonh:
Cobra has a helper methods which can make shell completion a little nicer:

- [`MarkFlagFilename`](https://pkg.go.dev/github.com/spf13/cobra#MarkFlagFilename) tells Cobra that the flag represents a filename. This tells the `ko completion` logic to generate helpers to match filenames. (e.g., `ko apply -f <tab>`)
- [`MarkFlagRequired`](https://pkg.go.dev/github.com/spf13/cobra#MarkFlagRequired) tells Cobra that the flag is required, and prioritizes in completion if it's unset (e.g., `ko apply <tab>` prioritizes `-f` over `--preserve-import-paths`).

Audit flags we configure and see if there are any that would benefit from being marked this way. Check that the expected behavior is actually observed when shell completion is configured.

imjasonh:
I still want this.

MohabMohamed:
Hello @imjasonh , is this issue still available?
I would love to work on it

MohabMohamed:
Thank you

MohabMohamed:
@imjasonh I made the required change in PR #705, but are there any flags other than `-f` that need to be required? 
as I didn't use ko before.

PR #705, opened 2022-05-16
Add Cobra flag helpers for file options

MohabMohamed:
- Marked -f flag as a filename or a directory for better completion.
- Marked it as required for every command that requires this flag.

Fixes https://github.com/google/ko/issues/541

### DIFF
--- a/pkg/commands/options/filestuff.go
+++ b/pkg/commands/options/filestuff.go
@@ -35,21 +35,15 @@ func AddFileArg(cmd *cobra.Command, fo *FilenameOptions) {
 	cmd.Flags().BoolVarP(&fo.Recursive, "recursive", "R", fo.Recursive,
 		"Process the directory used in -f, --filename recursively. Useful when you want to manage related manifests organized within the same directory.")
 
-	err := cmd.MarkFlagFilename("filename", "yaml", "yml", "json")
-
-	if err != nil {
+	if err := cmd.MarkFlagFilename("filename", "yaml", "yml", "json"); err != nil {
 		log.Fatalf("Error marking filename flag as Cobra's filename: %v", err)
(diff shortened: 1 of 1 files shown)


pkg/commands/options/filestuff.go
```diff
@@ -34,6 +34,24 @@ func AddFileArg(cmd *cobra.Command, fo *FilenameOptions) {
 		"Filename, directory, or URL to files to use to create the resource")
 	cmd.Flags().BoolVarP(&fo.Recursive, "recursive", "R", fo.Recursive,
 		"Process the directory used in -f, --filename recursively. Useful when you want to manage related manifests organized within the same directory.")
+
+	err := cmd.MarkFlagFilename("filename", "yaml", "yml", "json")
```
imjasonh, line 38: Have you checked that the combination of `MarkFlagFilename` and `MarkFlagDirname` work together, to prefer shell completion for file names _or_ directory names? It's not clear from the docs that these won't conflict. I can also check if that would be helpful.

MohabMohamed: sorry for coming back late as I was busy.

> Have you checked that the combination of `MarkFlagFilename` and `MarkFlagDirname` work together, to prefer shell completion for file names _or_ directory names? It's not clear from the docs that these won't conflict. I can also check if that would be helpful.

I can't get kubectl to work on my local machine, and the docs are not clear. but however I checked cobra's source code, and it handles this part by a map called annotations and filename has a key different from FlagDirname and it iterates on the keys while doing auto-completion.

so if you got some time can you check it's behavior?

MohabMohamed:
hello @imjasonh ,
can you please review the PR when you have free time?

Issue #729, opened 2022-06-14
Generate SLSA provenance for your release builds


laurentsimon:
Hi

I am one of the authors of the SLSA3+ builder for GitHub workflows (https://github.com/slsa-framework/slsa-github-generator projects).

We released the v1 of the SLSA3+ builder last week. It will be officially announced during the [Open Source Summit](https://events.linuxfoundation.org/open-source-summit-north-america/) next week.

We are reaching out to projects to see if they'd be interested in using it. The [scorecard project](https://github.com/ossf/scorecard/blob/main/.github/workflows/slsa-goreleaser.yml) has recently added support for their Linux amd64 build.

Practically speaking, adoption should be easy. The configuration file is similar to Goreleaser's, except that we support a single build for this first release. You can select an OS/Arch to generate provenance for, and disable the correpsonding build for Goreleaser using the `ignore option`, as explained in [builders/go/README.md#migration-from-goreleaser](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/go/README.md#migration-from-goreleaser)

I have prepared https://github.com/google/ko/pull/730 to help you see what the changes are.

Feedback welcome!

/cc @asraa @ ianlewis

developer-guy:
@imjasonh I think we can close this because it has already been fixed by #730.

PR #730, opened 2022-06-14
feat: generate SLSA provenance for release binaries

laurentsimon:
Generate non-forgeable provenance, as proposed in https://github.com/google/ko/issues/729

Below is an example of what the provenance looks like:
```
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "predicateType": "https://slsa.dev/provenance/v0.2",
  "subject": [
    {
      "name": "ko-copy_5.0.1_Windows_arm64.tar.gz",
      "digest": {
        "sha256": "01a6a93df134ff638c7da1ccdd735398a979cb08c9d103e64d3a22f29e23223b"
      }
    },
    {
      "name": "ko-copy_5.0.1_Linux_i386.tar.gz",
      "digest": {
        "sha256": "520bcf95c284d0e9c43f02ae516f7ad35d26a24d9103d49e5977ed9a94511017"
      }
    },
    ...
  ],
  "predicate": {
    "builder": {
      "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v1.0.0"
    },
    "buildType": "https://github.com/slsa-framework/slsa-github-generator@v1",
    "invocation": {
      "configSource": {
        "uri": "git+https://github.com/google/ko@refs/tags/v0.0.4",
        "digest": {
          "sha1": "b6e9b26f5c8af4fbcb880f14ac9ad961a2b84e43"
        },
        "entryPoint": ".github/workflows/slsa-release.yml"
      },
      "parameters": {},
      "environment": {
        "arch": "X64",
        "github_actor": "laurentsimon",
        "github_actor_id": "64505099",
        "github_base_ref": "",
        "github_event_name": "push",
        "github_event_payload": {...},
        "github_head_ref": "",
        "github_ref": "refs/tags/v0.0.4",
        "github_ref_type": "tag",
        "github_repository_id": "501698356",
        "github_repository_owner": "laurentsimon",
        "github_repository_owner_id": "64505099",
        "github_run_attempt": "1",
        "github_run_id": "2497311143",
        "github_run_number": "4",
        "github_sha1": "b6e9b26f5c8af4fbcb880f14ac9ad961a2b84e43",
        "os": "ubuntu20"
      }
    },
    "buildConfig": nil,
    "metadata": {
      "buildInvocationID": "2497311143-1",
      "completeness": {
        "parameters": true,
        "environment": false,
        "materials": false
      },
      "reproducible": false
    },
    "materials": [
      {
        "uri": "git+https://github.com/google/ko@refs/tags/v0.0.4",
        "digest": {
          "sha1": "b6e9b26f5c8af4fbcb880f14ac9ad961a2b84e43"
        }
      },
      {
        "uri": "https://github.com/actions/virtual-environments/releases/tag/ubuntu20/20220605.1"
      }
    ]
  }
}
```


.github/workflows/slsa-goreleaser.yml
```diff
@@ -0,0 +1,35 @@
...
+
+jobs:
+  # Generate ldflags dynamically.
+  args:
+    runs-on: ubuntu-latest
+    outputs:
+      version: ${{ steps.ldflags.outputs.version }}
+    steps:
+      - id: checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v3.0.2
```
imjasonh, line 18: These comments fall out of date pretty quickly when dependabot bumps them. Can you remove that part?

README.md
```diff
@@ -25,12 +25,28 @@ tool for Kubernetes applications ([See below](#Kubernetes-Integration)).
 
 ### Install from [Releases](https://github.com/google/ko/releases)
 
-```
-VERSION=TODO # choose the latest version
-OS=Linux     # or Darwin
-ARCH=x86_64  # or arm64, i386, s390x
-curl -L https://github.com/google/ko/releases/download/v${VERSION}/ko_${VERSION}_${OS}_${ARCH}.tar.gz | tar xzf - ko
-chmod +x ./ko
+We are one of the first repositories to generate SLSA3+ provenance using the OpenSSF's [slsa-framework/slsa-github-generator](https://github.com/slsa-framework/slsa-github-generator) project.
```
imjasonh, line 28: This comment will (hopefully) fall out of date pretty fast. Can we just say, "here's how to install a release", and then later, "here's how to verify the released artifact"? The verifying-the-release section may even make more sense lower in the README, since folks who already have `ko` installed may not want to scroll past it.

(Longer term I'd love to not have one big README and instead have a real website with docs and tutorials and everything, but until then...)

imjasonh (commented):
Cool! Always happy to test out better supply chain security workflows 😎

.github/workflows/slsa-goreleaser.yml
laurentsimon: Fixed. (Note: Renovatebot PRs update the comment, but not dependabots's.)

README.md
laurentsimon: I've made some changes. Can you take a look and tell me what you think?

laurentsimon:
> Cool! Always happy to test out better supply chain security workflows 😎

Awesome, thanks!

.slsa-goreleaser.yml
```diff
@@ -0,0 +1,13 @@
+version: 1
```
imjasonh, line 1: Does this need to be its own file? I'm worried about this diverging from goreleaser eventually.

Or, if the idea is to eventually only have the SLSA-flavored goreleaser workflow, I wonder if you've considered just adding the SLSA provenance functionality to goreleaser proper, and not having your own fork?

.github/workflows/slsa-goreleaser.yml
```diff
@@ -0,0 +1,35 @@
...
+          echo "::set-output name=version::$(git describe --tags --always --dirty | cut -c2-)"
+
+  # Trusted builder.
+  build:
+    permissions:
+      id-token: write
+      contents: write
+      actions: read
+    needs: args
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@v1.0.0
```
imjasonh, line 32: I'm not as familiar with this style of workflow reference, will dependabot keep this up to date for us as slsa-github-generator puts out new releases?

.github/workflows/slsa-goreleaser.yml
```diff
@@ -0,0 +1,35 @@
...
+          fetch-depth: 0
+      - id: ldflags
+        run: |
+          echo "::set-output name=version::$(git describe --tags --always --dirty | cut -c2-)"
+
+  # Trusted builder.
+  build:
+    permissions:
+      id-token: write
+      contents: write
```
imjasonh, line 29: This needs `contents: write` in order to attach to Releases right? It's not going to push anything back to the repo I hope?

.github/workflows/slsa-goreleaser.yml
```diff
@@ -0,0 +1,35 @@
...
+    outputs:
+      version: ${{ steps.ldflags.outputs.version }}
+    steps:
+      - id: checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579
+        with:
+          fetch-depth: 0
+      - id: ldflags
+        run: |
+          echo "::set-output name=version::$(git describe --tags --always --dirty | cut -c2-)"
```
imjasonh, line 23: Is this really the best way to get this information? Can't we use `${{github.sha}}` or `${{github.commit}}`?

.github/workflows/slsa-goreleaser.yml
laurentsimon: it should, yes

.slsa-goreleaser.yml
laurentsimon: > Does this need to be its own file? I'm worried about this diverging from goreleaser eventually.

right now it needs to be its own file, yes.

> Or, if the idea is to eventually only have the SLSA-flavored goreleaser workflow, 

correct. In the next release we will support multiple builds.

> I wonder if you've considered just adding the SLSA provenance functionality to goreleaser proper, and not having your own fork?

We have (see https://github.com/goreleaser/goreleaser/issues/2737) and we emailed them, but no concrete plan yet :/

.github/workflows/slsa-goreleaser.yml
imjasonh: Yeah, that seems easier to understand to me than running `git describe` and parsing output.

laurentsimon: Done. Thanks

laurentsimon:
Any further concern I should address?

.slsa-goreleaser.yml
imjasonh: It sounds like there's still some discussion happening on https://github.com/goreleaser/goreleaser/issues/2737, and I'd kind of like that to land somewhere before proceeding to fork the release process here. If you need any help getting the SLSA provenance generation landed in goreleaser, let me know.

I'd rather not have a forked release process per-target-arch, and long term, I'd prefer to rely on the standard upstream goreleaser if at all possible.

laurentsimon: > It sounds like there's still some discussion happening on [goreleaser/goreleaser#2737](https://github.com/goreleaser/goreleaser/issues/2737), and I'd kind of like that to land somewhere before proceeding to fork the release process here. If you need any help getting the SLSA provenance generation landed in goreleaser, let me know.

Would love your help :-)

> I'd rather not have a forked release process per-target-arch, and long term, I'd prefer to rely on the standard upstream goreleaser if at all possible.

Any particular reason? We'd like to learn about blockers to adoption to improve. Any feedback welcome.

imjasonh: I'm not sure I understand what multi-build support entails, or v2 of what project. In my ideal scenario, goreleaser would be able to be configured to generate and attach attestation documents to releases, which users could use to verify the provenance of each goreleaser-built artifact.

Ideally that's done by just adding `generate-attestations: true` in the goreleaser.yaml somewhere, and no new dependencies on other projects.

laurentsimon: Sorry for the confusion. What I mean is that in the next release of our SLSA builder, we will support building for multiple arch/os at once, like Goreleaser does. So it would be easy to migrate to our builder instead of Goreleaser with a similar config file.

> Ideally that's done by just adding generate-attestations: true in the goreleaser.yaml somewhere, and no new dependencies on other projects.

Any update on #2737?

laurentsimon:
I've updated the PR:
1. Keep the current GoReleaser config
2. Add a call to the SLSA generator to create provenance using the output of GoReleaser. The provenance file attest that the build was created in the release.yaml workflow, at a particular commit.

A follow-up PR is to verify the binary pulled in https://github.com/imjasonh/setup-ko/blob/main/action.yml#L41. We're still working on a GHA installer for the verifier, so that is not ready yet.

Appreciate your feedback. Thanks again

.github/workflows/release.yml
```diff
@@ -16,8 +19,29 @@ jobs:
...
+
+  provenance:
+    needs: [goreleaser]
+    permissions:
+      actions: read   # To read the workflow path.
+      id-token: write # To sign the provenance.
+      contents: write # To add assets to a release.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
+    with:
+      base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
```
laurentsimon, line 46: note: this will create an attestation file `attestation.intoto.jsonl` by default. The name can be customized with the `attestation-name: xxx.intoto.jsonl` if you want.

laurentsimon:
Fyi, this PR is ready to be reviewed.
Let me know if you have questions. Thanks!

imjasonh:
> Fyi, this PR is ready to be reviewed. Let me know if you have questions. Thanks!

I probably won't have the bandwidth to review this closely for at least a week or so. I'm excited to see this progressing though, and especially now that it seems to require fewer changes to the release process. Thanks for pushing on that! 👍

laurentsimon:
Friendly ping. Would love to hear some feedback. Thanks again.

imjasonh:
This looks pretty good to me. Would it be possible to have the release workflow also attempt to verify it, so we can detect quickly if verification is silently broken due to some future change?

@jonjohnsonjr any concerns?

laurentsimon:
> This looks pretty good to me. Would it be possible to have the release workflow also attempt to verify it, so we can detect quickly if verification is silently broken due to some future change?
> 
> @jonjohnsonjr any concerns?

I can do that. How to you typically create the tag? Via a git command or via the GitHub UX?

imjasonh:
I usually do it through the UI, but I think triggering based on new tags should catch both cases.

laurentsimon:
I have added the verification job. PTAL.

laurentsimon:
@imjasonh any other feedback you'd like me to address?

imjasonh (approved):

developer-guy:
As we completed the SLSA provenance generation for release binaries, should we move on with the same for ko images? @imjasonh ^^

PR #737, opened 2022-06-29
Update base image to ghcr.io/distroless/static:latest

### LINKED ISSUE #722: RFC: Switch default base image to ghcr.io/distroless/static
The current default base image is `gcr.io/distroless/static:nonroot`, which we switched two about 2 years ago in https://github.com/google/ko/issues/160 (from the non-`nonroot` variant).

I'd like to propose we switch the default base image to `ghcr.io/distroless/static` (which is also non-root), produced using `apko` nightly, from here https://github.com/distroless/static

Like `gcr.io/distroless/static:nonroot`, `ghcr.io/distroless/static` is signed (keylessly!). It also includes a generated SBOM of the base image contents, and an attestation about how it was built.

Being on GitHub Actions, its build process and logs are more visible and publicly auditable.

Being defined in not-Bazel, in a not-Google-owned repo, makes it easier to contribute changes to it.

---

If this proposal is accepted, I'd like to suggest adding a warning when `defaultBaseImage` is not set, saying that the change is coming in a future release.

If you want to retain the existing base image, simply add `defaultBaseImage: gcr.io/distroless/static:nonroot`. To opt-in to the change before it's imposed on you automatically in a future release, `defaultBaseImage: ghcr.io/distroless/static`.

Then we cut a release with that warning, wait a release, change the default base image and remove the warning.

WDYT?

@jonjohnsonjr @mattmoor

imjasonh:
Fixes #722 

`ghcr.io/distroless/static` now includes tzdata, but this also adds a test that ensures timezone conversions are unaffected in case we change the base image again in the future.

I don't want to update the README until after we cut a release.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -50,7 +50,7 @@ jobs:
 
         # Build and run the test/ binary, which should log "Hello there" served from KO_DATA_PATH
         testimg=$(go run ./ build ./test --platform=${PLATFORM} --preserve-import-paths)
-        docker run ${testimg} --wait=false 2>&1 | grep "Hello there"
+        docker run ${testimg} --wait=false 2>&1 | tee >(grep "Hello there") # Log all output too.
 
         # Check that symlinks in kodata are chased.
         # Skip this test on Windows.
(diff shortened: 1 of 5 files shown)


jonjohnsonjr:
I'm ok to merge this but with the caveat that I wouldn't hesitate to revert it if something unexpectedly breaks.

jonjohnsonjr (approved):

Issue #675, opened 2022-03-29
Compile to WASM


imjasonh:
Go can compile to WASM (`GOOS=js GOARCH=wasm`), and WASM can be packaged into images and executed on K8s clusters using [krustlet](https://krustlet.dev): https://developer.okta.com/blog/2022/01/28/webassembly-on-kubernetes-with-rust

So, maybe `ko` can build a WASM executable, and put it in a multiarch image, alongside the other ko-built archs.

This wouldn't work for Go executables that need to call into their base images (I don't think WASM images have multiple layers, and the environment isn't very linux-ey...yet?), but it should mostly maybe kinda work if they're just static binaries, which is exactly ko's sweet spot.

An example WASM image that kruslet can apparently run: https://explore.ggcr.dev/?image=webassembly.azurecr.io%2Fhello-wasm%3Av1 -- note media types and the annotation.

Questions:
- What OCI platform does kruslet expect when it's given a multiplatform image? `js/wasm`?

- How should users express this? `ko build ./ --wasm` to additionally produce a WASM image in the index?

imjasonh:
> * What OCI platform does kruslet expect when it's given a multiplatform image? `js/wasm`?

Kruslet doesn't appear to support pulling from indexes at all, only OCI artifacts with the WASM media types. So this is probably not going to be something worth shoehorning into ko.

This might still make sense as a separate Go-specific pipeline for building WASM images, a la [wasm-to-oci](https://github.com/engineerd/wasm-to-oci) with fewer steps, but it won't coexist alongside regular OCI images.

imjasonh:
Reopening as I still think this is an idea worth pulling on.

imjasonh:
cf: Istio's WASM plugins are also packaged as OCI images ([building tutorial](https://github.com/istio-ecosystem/wasm-extensions/blob/master/doc/how-to-build-oci-images.md), [spec](https://github.com/solo-io/wasm/blob/master/spec/spec-compat.md))

TL;DR: In this case, the format seems to be:

- the image must be a single-arch OCI image (not index)
- the image must contain one layer
- the layer must contain a file named `plugin.wasm` (entrypoint is ignored)
- and optionally a `runtime-config.json`

This seems fine for Istio, there's not a lot of use cases for multi-platform WASM and non-WASM plugins there I think, but it's something to consider as folks have gone down this path before and made certain decisions.

imjasonh:
[...]

PR #738, opened 2022-06-30
feat: wasi platform build

hown3d:
This PR implements the scenario described in #675 by using tinygo as the builder for target **WASI**.

Documentation is not yet done, but can be created if this feature is wanted.

SBOM generation is currently not implemented, since tinygo doesn't produce the same buildInfo that go does.

Signed-off-by: Höhl, Lukas <lukas.hoehl@accso.de>


pkg/build/gobuild.go
```diff
@@ -720,30 +742,41 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
...
 
 	appDir := "/ko-app"
-	appPath := path.Join(appDir, appFilename(ref.Path()))
+	var filename string
+	if platform.String() == "wasm/wasi" {
+		// module.wasm.image/variant=compat-smart expects the entrypoint binary to be of form .wasm
+		filename = appFilename(fmt.Sprintf("%v.wasm", ref.Path()))
+	} else {
+		filename = appFilename(ref.Path())
+	}
```
imjasonh, line 777:
```suggestion
	filename := appFilename(ref.Path())
	if platform.String() == "wasm/wasi" {
		// module.wasm.image/variant=compat-smart expects the entrypoint binary to be of form .wasm
		filename += ".wasm"
	}
```

It may be a silly nit, but I tend to prefer this since it makes it clearer that the wasi case is unusual, and normally this will just be `ref.Path()`

pkg/build/gobuild.go
```diff
@@ -784,7 +817,7 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 		cfg.Config.Entrypoint = []string{`C:\ko-app\` + appFilename(ref.Path())}
 		updatePath(cfg, `C:\ko-app`)
 		cfg.Config.Env = append(cfg.Config.Env, `KO_DATA_PATH=C:\var\run\ko`)
-	} else {
+	} else if platform.String() != "wasm/wasi" {
```
imjasonh, line 820: If we keep doing this comparison, I wonder if it warrants having a helper method `isWasi(platform)` (and `isWindows(platform)`) to make this easier to read.

Then we can also avoid the string comparison without bloating the callsite:

```
func isWasi(p v1.Platform) bool {
  return p.OS == "wasm" && p.Architecture == "wasi"
}
```

pkg/build/gobuild.go
```diff
@@ -909,7 +942,12 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 		if !ok {
 			return nil, fmt.Errorf("failed to interpret base as image: %v", base)
 		}
-		res, err = g.buildOne(ctx, s, baseImage, nil)
+		// user specified a platform, so force this platform, even if it fails
+		var platform *v1.Platform
+		if len(g.platformMatcher.platforms) == 1 {
+			platform = &g.platformMatcher.platforms[0]
+		}
+		res, err = g.buildOne(ctx, s, baseImage, platform)
```
imjasonh, line 950: Why's this necessary?

pkg/commands/config.go
```diff
@@ -98,6 +99,15 @@ func getBaseImage(bo *options.BuildOptions) build.GetBase {
 		if bo.InsecureRegistry {
 			nameOpts = append(nameOpts, name.Insecure)
 		}
+
+		// look for wasm/wasi platform
+		for _, p := range bo.Platforms {
```
imjasonh, line 104: Does this mean that if a user passes `--platform=linux/amd64,wasm/wasi`, that they'll only get the wasi image?

imjasonh:
cc @squillace 👀

pkg/commands/config.go
hown3d: Ah good argument, haven't thought about providing multiple architectures with wasi. I'll figure out, if there's a good way when using wasi in multiple platforms.

pkg/build/gobuild.go
hown3d: Yes, this is a good idea. The constant comparison with platform string was kinda hacky anyway.

pkg/build/gobuild.go
```diff
@@ -720,30 +742,41 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
...
-			Created:   g.kodataCreationTime,
-			Comment:   "kodata contents, at $KO_DATA_PATH",
-		},
-	})
 
 	appDir := "/ko-app"
-	appPath := path.Join(appDir, appFilename(ref.Path()))
+	var filename string
+	if platform.String() == "wasm/wasi" {
+		// module.wasm.image/variant=compat-smart expects the entrypoint binary to be of form .wasm
```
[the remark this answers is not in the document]
hown3d: I'm currently focusing on the implementation in crun: https://github.com/containers/crun/blob/main/src/libcrun/handlers/handler-utils.c

https://github.com/containers/crun/blob/65028ce3057180395a1826d8c0906f9ff299030a/src/libcrun/custom-handler.c#L208 will check all the registered handlers by calling **wasm_can_handle_container** function from handler-utils.c.

pkg/build/gobuild.go
```diff
@@ -689,6 +705,8 @@ func (g *gobuild) configForImportPath(ip string) Config {
 	return config
 }
 
+const wasmLayerAnnotationKey = "module.wasm.image/variant"
```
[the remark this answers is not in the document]
hown3d: Specification of this annotation are here: https://github.com/solo-io/wasm/blob/master/spec/spec-compat.md

crun also has an experimental annotation **run.oci.handler** which can be used to specifiy the handler for the container (e.g. wasm, wasm-smart). More in https://man.archlinux.org/man/crun.1.en

pkg/build/gobuild.go
imjasonh: Ah okay. In that case, I think we can remove this check here: https://github.com/hown3d/ko/blob/ceac5e5fcea61bea80231a689bb990cce3198827/pkg/build/gobuild.go#L722

This does make me wonder how `ko` will behave (before and after this change) when asked to build on top of a single-platform image that isn't `linux/amd64`, with `--platform` unset... 🤔

hown3d: I created a single platform image (no OS or Architecture set) here: https://quay.io/repository/hown3d/ko-wasm-tests/manifest/sha256:2374307703312554852605270f7979f1057e4d6b98bcb9f5f7cf4cabf9ca8b4f. and used it as as the base image. The result was that, the new image didn't have architecture and os set aswell.

pkg/commands/config.go
imjasonh: I've filed a request for clarification here: https://github.com/solo-io/wasm/issues/291

hown3d:
I tried to declutter the build package a bit to provide a more readable structure. I moved everything regarding platform inside a platform and binary building into a binary package.

hown3d:
Yes, that's reasonable. I think I'll go for the restructuring first because that makes implementing wasi support a bit easier and cleaner.

hown3d:
I did manage to find some time to reiterate on this feature, and I created a new branch to start of fresh.
I'll open a second pull request with the new changes.

Issue #655, opened 2022-03-18
Create an "index SBOM" for multi-arch images


mattmoor:
Currently, as associate SBOMs with each of the constituent images in a multi-arch image, but we do not associate an SBOM with the top-level "index" manifest.

I was talking to @puerco about this (as well as https://github.com/google/ko/issues/654).

Like the manifest list itself, I believe this would be a relatively thin SBOM that somehow referenced the constituent SBOMs (maybe in the same way as the linked issue?), but I defer to @puerco as the expert on the best practices here.

mattmoor:
cc @imjasonh @jonjohnsonjr @jdolitsky

imjasonh:
What should the "index" SBOM say is included? We already point to the base image(/index) we build on using the OCI annotations, is there something else we want to surface?

If the index SBOM is just a collection of pointers to platform-specific SBOMs, we kind of already get that by virtue of the index being a collection of pointers to the images, which each themselves have SBOMs.

(But I agree, let's defer to @puerco's expertise here 😄 )

mattmoor:
Hopefully @puerco can provide an example of what we should be doing here. 🙏

mattmoor:
@puerco DM'd me this, posting here so I don't lose it 🤩 
```json
{
  "SPDXID": "SPDXRef-DOCUMENT",
  "name": "sbom-sha256:af1c5f9673f78aa7a575d627cd8a210bf6a895b0065f719a098dc035eee55a58",
  "spdxVersion": "",
  "creationInfo": {
    "created": "1970-01-01T00:00:00Z",
    "creators": [
      "Tool: apko (devel)",
      "Organization: Chainguard, Inc"
    ],
    "licenseListVersion": "3.16"
  },
  "dataLicense": "CC0-1.0",
  "documentNamespace": "https://spdx.org/spdxdocs/apko/",
  "documentDescribes": [
    "sha256-af1c5f9673f78aa7a575d627cd8a210bf6a895b0065f719a098dc035eee55a58"
  ],
  "packages": [
    {
      "SPDXID": "sha256-af1c5f9673f78aa7a575d627cd8a210bf6a895b0065f719a098dc035eee55a58",
      "name": "sha256:af1c5f9673f78aa7a575d627cd8a210bf6a895b0065f719a098dc035eee55a58",
      "versionInfo": "",
      "filesAnalyzed": false,
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "NOASSERTION",
[...]

PR #743, opened 2022-07-01
Start emitting multi-arch SBOMs for `SPDX` with `ko`

mattmoor:
Tweaking some things to prepare for the meat of: https://github.com/google/ko/issues/655

### DIFF
--- a/internal/sbom/mod_1.18.go
+++ b/internal/sbom/mod_1.18.go
@@ -20,6 +20,7 @@ package sbom
 import (
 	"fmt"
 	"runtime/debug"
+	"strings"
 )
 
 type BuildInfo debug.BuildInfo
@@ -31,4 +32,30 @@ func ParseBuildInfo(data string) (*BuildInfo, error) {
 	}
 	bi := BuildInfo(*dbi)
 	return &bi, nil
+}
+
+func modulePackageName(mod *debug.Module) string {
+	return fmt.Sprintf("SPDXRef-Package-%s-%s",
+		strings.ReplaceAll(mod.Path, "/", "."),
+		mod.Version)
+}
+
+func bomRef(mod *debug.Module) string {
+	return fmt.Sprintf("pkg:golang/%s@%s?type=module", mod.Path, mod.Version)
+}
+
+func goRef(mod *debug.Module) string {
+	path := mod.Path
+	// Try to lowercase the first 2 path elements to comply with spec
+	// https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang
+	p := strings.Split(path, "/")
+	if len(p) > 2 {
+		path = strings.Join(
+			append(
+				[]string{strings.ToLower(p[0]), strings.ToLower(p[1])},
+				p[2:]...,
+			), "/",
+		)
+	}
+	return fmt.Sprintf("pkg:golang/%s@%s?type=module", path, mod.Version)
 }
--- a/internal/sbom/spdx.go
+++ b/internal/sbom/spdx.go
@@ -71,7 +71,8 @@ func GenerateImageSPDX(koVersion string, mod []byte, img oci.SignedImage) ([]byt
 		ID:   imageID,
 		Name: imgDigest.String(),
 		// TODO: PackageSupplier: "Organization: " + bs.Main.Path
-		FilesAnalyzed: false,
+		DownloadLocation: NOASSERTION,
+		FilesAnalyzed:    false,
 		// TODO: PackageHomePage:  "https://" + bi.Main.Path,
 		LicenseConcluded: NOASSERTION,
 		LicenseDeclared:  NOASSERTION,
@@ -195,6 +196,7 @@ func GenerateIndexSPDX(koVersion string, sii oci.SignedImageIndex) ([]byte, erro
 	doc.Packages = []Package{{
 		ID:               indexID,
 		Name:             d.String(),
+		DownloadLocation: NOASSERTION,
 		FilesAnalyzed:    false,
 		LicenseConcluded: NOASSERTION,
 		LicenseDeclared:  NOASSERTION,
--- a/internal/sbom/spdx.go
+++ b/internal/sbom/spdx.go
@@ -17,16 +17,20 @@ package sbom
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"runtime/debug"
 	"strings"
 	"time"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/sigstore/cosign/pkg/oci"
 )
 
 const dateFormat = "2006-01-02T15:04:05Z"
 
-func GenerateSPDX(koVersion string, date time.Time, mod []byte, imgDigest v1.Hash) ([]byte, error) {
+func GenerateImageSPDX(koVersion string, mod []byte, img oci.SignedImage) ([]byte, error) {
 	var err error
 	mod, err = massageGoVersionM(mod)
 	if err != nil {
@@ -38,26 +42,48 @@ func GenerateSPDX(koVersion string, date time.Time, mod []byte, imgDigest v1.Has
 		return nil, err
 	}
 
-	mainPackageID := "SPDXRef-Package-" + strings.ReplaceAll(bi.Main.Path, "/", ".")
-
-	doc := Document{
-		Version:           Version,
-		DataLicense:       "CC0-1.0",
-		ID:                "SPDXRef-DOCUMENT",
-		Name:              bi.Main.Path,
-		Namespace:         "http://spdx.org/spdxdocs/" + bi.Main.Path,
-		DocumentDescribes: []string{mainPackageID},
-		CreationInfo: CreationInfo{
-			Created:  date.Format(dateFormat),
-			Creators: []string{"Tool: ko " + koVersion},
-		},
-		Packages:      make([]Package, 0, 1+len(bi.Deps)),
-		Relationships: make([]Relationship, 0, 1+len(bi.Deps)),
+	imgDigest, err := img.Digest()
+	if err != nil {
+		return nil, err
 	}
+	cfg, err := img.ConfigFile()
+	if err != nil {
+		return nil, err
+	}
+
+	doc, imageID := starterDocument(koVersion, cfg.Created.Time, imgDigest)
+
+	// image -> main package -> transitive deps
+	doc.Packages = make([]Package, 0, 2+len(bi.Deps))
+	doc.Relationships = make([]Relationship, 0, 2+len(bi.Deps))
 
 	doc.Relationships = append(doc.Relationships, Relationship{
 		Element: "SPDXRef-DOCUMENT",
 		Type:    "DESCRIBES",
+		Related: imageID,
+	})
+
+	doc.Packages = append(doc.Packages, Package{
+		ID:   imageID,
+		Name: imgDigest.String(),
+		// TODO: PackageSupplier: "Organization: " + bs.Main.Path
+		FilesAnalyzed: false,
+		// TODO: PackageHomePage:  "https://" + bi.Main.Path,
+		LicenseConcluded: NOASSERTION,
+		LicenseDeclared:  NOASSERTION,
+		CopyrightText:    NOASSERTION,
+		ExternalRefs: []ExternalRef{{
+			Category: "PACKAGE_MANAGER",
+			Type:     "purl",
+			Locator:  ociRef("image", imgDigest),
+		}},
+	})
+
+	mainPackageID := modulePackageName(&bi.Main)
+
+	doc.Relationships = append(doc.Relationships, Relationship{
+		Element: imageID,
+		Type:    "CONTAINS",
 		Related: mainPackageID,
 	})
 
@@ -74,14 +100,12 @@ func GenerateSPDX(koVersion string, date time.Time, mod []byte, imgDigest v1.Has
 		ExternalRefs: []ExternalRef{{
 			Category: "PACKAGE_MANAGER",
 			Type:     "purl",
-			Locator:  ociRef(bi.Path, imgDigest),
+			Locator:  goRef(&bi.Main),
 		}},
 	})
 
 	for _, dep := range bi.Deps {
-		depID := fmt.Sprintf("SPDXRef-Package-%s-%s",
-			strings.ReplaceAll(dep.Path, "/", "."),
-			dep.Version)
+		depID := modulePackageName(dep)
 
 		doc.Relationships = append(doc.Relationships, Relationship{
 			Element: mainPackageID,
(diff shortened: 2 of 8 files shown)


imjasonh (approved):
Looks good so far

mattmoor:
Ok, I pushed what I have so far.  I haven't tried things yet to make sure they work, but I'll do that now.

mattmoor:
Oh fun, I think there are some Go 1.18 problems :-/

mattmoor:
Here's an example of an image SBOM with the latest refactoring:
```json
{
  "SPDXID": "SPDXRef-DOCUMENT",
  "name": "sbom-sha256:8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
  "spdxVersion": "SPDX-2.2",
  "creationInfo": {
    "created": "2022-07-01T02:27:39Z",
    "creators": [
      "Tool: ko (devel)"
    ]
  },
  "dataLicense": "CC0-1.0",
  "documentNamespace": "http://spdx.org/spdxdocs/kosha256:8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
  "documentDescribes": [
    "SPDXRef-Package-sha256-8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85"
  ],
  "packages": [
    {
      "SPDXID": "SPDXRef-Package-sha256-8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
      "name": "sha256:8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
      "versionInfo": "",
      "filesAnalyzed": false,
      "licenseDeclared": "NOASSERTION",
      "licenseConcluded": "NOASSERTION",
      "downloadLocation": "",
      "copyrightText": "NOASSERTION",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE_MANAGER",
          "referenceLocator": "pkg:oci/image@sha256:8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
          "referenceType": "purl"
        }
      ]
    },
    {
      "SPDXID": "SPDXRef-Package-github.com.google.ko-(devel)",
      "name": "github.com/google/ko",
      "versionInfo": "",
      "filesAnalyzed": false,
      "licenseDeclared": "NOASSERTION",
      "licenseConcluded": "NOASSERTION",
      "downloadLocation": "https://github.com/google/ko",
      "copyrightText": "NOASSERTION",
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE_MANAGER",
          "referenceLocator": "pkg:golang/github.com/google/ko@(devel)?type=module",
          "referenceType": "purl"
        }
      ]
    },
    {
      "SPDXID": "SPDXRef-Package-github.com.google.go-containerregistry-v0.10.0",
      "name": "github.com/google/go-containerregistry",
      "versionInfo": "v0.10.0",
      "filesAnalyzed": false,
      "licenseDeclared": "NOASSERTION",
      "licenseConcluded": "NOASSERTION",
      "downloadLocation": "https://proxy.golang.org/github.com/google/go-containerregistry/@v/v0.10.0.zip",
      "copyrightText": "NOASSERTION",
      "checksums": [
        {
          "algorithm": "SHA256",
          "checksumValue": "a9dfdfbf69d06a319925e9da35c75a82197048f8d0d0da6137d87302baf65968"
        }
      ],
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE_MANAGER",
          "referenceLocator": "pkg:golang/github.com/google/go-containerregistry@v0.10.0?type=module",
          "referenceType": "purl"
        }
      ]
    }
  ],
  "relationships": [
    {
      "spdxElementId": "SPDXRef-DOCUMENT",
      "relationshipType": "DESCRIBES",
      "relatedSpdxElement": "SPDXRef-Package-sha256-8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85"
    },
    {
      "spdxElementId": "SPDXRef-Package-sha256-8955c18a6362f783536ec377fa0508646102895d89c08ee9d2f186c26e949d85",
      "relationshipType": "CONTAINS",
      "relatedSpdxElement": "SPDXRef-Package-github.com.google.ko-(devel)"
    },
    {
      "spdxElementId": "SPDXRef-Package-github.com.google.ko-(devel)",
      "relationshipType": "DEPENDS_ON",
      "relatedSpdxElement": "SPDXRef-Package-github.com.google.go-containerregistry-v0.10.0"
    }
  ]
}
```

mattmoor:
Here's an example of a multi-arch SBOM with this change:
```json
{
  "SPDXID": "SPDXRef-DOCUMENT",
  "name": "sbom-sha256:3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6",
  "spdxVersion": "SPDX-2.2",
  "creationInfo": {
    "created": "2022-07-01T02:27:39Z",
    "creators": [
      "Tool: ko (devel)"
    ]
  },
  "dataLicense": "CC0-1.0",
  "documentNamespace": "http://spdx.org/spdxdocs/kosha256:3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6",
  "documentDescribes": [
    "SPDXRef-Package-sha256-3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6"
  ],
  "packages": [
    {
      "SPDXID": "SPDXRef-Package-sha256-3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6",
      "name": "sha256:3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6",
      "versionInfo": "",
      "filesAnalyzed": false,
      "licenseDeclared": "NOASSERTION",
      "licenseConcluded": "NOASSERTION",
      "downloadLocation": "",
      "copyrightText": "NOASSERTION",
      "checksums": [
        {
          "algorithm": "SHA256",
          "checksumValue": "3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6"
        }
      ],
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE_MANAGER",
          "referenceLocator": "pkg:oci/index@sha256:3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6",
          "referenceType": "purl"
        }
      ]
    }
  ],
  "relationships": [
    {
      "spdxElementId": "DocumentRef-amd64-image-sbom:SPDXRef-Package-sha256-f03f678ab9cefa198a2334c4ea1e4d740c9b9686643db54b0053a80e6d5041e9",
      "relationshipType": "VARIANT_OF",
      "relatedSpdxElement": "SPDXRef-Package-sha256-3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6"
    },
    {
      "spdxElementId": "DocumentRef-arm64-image-sbom:SPDXRef-Package-sha256-569a8f91aed6b872d3e0c00f339b834ba671b2d04da80a14823f0db381890928",
      "relationshipType": "VARIANT_OF",
      "relatedSpdxElement": "SPDXRef-Package-sha256-3b72bb7f7ce12357a17d40f5198708823333790dd20d460613a9cdaabc97bdd6"
    }
  ],
  "externalDocumentRefs": [
    {
      "checksum": {
        "algorithm": "SHA256",
        "checksumValue": "c066df5c956def2fd36a340d4dcf67a51bbaf9e31201ae43088406b5ff99681a"
      },
      "externalDocumentId": "DocumentRef-amd64-image-sbom",
      "spdxDocument": "pkg:oci/sbom@sha256:f03f678ab9cefa198a2334c4ea1e4d740c9b9686643db54b0053a80e6d5041e9"
    },
    {
      "checksum": {
        "algorithm": "SHA256",
        "checksumValue": "03789b9d525f55260ea83ee2c173a54c40fda3f99110fd328d270e8a01e8b3e1"
      },
      "externalDocumentId": "DocumentRef-arm64-image-sbom",
      "spdxDocument": "pkg:oci/sbom@sha256:569a8f91aed6b872d3e0c00f339b834ba671b2d04da80a14823f0db381890928"
    }
  ]
}

```

puerco:
OK, its missing the external document refs, you need to merge this patch first:

https://github.com/google/ko/pull/741

mattmoor:
It's in there, I didn't see your PR until just now!
```json
  "externalDocumentRefs": [
    {
      "checksum": {
        "algorithm": "SHA256",
        "checksumValue": "c066df5c956def2fd36a340d4dcf67a51bbaf9e31201ae43088406b5ff99681a"
      },
      "externalDocumentId": "DocumentRef-amd64-image-sbom",
      "spdxDocument": "pkg:oci/sbom@sha256:f03f678ab9cefa198a2334c4ea1e4d740c9b9686643db54b0053a80e6d5041e9"
    },
    {
      "checksum": {
        "algorithm": "SHA256",
        "checksumValue": "03789b9d525f55260ea83ee2c173a54c40fda3f99110fd328d270e8a01e8b3e1"
      },
      "externalDocumentId": "DocumentRef-arm64-image-sbom",
      "spdxDocument": "pkg:oci/sbom@sha256:569a8f91aed6b872d3e0c00f339b834ba671b2d04da80a14823f0db381890928"
    }
  ]
```

puerco:
Ah sorry I ran it through the validator and it failed because it could not find the ref. I assumed it was not there. But I just saw that it's failing because the `spdxDocument` field has the purl in it. This prevents the validator from pulling the doc to complete the sbom data :(

puerco:
If we cannot reliably get a link to the SBOMs of the arch images, we can simply not reference the sboms, only reference the image as a package.

We can think about it semantically as having two separate SBOM as opposed to this approach which is more akin to having one single SBOM broken into different files. From our discussion today and what I was discussing with @jonjohnsonjr I think I'm leaning more towards defining them this way.

mattmoor:
Let me try getting the validator running against index SBOMs here.  I was being a bit lazy since it doesn't have a registry and just uses `ko deps`.

mattmoor:
oh it does 😈

mattmoor:
> From our discussion today and what I was discussing with @jonjohnsonjr I think I'm leaning more towards defining them this way

Just as packages?  Ok, yeah I think that makes sense, and then we have the same problem for index -> image as we do for image -> other stuff's sboms.

mattmoor:
Ok I switched it to using Packages, lmk if there’s a better ref type

mattmoor:
Ok, so I realized my first crack at packages lost the platform information, so I brought it back in the `version` field of the package: https://github.com/google/ko/runs/7163036130?check_suite_focus=true#step:7:130

mattmoor:
Here's a screenshot highlighting the digest of the image
image

mattmoor:
Another reason I just thought of for why we need a non-native SPDX way of linking to other SBOMs is that we might have a mix of SPDX (derived) and CycloneDX (base).  That's not the case here, but clearly we need to handle that for base images.

puerco:
The platform info can be encoded in the purl. But I'm looking and for some reason, it only has an `arch` field!?!?!   
https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci

We could propose another `os` field or simply cram the whole linux/arm/v8 field in there. I would prefer to have it there because  the purl is that key to link elements spread across sboms in spdx and cyclonedx

mattmoor:
Sure, just lmk what you'd like to see here and we can adjust!

mattmoor:
Added the architecture to the pURLs as well.

puerco:
I've opened https://github.com/package-url/purl-spec/pull/179 to propose adding a `os` qualifier to the purl spec. Its the missing bit to complete the platform information.

internal/sbom/mod_1.18.go
```diff
@@ -32,3 +33,29 @@ func ParseBuildInfo(data string) (*BuildInfo, error) {
 	bi := BuildInfo(*dbi)
 	return &bi, nil
 }
+
+func modulePackageName(mod *debug.Module) string {
```
imjasonh, line 37: Should these be in some shared file that both 1.18 and pre-1.18 code can reuse?

internal/sbom/spdx.go
```diff
@@ -125,6 +170,151 @@ func GenerateSPDX(koVersion string, date time.Time, mod []byte, imgDigest v1.Has
 	return buf.Bytes(), nil
 }
 
+func extractDate(sii oci.SignedImageIndex) (*time.Time, error) {
```
imjasonh, line 173: Instead of extracting the date from the index we just built, can we pass the created time from gobuild.go into where we generate the SBOMs?

Or is this just to support `ko deps`?

pkg/build/gobuild_test.go
```diff
@@ -399,7 +399,7 @@ func nilGetBase(_ context.Context, _ string) (name.Reference, Result, error) {
 const wantSBOM = "This is our fake SBOM"
 
 // A helper method we use to substitute for the default "build" method.
-func fauxSBOM(_ context.Context, _ string, _ string, _ v1.Image) ([]byte, types.MediaType, error) {
+func fauxSBOM(_ context.Context, _ string, _ string, _ oci.SignedEntity) ([]byte, types.MediaType, error) {
```
imjasonh, line 402:
```suggestion
func fauxSBOM(context.Context, string, string, oci.SignedEntity) ([]byte, types.MediaType, error) {
```

pkg/commands/deps.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/google/ko/internal/sbom"
+	"github.com/sigstore/cosign/pkg/oci/signed"
```
imjasonh, line 33: Is this `signed` package new? Why is there `oci.SignedImage` and `oci/signed.Image`?

mattmoor: No, it is a convenient way of turning a `v1.Image` into an `oic.SignedImage` with empty signatures et al

pkg/build/gobuild_test.go
mattmoor: done

internal/sbom/spdx.go
mattmoor: That's what the image SBOM logic was doing, but extracting it from the image's config, which is frankly more accessible.  I'll take a quick look at how hard it would be to pass it in without diverging the source of truth logic for setting these timestamps.

mattmoor: Looks doable, lemme plumb that through instead. 👍

mattmoor: Done!

mattmoor: Ugh, so it looks like we only use `g.creationTime` under certain circumstances, and we let the base image's creation time through otherwise, so for the index SBOM I think we'll end up needing some sort of extraction method to make that really work.

mattmoor: e.g. 
```
	empty := v1.Time{}
	if g.creationTime != empty {
		cfg.Created = g.creationTime
	}
```

mattmoor: I'm reverting to what I had, as I think it's cleanest modulo a bigger refactor of this code.

internal/sbom/mod_1.18.go
imjasonh: Boooo I guess what you had was fine. I hate this pre-1.18 and count down the days until we don't have to support it.

mattmoor: Yeah, I had no idea until CI blew up on me, and I had to contort things to be able to leverage module types.

Will be good to kill this with 🔥

PR #744, opened 2022-07-03
Add base image information to SPDX SBOMs

mattmoor:
WIP until https://github.com/google/ko/pull/743 lands.

I'm going to split off the annotation change as well, but including here so that things work.

PR #745, opened 2022-07-03
Unconditionally set the base image annotation.

mattmoor:
Previously we only set this annotation when our base image was a tag, but this means when folks actually follow the best practice of use digest base images they get strictly worse resulting images!

It sounds like maybe the original motivation for this condition was that it was supposed to contain the mutable reference (tag), but I don't see anything detailing such a restriction (just an example), so there should be nothing
precluding this from the spec.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -889,8 +889,8 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 
 		anns := map[string]string{
 			specsv1.AnnotationBaseImageDigest: baseDigest.String(),
+			specsv1.AnnotationBaseImageName:   baseRef.Name(),
 		}
-		anns[specsv1.AnnotationBaseImageName] = baseRef.Name()
 		base = mutate.Annotations(base, anns).(Result)
 	}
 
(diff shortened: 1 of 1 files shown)


mattmoor:
This also supports: https://github.com/google/ko/pull/744

pkg/build/gobuild.go
```diff
@@ -890,9 +890,7 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 		anns := map[string]string{
 			specsv1.AnnotationBaseImageDigest: baseDigest.String(),
 		}
-		if _, ok := baseRef.(name.Tag); ok {
-			anns[specsv1.AnnotationBaseImageName] = baseRef.Name()
-		}
+		anns[specsv1.AnnotationBaseImageName] = baseRef.Name()
```
imjasonh, line 893: The nittiest of nits: can you just put this in the map initialization expression above, with the other one?

imjasonh (changes requested):

pkg/build/gobuild.go
mattmoor: Done!

PR #749, opened 2022-07-06
Move docs to ko.build

imjasonh:
This copies content from README.md into separate docs in `docs/`, and uses [`mkdocs-material`](https://squidfunk.github.io/mkdocs-material/) to generate a static site that's served from [GitHub Pages](https://pages.github.com/), in the `gh-pages` branch.

A preview of this site served from https://github.com/imjasonh/ko.build is available at https://ko.build -- when this merges and is published I'll update the domain to point to this repo's site and archive that one. I'll also update our README.md to link to ko.build instead of duplicating content.

This also moves generated CLI docs from `doc/` to `docs/reference/`, so we don't have to keep two copies in sync.


imjasonh:
Flipping this to Draft since we'll need a google GH org admin to set up the domain. This shouldn't block the 0.12 release.

imjasonh:
I'd like to get this merged so we can start running ko.build out of this repo instead of out of my personal repo.

This change also sets us up for the `ko.build` importpath, which will let us change the module name in 0.13 (e.g., `go install ko.build`, `import "ko.build/pkg/build"`, etc.

jonjohnsonjr (approved):

PR #757, opened 2022-07-07
Set -extldflags="-static-pie" by default

### LINKED ISSUE #756: Build with -extldflags="-static" by default
> -extldflags flags
	Set space-separated flags to pass to the external linker.

from https://pkg.go.dev/cmd/link

> -static
> Do not link against shared libraries. This is only meaningful on platforms for which shared libraries are supported. The different variants of this option are for compatibility with various systems. You may use this option multiple times on the command line: it affects library searching for -l options which follow it. This option also implies --unresolved-symbols=report-all. This option can be used with -shared. Doing so means that a shared library is being created but that all of the library's external references must be resolved by pulling in entries from static libraries.

from https://linux.die.net/man/1/ld via https://stackoverflow.com/a/49600956

We should make sure this cooperates with any other `-extldflags` the user may have specified.

imjasonh:
Fixes #756

If a user specifies these flags, their flags take precedence, by virtue of coming later in the list.

cc @kaniini

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -693,6 +693,13 @@ func createBuildArgs(buildCfg Config) ([]string, error) {
 
 func (g *gobuild) configForImportPath(ip string) Config {
 	config := g.buildConfigs[ip]
+
+	// Set -extldflags by default, but set it first so that it can be overridden.
+	config.Ldflags = append([]string{"-extldflags", `"-static-pie"`}, config.Ldflags...)
+
+	// TODO: Set -buildmode=pie (https://github.com/google/ko/issues/375)
+	// config.Flags = append([]string{"-buildmode=pie"}, config.Flags...)
+
 	if g.trimpath {
 		// The `-trimpath` flag removes file system paths from the resulting binary, to aid reproducibility.
 		// Ref: https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies
--- a/pkg/build/gobuild_test.go
+++ b/pkg/build/gobuild_test.go
@@ -301,79 +301,97 @@ func TestBuildEnv(t *testing.T) {
 }
 
 func TestBuildConfig(t *testing.T) {
-	tests := []struct {
+	for _, test := range []struct {
 		description  string
 		options      []Option
 		importpath   string
 		expectConfig Config
-	}{
-		{
-			description: "minimal options",
-			options: []Option{
-				WithBaseImages(nilGetBase),
-			},
+	}{{
+		description: "minimal options",
+		options: []Option{
+			WithBaseImages(nilGetBase),
 		},
-		{
-			description: "trimpath flag",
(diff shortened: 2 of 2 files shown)


imjasonh:
So with `-buildmode=pie`, running the built image on an M1 mac gives me either:

```
$ docker run --rm $(go run ./ build ./test/) --wait=false
...
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
```

or 

```
$ docker run --rm $(go run ./ build ./test/ --platform=linux/arm64) --wait=false
exec /ko-app/test: no such file or directory
```

Both `-extldflags "-static"` and `-extldflags "-static-pie"` seem to work fine, I might just go with that for now and figure out `-buildmode=pie` later. 🥧

kaniini:
We should verify that the built binary is indeed static PIE before this is merged imo.  You can do it by using file(1) on the built go binary, it should say it is a static PIE binary.

imjasonh:
Reading around it sounds like `-extldflags="-static"` (or `-static-pie`) only really matter when you end up depending on cgo, which we don't by default anyway. I wonder if we can just not set these flags, and remove them as unnecessary when they are used.

Maybe we could even warn if we see suspect ldflags when the user hasn't explicitly enabled `CGO_ENABLED=1`, since it may just lead to more confusion.

kaniini:
hmm, I'm not sure why it would only matter in cgo case -- static PIE means that all of the addresses of things get randomized on every program load, rather than at build time, which pure-go binaries should still benefit from?

imjasonh:
> hmm, I'm not sure why it would only matter in cgo case -- static PIE means that all of the addresses of things get randomized on every program load, rather than at build time, which pure-go binaries should still benefit from?

I think there's still a benefit to setting `-buildmode=pie`, it just fails to run when you build it that way for some unknown reason (https://github.com/google/ko/pull/757#issuecomment-1178987855), so I want to try that later.

My point was I'm not sure that we need `-static` or `-static-pie` if we're already defaulting to `CGO_ENABLED=0`, where these flags (AFAIK) don't really do anything useful, since there's no cgo to statically compile. (I'm only like 30% sure of anything anymore.)

kaniini:
AFAIK go links against libc dynamically without static PIE.

imjasonh:
https://www.arp242.net/static-go.html and [this rando's Reddit comment](https://www.reddit.com/r/golang/comments/8m4xrh/do_linux_golang_binaries_depend_on_libc/dzktybb/) both make it seem like `CGO_ENABLED=0` is the magic talisman against depending on libc at all. Otherwise, net and os/user sneak in libc deps and the binary is dynamically linked.

How should I test that a `CGO_ENABLED=0 go build`-built binary is already static for whatever definition we care about?

PR #759, opened 2022-07-10
Decorate per-architecture images with base image annotations.

mattmoor:
Currently we only decorate the topmost index/image with base image data for the base index/image (for everything except docker manifest lists).

This change makes multi-arch builds decorate the manifest of the per-architecture images with the reference of the base index, and the digest of the specific base image used for that particular architecture's image.

This results in the per-architecture SBOMs starting to encode `DESCENDENT_OF` links as well.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -925,7 +925,7 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 		if !ok {
 			return nil, fmt.Errorf("failed to interpret base as index: %v", base)
 		}
-		res, err = g.buildAll(ctx, s, baseIndex)
+		res, err = g.buildAll(ctx, s, baseRef, baseIndex)
 	case types.OCIManifestSchema1, types.DockerManifestSchema2:
 		baseImage, ok := base.(v1.Image)
 		if !ok {
@@ -941,7 +941,7 @@ func (g *gobuild) Build(ctx context.Context, s string) (Result, error) {
 	return res, nil
 }
 
-func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIndex) (Result, error) {
+func (g *gobuild) buildAll(ctx context.Context, ref string, baseRef name.Reference, baseIndex v1.ImageIndex) (Result, error) {
 	im, err := baseIndex.IndexManifest()
 	if err != nil {
 		return nil, err
@@ -968,8 +968,12 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIn
 		if err != nil {
 			return nil, fmt.Errorf("error getting matching image from index: %w", err)
 		}
-		// Carry forward the base index's annotations, which include base image annotations.
-		img = mutate.Annotations(img, im.Annotations).(v1.Image)
+		// Decorate the image with the ref of the index, and the matching
+		// platform's digest.
+		img = mutate.Annotations(img, map[string]string{
+			specsv1.AnnotationBaseImageDigest: matches[0].Digest.String(),
+			specsv1.AnnotationBaseImageName:   baseRef.Name(),
+		}).(v1.Image)
 		return g.buildOne(ctx, ref, img, matches[0].Platform)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/build/gobuild.go
```diff
@@ -968,8 +968,12 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIn
 		if err != nil {
 			return nil, fmt.Errorf("error getting matching image from index: %w", err)
 		}
-		// Carry forward the base index's annotations, which include base image annotations.
```
imjasonh, line 971: I think I'm okay with it, but just to make sure, this change also has the result of _not_ carrying over any base image annotations, and only settings these new annotations.

I think that's probably fine, I just want to make sure this was a conscious decision.

pkg/build/gobuild.go
```diff
@@ -985,6 +989,14 @@ func (g *gobuild) buildAll(ctx context.Context, ref string, baseIndex v1.ImageIn
 			if err != nil {
 				return err
 			}
+
+			// Decorate the image with the ref of the index, and the matching
+			// platform's digest.
+			baseImage = mutate.Annotations(baseImage, map[string]string{
+				specsv1.AnnotationBaseImageDigest: desc.Digest.String(),
+				specsv1.AnnotationBaseImageName:   baseRef.Name(),
```
imjasonh, line 997: So this feels weird. The platform-specific image's annotations will point to the platform-specific base image's digest (good), and containing the _index's_ ref?

I think I would have expected this to be the platform-specific image's ref(-by-digest, like we get when we base an image on an image-by-digest), rather than the two annotations effectively pointing to two different things.

Are you doing this so the platform-specific image's SBOM can point up to the containing index? That seems like information we can inject into the SBOM without having to pass it through these annotations.

pkg/build/gobuild.go
mattmoor: I got the impression that the main use case of this was to get the base image annotation carried forward.  Without another use case in mind it felt like apples/oranges to me, where blindly copying context could lead to confusion.  I haven't had enough ☕ to produce contrived examples yet

imjasonh: Yeah, having had two ☕ s so far I'm only becoming more convinced that blindly carrying forward the annotations is wrong. These may include things like source repo or description, that don't make sense for derived images.

pkg/build/gobuild.go
mattmoor: I agree it is somewhat weird.  Beyond the digest being accurate (and the ref present) there's no real agenda here.

Preserving `name` felt like a way to leave sufficient breadcrumbs for the use case @jonjohnsonjr had in mind (monitoring the actual `FROM` for updates) while `digest` provides the most accurate variant of the image we actually used as our basis.

I could make these `repo@digest` vis `baseRef.Context().Digest(desc.Digest.String())` or just the `repo` via `baseRef.Context().String()`.  Ultimately as long as `repo` is there, the additional information feels somewhat extraneous, which is why I biased towards enabling @jonjohnsonjr use case (albeit not via simple `HEAD`)

pkg/build/gobuild.go
mattmoor: Great examples, and not even contrived! 😂

pkg/build/gobuild.go
imjasonh: I see. You just need `image.name` to include the repo, Jon wants a mutable ref so he can HEAD to see if a rebase is needed, and I (seem to) want repo@digest so that the `image.digest` and `image.name` are both pointing to the same thing.

Thinking this through more I think I'm coming around to Jon's standpoint, even though it is weird. If a multiarch image changes but a platform-specific image within it doesn't*, then the rebase detection logic will have to do more than a HEAD to tell (maybe a GET and a HEAD, or two HEADs).

*this is somewhat relevant to how multiarch base images are updated on Dockerhub -- when `alpine:latest` is updated e.g., when `3.17` is released, a build farm kicks of images for each arch, and the index is updated in place as builds complete. So you might see `alpine:latest` update to point to `alpine[linux/amd64]:3.16` and `alpine[linux/arm64]:3.17`, then update again so they both include `:3.17`. If there are N platforms in the manifest list, it may get updated N times. If each of those updates kick off a rebase flow that is a noop (N-1)/N times, that's kinda wasteful. But I'm not really sure how else to address it and retain the workflow we wanted.

I didn't realize this was how things worked when we talked through the annotations forever ago. 😢

imjasonh: I'm happy to resolve/unblock, but maybe add a comment to save me a 🤨 at this in a few months.

mattmoor: > If each of those updates kick off a rebase flow that is a noop (N-1)/N times, that's kinda wasteful

Right, this was sort of my (poorly articulated) point.  The initial check may be more expensive than a HEAD, but as-is you fundamentally lack the information to judge whether it is a NOP or not without doing the work (we don't know the original digest of the base *image*, only that the index has changed).  For rebasing, the work to be redone is pretty light, but rebasing is kinda gross and if someone were actually using this to drive rebuilds, they would be in for a pretty bad time.

mattmoor:
Gonna rebase this on `main` now that my other change landed, so I can get it all in a local install.

PR #760, opened 2022-07-11
Several SPDX SBOM adjustments.

mattmoor:
1. Change index -> image relationship type from `CONTAINS` to `VARIANT_OF` (I think this was an oversight in my original PR),
2. Always include `mediaType` in pURLs for index/images we produce (I'm not adding this to the base image, since it's not readily available, but we can add it there if we want to find a way to plumb it through),
3. Include more platform discriminator information to the pURLs we use in index -> image.

cc @puerco

### DIFF
--- a/internal/sbom/spdx.go
+++ b/internal/sbom/spdx.go
@@ -103,7 +103,10 @@ func GenerateImageSPDX(koVersion string, mod []byte, img oci.SignedImage) ([]byt
 		ExternalRefs: []ExternalRef{{
 			Category: "PACKAGE_MANAGER",
 			Type:     "purl",
-			Locator:  ociRef("image", imgDigest),
+			Locator: ociRef("image", imgDigest, qualifier{
+				key:   "mediaType",
+				value: string(m.MediaType),
+			}),
 		}},
 	})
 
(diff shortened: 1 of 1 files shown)


internal/sbom/spdx.go
```diff
@@ -263,10 +269,39 @@ func GenerateIndexSPDX(koVersion string, sii oci.SignedImageIndex) ([]byte, erro
 
 			doc.Relationships = append(doc.Relationships, Relationship{
 				Element: ociPackageName(indexDigest),
-				Type:    "CONTAINS",
+				Type:    "VARIANT_OF",
 				Related: depID,
 			})
 
+			qual := []qualifier{{
```
imjasonh, line 276: This makes me want a `newQualifier(k, v string) qualifier` helper method, and/or a `platformQualifiers(p v1.Platform) []qualifier`.

imjasonh (approved):

internal/sbom/spdx.go
mattmoor: Yeah, it's worth surveying our usage of qualifiers as this evolves and figuring out what helpers make sense.  I'd also be inclined to just create a decent OCI pURL library in GGCR and use that TBH.

Issue #769, opened 2022-07-19
RFC: deprecate `ko run` and `ko deps`?


imjasonh:
(or demote to a `beta` CLI surface?)

Among the CLI commands in `ko`, most are pretty well-exercised, at least by myself and folks I talk to. There are two exceptions:

### `ko run`

`ko run` builds an image and [`kubectl run`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#run)s it. `kubectl run` itself is not a terribly common command, and while it's convenient for quickly running some code in a cluster and seeing logs, there are some notable limitations -- you can't set node selectors, you can't run multiple containers (init containers, sidecars, etc.), and more, and AFAIK it's not really scheduled to get many enhancements. It's useful for testing, but not much else.

`ko run` is not well tested end-to-end at all, which has led to it being silently broken for quite a while, and only recently fixed.

`ko run` was [added all the way back in 2018](https://github.com/google/go-containerregistry/pull/331), back when `ko` shared a home with `go-containerregistry` (which maybe explains why it didn't get much testing love). In the ensuing 4 years, `ko` has grown a lot, and `ko run` really hasn't.

I think we should deprecate `ko run` or invest in better tests for it.

### `ko deps`

`ko deps` was [added in v0.9.X, in September last year](https://github.com/google/ko/releases/tag/v0.9.1), to pull and extract build info from a `ko`-built image and output an SBOM. It was useful work, which formed the foundation for our built-in SBOM generation work. Instead of running `ko deps` on a built image, SBOMs are generated and pushed when images are built.

The use case of needing to generate an SBOM on-demand from an already-pushed image and to just print it out, is pretty niche, especially when the default behavior of `ko` is to push that same SBOM to the registry, where you can get it with `cosign download sbom` or other tools (e.g., policy-controller).

I think we should deprecate `ko deps`, having served its purpose.

---

I'd like to make a decision on these one way or another before any future v1.0 release of `ko`.

cc @jonjohnsonjr @mattmoor

mattmoor:
`{insert Kermit sith "do it"}`

imjasonh:
This question is now phrased in the form of a PR: https://github.com/google/ko/pull/770

imjasonh:
@jonjohnsonjr says he still uses and likes `ko run`, so I'll punt its deprecation for a later date. It sounds like we have consensus about `ko deps` though, so I've updated #770 to only deprecate that. PTAL, I'd like to get it in before v0.12.

PR #770, opened 2022-07-19
Deprecate ko deps

imjasonh:
Fixes https://github.com/google/ko/issues/769


mattmoor (approved):

imjasonh:
It seems that `Deprecated` in cobra means the command also gets hidden from `ko help` and generated Markdown docs. If that's not something we want, I can try to figure out how to make that not happen.

jonjohnsonjr (approved):

PR #785, opened 2022-08-04
Error if image platform does not match desired

### LINKED ISSUE #699: --platform can be ignored by what the base image supports
If you have a `defaultBaseImage` that is an image manifest rather than a image index, it seems that ko will just build for that image's platform, happily ignoring whatever you pass via `--platforms`.

Repro here: https://github.com/benmoss/wtf-ko
```
❯ ko build --platform=all .
2022/05/05 10:57:43 Using base gcr.io/distroless/static@sha256:abb120b4ebb4e734dc4c82181bb32b5a4f8bcf78e4540b094b9d59ba7a29a5ad for github.com/benmoss/wtf-ko
2022/05/05 10:57:43 Building github.com/benmoss/wtf-ko for linux/amd64
2022/05/05 10:57:44 Loading kind.local/wtf-ko-3d3e10e27ff19c6d1d54aaa50fb62fd9:843abadc2e51b54f34fe7c0bee50f99c4462593efc183b6867b72573aa0bf572
2022/05/05 10:57:44 Loaded kind.local/wtf-ko-3d3e10e27ff19c6d1d54aaa50fb62fd9:843abadc2e51b54f34fe7c0bee50f99c4462593efc183b6867b72573aa0bf572
2022/05/05 10:57:44 Adding tag latest
2022/05/05 10:57:44 Added tag latest
kind.local/wtf-ko-3d3e10e27ff19c6d1d54aaa50fb62fd9:843abadc2e51b54f34fe7c0bee50f99c4462593efc183b6867b72573aa0bf572
```

benmoss:
Fixes #699

If the user specifies an image manifest and its platform does not match the platforms specified or defaulted, error out.

With a `.ko.yaml` specifying the linux/amd64 image manifest instead of the multi-platform index:
```yaml
baseImageOverrides:
  github.com/google/ko: golang@sha256:a4081692fa3015104d84b125cbf623c566a0c9e2a39b9a29f6d939225d5687a4
```

## Prior behavior
```
❯ go run main.go build --platform=linux/arm64 .
2022/08/04 10:36:41 No matching credentials were found, falling back on anonymous
2022/08/04 10:36:42 Using base golang@sha256:a4081692fa3015104d84b125cbf623c566a0c9e2a39b9a29f6d939225d5687a4 for github.com/google/ko
2022/08/04 10:36:42 Building github.com/google/ko for linux/amd64
...
```

## New behavior
```
❯ go run main.go build --platform=linux/arm64 .
2022/08/04 10:43:52 No matching credentials were found, falling back on anonymous
2022/08/04 10:43:53 Using base golang@sha256:a4081692fa3015104d84b125cbf623c566a0c9e2a39b9a29f6d939225d5687a4 for github.com/google/ko
Error: failed to publish images: error building "ko://github.com/google/ko": base image platform "linux/amd64" does not match desired platforms [linux/arm64]
2022/08/04 10:43:53 error during command execution:failed to publish images: error building "ko://github.com/google/ko": base image platform "linux/amd64" does not match desired platforms [linux/arm64]
exit status 1
```

As mentioned in #699 `--platform=all` is a lil ambiguous so we're keeping that behavior:
```
❯ go run main.go build --platform=all .
2022/08/04 10:45:40 No matching credentials were found, falling back on anonymous
2022/08/04 10:45:41 Using base golang@sha256:a4081692fa3015104d84b125cbf623c566a0c9e2a39b9a29f6d939225d5687a4 for github.com/google/ko
2022/08/04 10:45:41 Building github.com/google/ko for linux/amd64
...
```

### DIFF
--- a/pkg/build/gobuild_test.go
+++ b/pkg/build/gobuild_test.go
@@ -437,6 +437,7 @@ func TestGoBuildNoKoData(t *testing.T) {
 		WithBaseImages(func(context.Context, string) (name.Reference, Result, error) { return baseRef, base, nil }),
 		withBuilder(writeTempFile),
 		withSBOMber(fauxSBOM),
+		WithPlatforms("all"),
 	)
 	if err != nil {
 		t.Fatalf("NewGo() = %v", err)
@@ -719,6 +720,7 @@ func TestGoBuild(t *testing.T) {
 		withSBOMber(fauxSBOM),
 		WithLabel("foo", "bar"),
 		WithLabel("hello", "world"),
+		WithPlatforms("all"),
 	)
 	if err != nil {
 		t.Fatalf("NewGo() = %v", err)
@@ -794,6 +796,7 @@ func TestGoBuildWithKOCACHE(t *testing.T) {
 		"",
 		WithCreationTime(creationTime),
 		WithBaseImages(func(context.Context, string) (name.Reference, Result, error) { return baseRef, base, nil }),
+		WithPlatforms("all"),
 	)
 	if err != nil {
 		t.Fatalf("NewGo() = %v", err)
@@ -831,6 +834,7 @@ func TestGoBuildWithoutSBOM(t *testing.T) {
 		WithLabel("foo", "bar"),
 		WithLabel("hello", "world"),
 		WithDisabledSBOM(),
+		WithPlatforms("all"),
 	)
 	if err != nil {
 		t.Fatalf("NewGo() = %v", err)
@@ -1183,6 +1187,7 @@ func TestGoBuildConsistentMediaTypes(t *testing.T) {
 				WithBaseImages(func(context.Context, string) (name.Reference, Result, error) { return baseRef, base, nil }),
 				withBuilder(writeTempFile),
 				withSBOMber(fauxSBOM),
+				WithPlatforms("all"),
 			)
 			if err != nil {
 				t.Fatalf("NewGo() = %v", err)
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -757,6 +757,9 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 		}
 	}
 
+	if !g.platformMatcher.matches(platform) {
+		return nil, fmt.Errorf("base image platform %q does not match desired platforms %v", platform, g.platformMatcher.platforms)
+	}
 	// Do the build into a temporary file.
 	file, err := g.build(ctx, ref.Path(), g.dir, *platform, g.configForImportPath(ref.Path()))
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/build/gobuild_test.go
```diff
@@ -437,6 +437,7 @@ func TestGoBuildNoKoData(t *testing.T) {
 		WithBaseImages(func(context.Context, string) (name.Reference, Result, error) { return baseRef, base, nil }),
 		withBuilder(writeTempFile),
 		withSBOMber(fauxSBOM),
+		WithPlatforms("all"),
```
imjasonh, line 440: I'm confused a bit why this addition is needed for tests. I would have expected `random.Image` to produce a `linux/amd64` image by default, and for `NewGo` to assume that platform if none is provided, so that platform matcher check you added wouldn't be necessary.

Does this change require callers of the Go API to explicitly pass `WithPlatforms`?

imjasonh (commented):
This mostly looks good, thanks for the contribution!

If there was an additional test in the e2e tests that checked that `ko build --platform=linux/fake` resulted in an error, I'd award extra bonus points 😁🏅

pkg/build/gobuild_test.go
benmoss: `random.Image` has no platform (or any config at all), so you get this somewhat odd error

```
gobuild_test.go:448: Build() = base image platform "" does not match desired platforms []
```

It looks like from `empty.Image` it just gets an empty Platform struct:
```go
*v1.Platform{Architecture: "", OS: "", OSVersion: "", OSFeatures: []string len: 0, cap: 0, nil, Variant: "", Features: []string len: 0, cap: 0, nil}
```

The CLI calls [gobuildOptions](https://github.com/benmoss/ko/blob/8211102f214d6f024797db637bdeead2e871aa28/pkg/commands/resolver.go#L50) which handles the defaulting of desired platforms.

My understanding from reading over the code is that the Go API (`NewGo`) doesn't do any defaulting, so if the user didn't set any desired platforms the current behavior would be:
- if given an image index/manifest list, [no platforms would match](https://github.com/benmoss/ko/blob/8211102f214d6f024797db637bdeead2e871aa28/pkg/build/gobuild.go#L985)
- if given an image manifest, it would [copy the platform](https://github.com/benmoss/ko/blob/8211102f214d6f024797db637bdeead2e871aa28/pkg/build/gobuild.go#L752-L758) from the base image

The tests were relying on that latter behavior, so the resulting image would also have a blank platform.

benmoss:
> If there was an additional test in the e2e tests that checked that `ko build --platform=linux/fake` resulted in an error, I'd award extra bonus points 😁🏅

We'd have to set `KO_DEFAULTBASEIMAGE` as well, since it defaults to an image index (go:1.18), and that already handles erroring for unsupported platforms. This is behavior that only triggers when you use image manifests as your base image.

imjasonh (approved):

PR #797, opened 2022-08-22
Support --tag and --tag-only with nop publisher

imjasonh:
Supercedes #689

cc @adambkaplan

With this change, if you use `--push=false` and `-t`, the output ref includes the tag. If you include `--tag-only` the resulting output ref will only include the tag, and _not_ the digest. I'm not sure what purpose that would be useful for, but it's consistent with pushing cases at least.

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -224,9 +224,16 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 		// If not publishing, at least generate a digest to simulate
 		// publishing.
 		if len(publishers) == 0 {
+			// If one or more tags are specified, use the first tag in the list
+			tag := "latest"
+			if len(po.Tags) >= 1 {
+				tag = po.Tags[0]
+			}
 			publishers = append(publishers, nopPublisher{
 				repoName: repoName,
 				namer:    namer,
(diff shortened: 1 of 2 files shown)


pkg/commands/resolver.go
```diff
@@ -224,9 +224,16 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 		// If not publishing, at least generate a digest to simulate
 		// publishing.
 		if len(publishers) == 0 {
+			// If one or more tags are specified, use the first tag in the list
+			tag := "latest"
```
jonjohnsonjr, line 228: This seems like a behavior change?

imjasonh: `po.Tags` is defaulted to `[]string{"latest"}` if not set, so this isn't exactly a behavior change, but it does subtly depend on logic elsewhere which might change and break assumptions.

Fixed!

jonjohnsonjr (approved):

adambkaplan:
@imjasonh thanks for getting this over the finish line!

imjasonh:
> @imjasonh thanks for getting this over the finish line!

Thanks for noticing the issue and for your very helpful PR!

PR #801, opened 2022-08-26
SPDX: Fix package manager label

puerco:
This commit fixes the package manager label in external references
to make them [conform to the SPDX spec](https://spdx.github.io/spdx-spec/package-information/#721-external-reference-field).

/cc @imjasonh 

Signed-off-by: Adolfo Garcia Veytia (puerco) <puerco@chainguard.dev>

### DIFF
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: ['main']
 
+env:
+  SPDX_TOOLS_VERSION: 1.1.0
+
 jobs:
   go-version-m:
     name: Generate go version -m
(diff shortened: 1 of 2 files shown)


imjasonh (approved):

imjasonh:
Thanks for this! Do you think this is something that warrants a v0.12.1 release?

puerco:
Mmh no, I think we should instead move to have ko@HEAD generate SPDX 2.3. This uncovered a bug in the SPDX tools, between 2.2.2 and 2.3 the external reference category label in the json schema was changed from using a dash to an underscore:

https://github.com/spdx/spdx-spec/blob/master/schemas/spdx-schema.json#L325

https://github.com/spdx/spdx-spec/blob/development/v2.2.2/schemas/spdx-schema.json#L320

But it seems the spdx tools will use the latest json schema published regardless of the version stated in the doc. So right now, we have an invalid 2.2 document that passes the tests. I'll modify the code to generate SPDX 2.3 documents and file a bug on the SPDX tools project

puerco:
Ref: https://github.com/spdx/tools-java/issues/64

imjasonh:
Thanks @puerco! ❤️

antoineco:
I was looking at the changelog and noticed that the original link is dead, so here is the current version: https://spdx.github.io/spdx-spec/v2.3/package-information/#7211-description

PR #810, opened 2022-09-03
feat: wasi platform build

hown3d:
Second iteration, for more information see #738 

This PR implements the scenario described in https://github.com/google/ko/issues/675 by using tinygo as the builder for target WASI.

Documentation is not yet done, but can be created if this feature is wanted.

SBOM generation is currently not implemented, since tinygo doesn't produce the same buildInfo that go does.

Signed-off-by: Höhl, Lukas [lukas.hoehl@accso.de](mailto:lukas.hoehl@accso.de)


pkg/build/gobuild.go
```diff
@@ -241,8 +239,51 @@ func getGoarm(platform v1.Platform) (string, error) {
...
-	buildArgs, err := createBuildArgs(config)
+type executableBuilder interface {
+	BuildExecutable(ctx context.Context, importPath string, config Config) (exectuablePath string, err error)
+}
+
+type tinygoExecutableBuilder struct {
+	dir string
+}
+
+func (b *tinygoExecutableBuilder) BuildExecutable(ctx context.Context, ip string, config Config) (string, error) {
```
imjasonh, line 250: gobuild.go is getting _huge_ (which is my fault, not yours). Could we move these wasi-specific things into a new pkg/build/wasi.go?

internal/ociconv/ociconv.go
```diff
@@ -0,0 +1,150 @@
...
+	base = mutate.Annotations(base, m.Annotations).(v1.Image)
+	base, err = mutate.ConfigFile(base, cfg)
+	if err != nil {
+		return nil, err
+	}
+	return base, nil
+}
+
+// OCIImageIndex mutates the provided v1.ImageIndex to be OCI compliant v1.ImageIndex
+func OCIImageIndex(base v1.ImageIndex) (v1.ImageIndex, error) {
```
imjasonh, line 112: Does the image have to be OCI-ified as part of this? If not, I think we should punt on OCI-fication methods until a future change.

If it is required, I might still want to add it in a later PR so we can focus on getting it right. I've tried and failed to do this before, and the way this manifests is weird registry-specific failures that aren't easy to test here in isolation.

.github/workflows/wasi.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  wasi:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - ubuntu-latest
+    name: wasi
+    runs-on: ${{ matrix.platform }}
```
imjasonh, line 15: If there's only one platform, I'm inclined to just do `runs-on: ubuntu-latest`.

.github/workflows/wasi.yaml
```diff
@@ -0,0 +1,66 @@
+name: Integration test for wasi
```
imjasonh, line 1: ❤️❤️❤️

.github/workflows/wasi.yaml
```diff
@@ -0,0 +1,66 @@
...
+        shell: bash
+        run: |
+          set -euxo pipefail
+          export KO_DOCKER_REPO=$(uuidgen |tr "[:upper:]" "[:lower:]")
+          # Build and run the test/wasi binary, which should print "hello from wasi"
+          testimg=$(go run ./ build ./test/wasi/ --platform="wasm/wasi")
+  
+          # run a nested podman container to use annotations (could be removed once docker is able to specify annotations on containers)
+          docker run --priviliged \
+            -v /tmp/crun/crun:/usr/bin/crun \
```
imjasonh, line 63: Is there a way to just run crun directly, instead of having to run it in a container?

imjasonh (commented):
Thanks for taking another swing at this! This looks really promising, and much less involved than the last one.

I'm excited to see where this goes! 👀

internal/ociconv/ociconv.go
hown3d: I think this is a sort of relict from my previous tests. I got the source code for the OCI conversion stuff from this PR https://github.com/google/go-containerregistry/pull/1293. I added the index conversion as well, because I was not sure how a DockerManifestV2 Index List handles OCI Images, but it seems like to work.

I'll remove the unnecessary functions.

pkg/build/gobuild.go
hown3d: Done.

hown3d:
@imjasonh I did some local testing on the integration tests, they should be working by now.
Could you kick them off to see if it works as expected?

PR #815, opened 2022-09-06
mchmarny after migration cleanup

mchmarny:
Updates the repo path in all the non-go files (release, test, readme etc)

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -13,7 +13,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-  - "-s -w -X github.com/google/ko/pkg/commands.Version={{.Version}}"
+  - "-s -w -X github.com/ko-build/ko/pkg/commands.Version={{.Version}}"
   goos:
   - windows
   - linux
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,5 +1,5 @@
 baseImageOverrides:
-  github.com/google/ko: golang:1.18
+  github.com/ko-build/ko: golang:1.18
 
 builds:
 - id: ko
--- a/README.md
+++ b/README.md
@@ -1,7 +1,5 @@
 # `ko`: Easy Go Containers
 
-> :warning: Note: the ko project is growing, and to better facilitate community needs, we will be moving to our own GitHub organization: [ko-build](https://github.com/ko-build). While we do not anticipate this move to have any impact on most users, please, use the [ko-project Slack channel](https://kubernetes.slack.com/archives/C01T7DTP65S) to let us know if you see any impact we may have not yet considered. The project migration is currently scheduled for Tue, Sep 6th.
-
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -70,7 +68,7 @@ apk add ko
 With Go 1.18+, build and install the latest released version:
 
 ```
-go install github.com/google/ko@latest
+go install github.com/ko-build/ko@latest
```
imjasonh, line 71: This one will actually have to wait until we rename the Go import path. Even though the repo name has changed, as long as the GH redirect is in place, Go considers this `github.com/google/ko`.

.ko.yaml
```diff
@@ -1,5 +1,5 @@
 baseImageOverrides:
-  github.com/google/ko: golang:1.18
+  github.com/ko-build/ko: golang:1.18
```
imjasonh, line 2:
```suggestion
  github.com/google/ko: golang:1.18
```

.goreleaser.yml
```diff
@@ -13,7 +13,7 @@ builds:
   flags:
   - -trimpath
   ldflags:
-  - "-s -w -X github.com/google/ko/pkg/commands.Version={{.Version}}"
+  - "-s -w -X github.com/ko-build/ko/pkg/commands.Version={{.Version}}"
```
imjasonh, line 16:
```suggestion
  - "-s -w -X github.com/google/ko/pkg/commands.Version={{.Version}}"
```

integration_test.sh
```diff
@@ -17,13 +17,13 @@ export GOPATH
...
-mkdir -p "$GOPATH/src/github.com/google/ko"
-cp -r "$ROOT_DIR/"* "$GOPATH/src/github.com/google/ko/"
+mkdir -p "$GOPATH/src/github.com/ko-build/ko"
+cp -r "$ROOT_DIR/"* "$GOPATH/src/github.com/ko-build/ko/"
 
 echo "Downloading github.com/go-training/helloworld"
 GO111MODULE=off go get -d github.com/go-training/helloworld
 
-pushd "$GOPATH/src/github.com/google/ko" || exit 1
+pushd "$GOPATH/src/github.com/ko-build/ko" || exit 1
```
imjasonh, line 26: This should refer to google/ko too for now.

imjasonh (commented):
Thanks! Some of these will need to wait for the next step.

README.md
BenTheElder: I couldn't find any discussion, is there one about doing this anywhere (renaming the go import paths)? Or should someone start one and / or help PR that change?

imjasonh: This is roughly sketched out in https://github.com/ko-build/ko/issues/791 

I don't have a set timeline for it yet, I think a couple other things have to happen first. I'd guess some time after Kubecon we'll deprecate `github.com/google/ko` and point folks to importing `ko.build/pkg/build` and `go install ko.build@latest`.

As part of that we'll also move the ko.build website into this repo (#749) which also sets the vanity importpath meta tags.

PR #820, opened 2022-09-12
Handle KO_DOCKER_REPO=ko.local/repo and --bare correctly

imjasonh:
Fixes https://github.com/ko-build/ko/issues/807

cc @developer-guy 

See comments in https://github.com/ko-build/ko/issues/807#issuecomment-1243695345 -- uniformly setting `base: po.LocalDomain` in `NewDaemon` was fine when we assumed/required that `KO_DOCKER_REPO=ko.local` would only be used without a repo, but we've since changed that assumption/requirement.

```
$ KO_DOCKER_REPO=ko.local/foo go run ./ build ./test --bare
...
ko.local/foo:3a28632fcb886b7a4c78be8cd36c34ec47e83f71ff55353e5602873400dca564
```

Note that this also handles `ko.local/foo` for other namers:

```
$ KO_DOCKER_REPO=ko.local/foo go run ./ build ./test --preserve-import-paths
...
ko.local/foo/github.com/google/ko/test:3a28632fcb886b7a4c78be8cd36c34ec47e83f71ff55353e5602873400dca564
```

```
KO_DOCKER_REPO=ko.local/foo go run ./ build ./test
...
ko.local/foo/test-46c4b272b3716c422d5ff6dfc7547fa9:3a28632fcb886b7a4c78be8cd36c34ec47e83f71ff55353e5602873400dca564
```

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -168,10 +168,17 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 	innerPublisher, err := func() (publish.Interface, error) {
 		repoName := po.DockerRepo
 		namer := options.MakeNamer(po)
+		// Default LocalDomain if unset.
 		if po.LocalDomain == "" {
 			po.LocalDomain = publish.LocalDomain
 		}
-		if strings.HasPrefix(repoName, po.LocalDomain) || po.Local {
+		// If repoName is unset with --local, default it to the local domain.
+		if po.Local && repoName == "" {
+			repoName = po.LocalDomain
+		}
+		// When in doubt, if repoName is under the local domain, default to --local.
+		po.Local = strings.HasPrefix(repoName, po.LocalDomain)
+		if po.Local {
 			// TODO(jonjohnsonjr): I'm assuming that nobody will
 			// use local with other publishers, but that might
 			// not be true.
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -168,10 +168,16 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 	innerPublisher, err := func() (publish.Interface, error) {
 		repoName := po.DockerRepo
 		namer := options.MakeNamer(po)
-		if strings.HasPrefix(repoName, publish.LocalDomain) || po.Local {
+		if po.LocalDomain == "" {
+			po.LocalDomain = publish.LocalDomain
+		}
+		if strings.HasPrefix(repoName, po.LocalDomain) || po.Local {
 			// TODO(jonjohnsonjr): I'm assuming that nobody will
 			// use local with other publishers, but that might
 			// not be true.
+			if po.Bare {
+				po.LocalDomain = repoName
+			}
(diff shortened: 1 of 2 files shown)


developer-guy:
^ @eddycharly

eddycharly:
:partying_face:

pkg/commands/resolver.go
```diff
@@ -168,10 +168,16 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 	innerPublisher, err := func() (publish.Interface, error) {
 		repoName := po.DockerRepo
 		namer := options.MakeNamer(po)
-		if strings.HasPrefix(repoName, publish.LocalDomain) || po.Local {
+		if po.LocalDomain == "" {
```
eddycharly, line 171: Can't you bake this logic in the namer instead ?

imjasonh: I don't think so, at least not without changing how namers are specified today.

Namers take in the `base` and `importpath`, and use them to construct the final image. If they're only passed `publish.LocalDomain` and the importpath `github.com/foo/bar`, they don't have the information necessary to construct `ko.local/foo/github.com/foo/bar` -- they don't know about the first `foo` component.

We could change the interface so they take a third argument, which we'd append together in the end, but then we'd just have to parse it out of `KO_DOCKER_REPO` anyway to get it, so we might as well pass it directly through and append them both together and not change the interface.

eddycharly: I was thinking about something like this:
```

func bareDockerRepo(po *PublishOptions) func(string, string) string {
	return func(base, _ string) string {
		if po.LocalDomain != "" {
			return po.LocalDomain
		}
		return base
	}
}

func MakeNamer(po *PublishOptions) publish.Namer {
	if po.ImageNamer != nil {
		return po.ImageNamer
	} else if po.PreserveImportPaths {
		return preserveImportPath
	} else if po.BaseImportPaths {
		return baseImportPaths
	} else if po.Bare {
		return bareDockerRepo(po)
	}
	return packageWithMD5
}
```

Not the real logic but you get the idea, `bareDockerRepo` acts as a namer factory taking `PublishOptions` arg...

imjasonh: Yeah, that works too. That's more or less how I meant changing the interface. We could do that for all the namers, or we could just pass through the actual local domain that was specified, instead of only the `ko.local` portion.

As part of this we'll also want to support cases where the local domain is configurable by clients using ko as a library, so that if they configure the local domain to be `skaffold.local` or whatever, then `skaffold.local/foo` will also work as expected. I don't know if that's something they plan to support, but it would be nice to be consistent.

eddycharly: I guess the main point is to have the naming logic written at a single place so that it can be reused in different parts of the code that need it (I had the feeling it was the responsibility of the namer).

Anyway, it can be changed later if necessary.

jonjohnsonjr (approved):

developer-guy:
it would be good to release ko v0.13.0 that includes that fix ☝️

PR #821, opened 2022-09-12
feat: adding support for using multiple keychain for sending sbom results to a different repository

### LINKED ISSUE #819: let putting SBOMs in a different repository
We can enable putting SBOMs in a different repository by creating another environment variable or a new configuration option.

```sh
# via environment variable
SBOM_REPOSITORY=<repo> ko build ..

# via the configuration file
.ko.yaml:
sbomRepository: <repo> 
```

cc: @dentrax

> 🗣 Based on the talk: https://kubernetes.slack.com/archives/C032MM2CH7X/p1662936574448189
> 🧵 Cross-ref: https://github.com/kyverno/kyverno/issues/4592

developer-guy:

Fixes #819

### DIFF
--- a/pkg/publish/default.go
+++ b/pkg/publish/default.go
@@ -40,8 +40,8 @@ type defalt struct {
 	base      string
 	t         http.RoundTripper
 	userAgent string
-	auth      authn.Authenticator
 	namer     Namer
+	keychain  authn.Keychain
 	tags      []string
 	tagOnly   bool
 	insecure  bool
@@ -54,7 +54,7 @@ type defaultOpener struct {
 	base      string
 	t         http.RoundTripper
 	userAgent string
-	auth      authn.Authenticator
+	keychain  authn.Keychain
 	namer     Namer
 	tags      []string
 	tagOnly   bool
@@ -67,8 +67,9 @@ type Namer func(string, string) string
 
 // identity is the default namer, so import paths are affixed as-is under the repository
 // name for maximum clarity, e.g.
-//   gcr.io/foo/github.com/bar/baz/cmd/blah
-//   ^--base--^ ^-------import path-------^
+//
+//	gcr.io/foo/github.com/bar/baz/cmd/blah
+//	^--base--^ ^-------import path-------^
 func identity(base, in string) string { return path.Join(base, in) }
 
 // As some registries do not support pushing an image by digest, the default tag for pushing
@@ -89,7 +90,7 @@ func (do *defaultOpener) Open() (Interface, error) {
 		base:      do.base,
 		t:         do.t,
 		userAgent: do.userAgent,
-		auth:      do.auth,
+		keychain:  do.keychain,
 		namer:     do.namer,
 		tags:      do.tags,
 		tagOnly:   do.tagOnly,
@@ -104,7 +105,7 @@ func NewDefault(base string, options ...Option) (Interface, error) {
 		base:      base,
 		t:         http.DefaultTransport,
 		userAgent: "ko",
-		auth:      authn.Anonymous,
+		keychain:  authn.DefaultKeychain,
 		namer:     identity,
 		tags:      defaultTags,
 	}
@@ -203,7 +204,8 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
-	ro := []remote.Option{remote.WithAuth(d.auth), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	ro := []remote.Option{remote.WithAuthFromKeychain(d.keychain), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+
 	no := []name.Option{}
 	if d.insecure {
 		no = append(no, name.Insecure)
--- a/pkg/publish/options.go
+++ b/pkg/publish/options.go
@@ -16,13 +16,18 @@ package publish
 
 import (
 	"crypto/tls"
-	"log"
 	"net/http"
-	"path"
 
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/name"
 )
+
+type staticKeychain struct {
+	auth authn.Authenticator
+}
+
+func (s staticKeychain) Resolve(resource authn.Resource) (authn.Authenticator, error) {
+	return s.auth, nil
+}
 
 // WithTransport is a functional option for overriding the default transport
 // on a default publisher.
@@ -44,9 +49,9 @@ func WithUserAgent(ua string) Option {
 
 // WithAuth is a functional option for overriding the default authenticator
 // on a default publisher.
-func WithAuth(auth authn.Authenticator) Option {
+func WithAuth(a authn.Authenticator) Option {
 	return func(i *defaultOpener) error {
-		i.auth = auth
+		i.keychain = staticKeychain{a}
 		return nil
 	}
 }
@@ -55,25 +60,7 @@ func WithAuth(auth authn.Authenticator) Option {
 // authenticator on a default publisher using an authn.Keychain
 func WithAuthFromKeychain(keys authn.Keychain) Option {
 	return func(i *defaultOpener) error {
-		// We parse this lazily because it is a repository prefix, which
-		// means that docker.io/mattmoor actually gets interpreted as
-		// docker.io/library/mattmoor, which gets tricky when we start
-		// appending things to it in the publisher.
-		//
-		// We append a fake path "ko" to KO_DOCKER_REPO in order to
-		// make parsing out the registry easier.
-		repo, err := name.NewRepository(path.Join(i.base, "ko"))
-		if err != nil {
-			return err
-		}
-		auth, err := keys.Resolve(repo.Registry)
-		if err != nil {
-			return err
-		}
-		if auth == authn.Anonymous {
-			log.Println("No matching credentials were found, falling back on anonymous")
-		}
-		i.auth = auth
+		i.keychain = keys
 		return nil
 	}
 }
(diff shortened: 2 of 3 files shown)


pkg/publish/options.go
```diff
@@ -42,38 +38,11 @@ func WithUserAgent(ua string) Option {
...
-	return func(i *defaultOpener) error {
-		i.auth = auth
-		return nil
-	}
-}
-
-// WithAuthFromKeychain is a functional option for overriding the default
+// WithKeyChain is a functional option for overriding the default
 // authenticator on a default publisher using an authn.Keychain
-func WithAuthFromKeychain(keys authn.Keychain) Option {
```
imjasonh, line 56: We shouldn't rename this, since other users who use ko as a library might use it. Same with `WithAuth` above.

pkg/publish/default.go
```diff
@@ -129,7 +133,7 @@ func pushResult(ctx context.Context, tag name.Tag, br build.Result, opt []remote
 		ociOpts := []ociremote.Option{ociremote.WithRemoteOptions(opt...)}
 
 		// Respect COSIGN_REPOSITORY
-		targetRepoOverride, err := ociremote.GetEnvTargetRepository()
```
imjasonh, line 132: Does this mean we _already_ support `COSIGN_REPOSITORY`? 🤯 

If so, maybe we should just use this for now and document a plan to support non-`COSIGN`-named env vars in the future.

developer-guy: >Does this mean we already support COSIGN_REPOSITORY? 🤯

Haha, yes 🫣

>If so, maybe we should just use this for now and document a plan to support non-COSIGN-named env vars in the future.

As you wish, but this keychain problem remains the same, at least we should fix this.

imjasonh: Ah I get what you're saying now. Yeah, instead of taking an auth we should take a keychain, since we might be pushing to two repos/registries due to COSIGN_REPOSITORY.

Thanks for trying this out and for figuring out the fix. 👍

developer-guy: >instead of taking an auth we should take a keychain

Exactly!

pkg/publish/default.go
```diff
@@ -67,8 +69,9 @@ type Namer func(string, string) string
 
 // identity is the default namer, so import paths are affixed as-is under the repository
 // name for maximum clarity, e.g.
-//   gcr.io/foo/github.com/bar/baz/cmd/blah
-//   ^--base--^ ^-------import path-------^
+
+//	gcr.io/foo/github.com/bar/baz/cmd/blah
+//	^--base--^ ^-------import path-------^
```
imjasonh, line 74: undo this change.

pkg/commands/resolver.go
```diff
@@ -209,7 +209,7 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 		if po.Push {
 			dp, err := publish.NewDefault(repoName,
 				publish.WithUserAgent(userAgent),
-				publish.WithAuthFromKeychain(keychain),
+				publish.WithKeyChain(keychain),
```
imjasonh, line 212: This has to keep the name `publish.WithAuthFromKeychain`.

pkg/publish/options.go
```diff
@@ -78,6 +78,15 @@ func WithAuthFromKeychain(keys authn.Keychain) Option {
 	}
 }
 
+// WithKeyChain is a functional option for overriding the default
+// authenticator on a default publisher using an authn.Keychain
+func WithKeyChain(keys authn.Keychain) Option {
```
imjasonh, line 83: We don't need to add a new method, we can just keep using `WithAuthFromKeychain`.

imjasonh (commented):
Lookin' good, just a few small things.

pkg/publish/default.go
```diff
@@ -203,7 +207,7 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
-	ro := []remote.Option{remote.WithAuth(d.auth), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	ro := []remote.Option{remote.WithAuthFromKeychain(d.keychain), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
```
imjasonh, line 210: Hmm... So now we have the case where a client may have used `publish.WithAuth`, and with this change we may be dropping it on the floor.

I think what we should do is check whether `d.auth == authn.Anonymous`, and only if it is, add `remote.WithAuthFromKeychain(d.keychain)` -- otherwise, keep `remote.WithAuth(d.auth)`. We might want to error if both are passed too, to avoid confusion.

This should only affect clients like Skaffold that use this code as a library though.

cc @halvards

developer-guy: I fixed it, it should be okay right now.

halvards: Thanks for keeping ko-as-library users in mind :slightly_smiling_face:

developer-guy: kindly ping @imjasonh ☝️

pkg/publish/default.go
```diff
@@ -203,7 +207,13 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
-	ro := []remote.Option{remote.WithAuth(d.auth), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	ro := []remote.Option{remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	if d.auth == authn.Anonymous {
```
imjasonh, line 211:
```suggestion
	if d.auth != authn.Anonymous {
```

Right?

Maybe we should only store a `keychain`, and if someone calls `WithAuth` we should set `d.keychain` to a keychain that only returns that auth. Then we don't have to handle a possible situation where we store both an auth and a keychain, and have to decide which one to use.

developer-guy: >Maybe we should only store a keychain
That's how I did it first. Do you want me to revert to the first version of the development?

^Based on that, if I understand correctly, we should remove the if logic and use the keychain only.

Right?

imjasonh: Yeah, this struct would only hold a `keychain`, either the exactly the one passed by `WithAuthFromKeychain`, or a new keychain impl that always returns the auth passed by `WithAuth`.

developer-guy: >this struct would only hold a keychain, either the exactly the one passed by WithAuthFromKeychain

I would like to go with this, wdyt?

realshuting:
Thanks for the work @developer-guy and the review @imjasonh ! Would like to get the support in!

developer-guy:
I think we are in good shape about this PR

pkg/publish/default.go
```diff
@@ -40,8 +40,9 @@ type defalt struct {
 	base      string
 	t         http.RoundTripper
 	userAgent string
-	auth      authn.Authenticator
 	namer     Namer
+	auth      authn.Authenticator
```
imjasonh, line 44: This shouldn't have an `auth` field anymore, only `keychain`. `publish.WithAuth` should set `keychain = staticKeychain{auth}`.

pkg/publish/default.go
```diff
@@ -55,6 +56,7 @@ type defaultOpener struct {
 	t         http.RoundTripper
 	userAgent string
 	auth      authn.Authenticator
```
imjasonh, line 58: Same here.

pkg/publish/default.go
```diff
@@ -203,7 +208,13 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
-	ro := []remote.Option{remote.WithAuth(d.auth), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	ro := []remote.Option{remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
+	if d.auth != authn.Anonymous {
```
imjasonh, line 212: This should just always pass `remote.WithAuthFromKeychain(d.keychain)`, which is `staticKeychain{authn.Anonymous}` by default.

pkg/publish/options.go
```diff
@@ -16,14 +16,19 @@ package publish
...
 
 	"github.com/google/go-containerregistry/pkg/authn"
-	"github.com/google/go-containerregistry/pkg/name"
 )
 
+type staticKeychain struct {
+	auth authn.Authenticator
+}
+
+func (s *staticKeychain) Resolve(resource authn.Resource) (authn.Authenticator, error) {
```
imjasonh, line 28: nit: just `staticKeychain` (no `*`), since there's no situation where a call should update the value.

pkg/publish/options.go
```diff
@@ -55,25 +60,7 @@ func WithAuth(auth authn.Authenticator) Option {
...
-			return err
-		}
-		auth, err := keys.Resolve(repo.Registry)
-		if err != nil {
-			return err
-		}
-		if auth == authn.Anonymous {
-			log.Println("No matching credentials were found, falling back on anonymous")
-		}
-		i.auth = auth
```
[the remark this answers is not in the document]
developer-guy: >will they still see this line?

I think they won't see this line anymore since we are not resolving any keychain to auth. Also, shouldn't it be an expected thing thus, we are changing the auth strategy by looking at the keychain itself with this change. What is your suggestion?

imjasonh (approved):

imjasonh:
I'd love to see a test for this new behavior, but I've already taken long enough reviewing this so I won't block on it.

Thanks for adding this! 👍

PR #822, opened 2022-09-14
feat: write sbom result to disk

### LINKED ISSUE #728: Output SBOM to file in addition to pushing
## Problem

In `ko` today, building a container image with the `publish` command, for example, will simultaneously push an SBOM to the same registry. But when attesting an SBOM with `cosign`, the SBOM must be downloaded first. This causes a push-pull-push to occur which is not optimal. 

## Solution

As discussed with @imjasonh in [this Slack discussion](https://kubernetes.slack.com/archives/C01T7DTP65S/p1655132831080329), it may be good to add a flag like `--sbom-path` which would allow simultaneously outputting the SBOM to a file so it can be attested to. Without the flag, there is no behavioral change. With the flag, the existing behavior still remains the same but additionally the SBOM will be written to a file locally.

developer-guy:

Fixes #728

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -37,7 +37,11 @@ jobs:
         KO_DOCKER_REPO="" go run ./ build --push=false ./test | grep ":latest@sha256:"
         KO_DOCKER_REPO="" go run ./ build --push=false -t test ./test | grep ":test@sha256:"
         KO_DOCKER_REPO="" go run ./ build --push=false -t test --tag-only ./test | grep ":test$"
-
+        
+        # Check that using sbom-dir works.
+        KO_DOCKER_REPO=""  go run ./ build -t test --push=false --sbom-dir ./sbom-data ./test
+        jq . ./sbom-data/test.spdx.json
+        
         export PLATFORM=${GOOS}/${GOARCH}
 
         if [[ "${{ matrix.platform }}" == "windows-latest" ]]; then
--- a/doc/ko_run.md
+++ b/doc/ko_run.md
@@ -44,6 +44,7 @@ ko run IMPORTPATH [flags]
   -P, --preserve-import-paths    Whether to preserve the full import path after KO_DOCKER_REPO.
       --push                     Push images to KO_DOCKER_REPO (default true)
       --sbom string              The SBOM media type to use (none will disable SBOM synthesis and upload, also supports: spdx, cyclonedx, go.version-m). (default "spdx")
+      --sbom-dir string          Path to file where the SBOM will be written.
       --tag-only                 Include tags but not digests in resolved image references. Useful when digests are not preserved when images are repopulated.
   -t, --tags strings             Which tags to use for the produced image instead of the default 'latest' tag (may not work properly with --base-import-paths or --bare). (default [latest])
       --tarball string           File to save images tarballs
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -63,7 +63,7 @@ type GetBase func(context.Context, string) (name.Reference, Result, error)
 
 type builder func(context.Context, string, string, v1.Platform, Config) (string, error)
 
-type sbomber func(context.Context, string, string, oci.SignedEntity) ([]byte, types.MediaType, error)
+type sbomber func(context.Context, string, string, string, oci.SignedEntity, string) ([]byte, types.MediaType, error)
 
 type platformMatcher struct {
 	spec      []string
@@ -77,6 +77,7 @@ type gobuild struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	sbom                 sbomber
+	sbomDir              string
 	disableOptimizations bool
 	trimpath             bool
 	buildConfigs         map[string]Config
@@ -98,6 +99,7 @@ type gobuildOpener struct {
 	kodataCreationTime   v1.Time
 	build                builder
 	sbom                 sbomber
+	sbomDir              string
 	disableOptimizations bool
 	trimpath             bool
 	buildConfigs         map[string]Config
@@ -125,6 +127,7 @@ func (gbo *gobuildOpener) Open() (Interface, error) {
 		kodataCreationTime:   gbo.kodataCreationTime,
 		build:                gbo.build,
 		sbom:                 gbo.sbom,
+		sbomDir:              gbo.sbomDir,
 		disableOptimizations: gbo.disableOptimizations,
 		trimpath:             gbo.trimpath,
 		buildConfigs:         gbo.buildConfigs,
@@ -301,7 +304,7 @@ func build(ctx context.Context, ip string, dir string, platform v1.Platform, con
 	return file, nil
 }
 
-func goversionm(ctx context.Context, file string, appPath string, se oci.SignedEntity) ([]byte, types.MediaType, error) {
+func goversionm(ctx context.Context, file string, appPath string, appFileName string, se oci.SignedEntity, dir string) ([]byte, types.MediaType, error) {
 	switch se.(type) {
 	case oci.SignedImage:
 		sbom := bytes.NewBuffer(nil)
@@ -314,7 +317,13 @@ func goversionm(ctx context.Context, file string, appPath string, se oci.SignedE
 
 		// In order to get deterministics SBOMs replace our randomized
 		// file name with the path the app will get inside of the container.
-		return []byte(strings.Replace(sbom.String(), file, appPath, 1)), "application/vnd.go.version-m", nil
+		s := []byte(strings.Replace(sbom.String(), file, appPath, 1))
+
+		if err := writeSBOM(s, appFileName, dir, "go.version-m"); err != nil {
+			return nil, "", err
+		}
+
+		return s, "application/vnd.go.version-m", nil
 
 	case oci.SignedImageIndex:
 		return nil, "", nil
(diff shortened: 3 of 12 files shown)


pkg/commands/resolver.go
```diff
@@ -120,6 +120,17 @@ func gobuildOptions(bo *options.BuildOptions) ([]build.Option, error) {
 		opts = append(opts, build.WithConfig(bo.BuildConfigs))
 	}
 
+	if bo.SBOMDir != "" {
+		if bo.SBOMDir == "." {
```
imjasonh, line 124: We shouldn't special-case `.`, since users might also do `../foo` etc. Look into what `--image-refs` does, so we're at least consistent. If that doesn't support `..` etc (I doubt it does), that's okay, but we should be consistent.

pkg/build/gobuild.go
```diff
@@ -874,6 +880,14 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
...
 			return nil, err
 		}
+
+		if g.sbomDir != "" {
+			fp := fmt.Sprintf("%s.%s", filepath.Join(g.sbomDir, appFilename(ref.Path())), g.sbomType)
+			log.Printf("Writing an SBOM to %s", fp)
+			if err := os.WriteFile(fp, sbom, 0644); err != nil {
+				return nil, err
+			}
+		}
```
imjasonh, line 890: Can this be part of the sbomber impl's responsibility? It already knows the type of SBOM it's generating, and can also know its file name extension.

pkg/build/options.go
```diff
@@ -158,6 +161,7 @@ func WithCycloneDX() Option {
 func withSBOMber(sbom sbomber) Option {
 	return func(gbo *gobuildOpener) error {
 		gbo.sbom = sbom
+		gbo.sbomType = "sbomber"
```
imjasonh, line 164: `withSBOMber` is used to pass a custom SBOM impl internally, it shouldn't set the sbomType like this.

This is actually kind of a signal that sbomType isn't a good way to express this. Maybe we can make writing the file part of the sbomer interface's responsibility?

pkg/build/gobuild.go
developer-guy: yes, it can but wouldn't it cause code duplication? Because I need to add this logic into all of the somber impls. WDYT?

developer-guy: I updated the code according to your feedback, can you please take a look? Thx

pkg/build/gobuild.go
```diff
@@ -349,11 +371,30 @@ func spdx(version string) sbomber {
...
+			if err != nil {
+				log.Println(err)
+			}
+		}
+		filePath := fmt.Sprintf("%s.%s", appDir, ext)
+		log.Printf("Writing an SBOM to a file %s", filePath)
+		err := os.WriteFile(filePath, sbom, 0644)
+		if err != nil {
+			return err
+		}
```
imjasonh, line 388:
```suggestion
		return os.WriteFile(filePath, sbom, 0644)
```

pkg/build/gobuild.go
```diff
@@ -349,11 +371,30 @@ func spdx(version string) sbomber {
...
+	if dir != "" {
+		appDir := path.Join(dir, appPath)
+		if _, err := os.Stat(appDir); errors.Is(err, os.ErrNotExist) {
+			err := os.MkdirAll(appDir, os.ModePerm)
+			if err != nil {
+				log.Println(err)
+			}
+		}
+		filePath := fmt.Sprintf("%s.%s", appDir, ext)
+		log.Printf("Writing an SBOM to a file %s", filePath)
```
imjasonh, line 384:
```suggestion
		log.Printf("Writing SBOM to %s", filePath)
```

pkg/build/gobuild.go
```diff
@@ -349,11 +371,30 @@ func spdx(version string) sbomber {
 	}
 }
 
+func writeSBOM(sbom []byte, appPath, dir, ext string) error {
+	if dir != "" {
+		appDir := path.Join(dir, appPath)
```
imjasonh, line 376: This should be `filepath.Join` in case we're running on Windows.

Should we `filepath.Clean` this too, just in case?

pkg/build/gobuild.go
```diff
@@ -349,11 +371,30 @@ func spdx(version string) sbomber {
...
+		if _, err := os.Stat(appDir); errors.Is(err, os.ErrNotExist) {
```
imjasonh, line 377: I think we can just always `MkdirAll`, which will exit successfully if it already exists.

pkg/commands/options/build.go
```diff
@@ -51,8 +51,10 @@ type BuildOptions struct {
 	ConcurrentBuilds     int
 	DisableOptimizations bool
 	SBOM                 string
-	Platforms            []string
-	Labels               []string
+	// SBOMDir is the directory to write the SBOM to.
```
imjasonh, line 54:
```suggestion
```

I don't think the comment adds much, and it's weird to have a comment on this and nothing else.

doc/ko_run.md
```diff
@@ -44,6 +44,7 @@ ko run IMPORTPATH [flags]
   -P, --preserve-import-paths    Whether to preserve the full import path after KO_DOCKER_REPO.
       --push                     Push images to KO_DOCKER_REPO (default true)
       --sbom string              The SBOM media type to use (none will disable SBOM synthesis and upload, also supports: spdx, cyclonedx, go.version-m). (default "spdx")
+      --sbom-dir string          The directory to write the SBOM to.
```
imjasonh, line 47:
```suggestion
      --sbom-dir string          Path to file where the SBOM will be written.
```

I'd like to think about making this more consistent with `--image-refs` if possible, which takes a path to a file, instead of a directory.

That may bite us later if/when we end up generating multiple SBOMs, but I'm not sure we'll do that anyway. If we do, we can only write the first one to disk or something. WDYT?

developer-guy: >I'd like to think about making this more consistent with --image-refs if possible, which takes a path to a file instead of a directory.
That's also ok for me. It should change the logic of determining the file path, that's all (I guess?). But what if I'm building more than one binary configured via the [builds](https://github.com/ko-build/ko#overriding-go-build-settings] section?

pkg/commands/resolver.go
imjasonh: ping on this comment.

doc/ko_run.md
developer-guy: Is there anything left that you want me to do? Or are we all good 🕺🏻

imjasonh: What do you think about adding an e2e test in e2e.yaml? You could even have it `--push=false` with `--sbom-dir` then check that the file exists and can be read with `jq . <file>`.

.github/workflows/e2e.yaml
```diff
@@ -72,4 +72,8 @@ jobs:
           - "-X main.version=${{ github.sha }}"
         EOF
         docker run $(go run ./ build ./test/ --platform=${PLATFORM}) --wait=false 2>&1 | grep "${{ github.sha }}"
+        
+        # Check that using sbom-dir works.
```
developer-guy, line 76: see @imjasonh

doc/ko_run.md
developer-guy: I have added a test, can you please take a look at them? Thx

.github/workflows/e2e.yaml
```diff
@@ -72,4 +72,8 @@ jobs:
...
+        KO_DOCKER_REPO=""  go run ./ build -t test --push=false --sbom-dir ./sbom-data ./test | grep "Writing SBOM to"
+        jq . ./sbom-data/ko-app/*.json
```
imjasonh, line 78: The output should not be written to a `ko-app` directory, it should be a filename like `sbom-dir/<import-path>.json` (with slashes replaced).

.github/workflows/e2e.yaml
```diff
@@ -72,4 +72,8 @@ jobs:
           - "-X main.version=${{ github.sha }}"
         EOF
         docker run $(go run ./ build ./test/ --platform=${PLATFORM}) --wait=false 2>&1 | grep "${{ github.sha }}"
+        
+        # Check that using sbom-dir works.
+        KO_DOCKER_REPO=""  go run ./ build -t test --push=false --sbom-dir ./sbom-data ./test | grep "Writing SBOM to"
```
imjasonh, line 77: If you `jq *.json` in the next line, you don't need to `grep` out the line. It might be nice to see all the output, and if writing fails or doesn't happen then the next line will fail anyway.

.github/workflows/e2e.yaml
```diff
@@ -37,7 +37,11 @@ jobs:
         KO_DOCKER_REPO="" go run ./ build --push=false ./test | grep ":latest@sha256:"
         KO_DOCKER_REPO="" go run ./ build --push=false -t test ./test | grep ":test@sha256:"
         KO_DOCKER_REPO="" go run ./ build --push=false -t test --tag-only ./test | grep ":test$"
-
+        
+        # Check that using sbom-dir works.
+        KO_DOCKER_REPO=""  go run ./ build -t test --push=false --sbom-dir ./sbom-data ./test
+        jq . ./sbom-data/*.json
```
imjasonh, line 43: I think now we could test that this is `./sbom-data/test.json` right?

Otherwise LGTM!

imjasonh (approved):

PR #846, opened 2022-10-12
doc: fix link to Installation page in Getting Started

antoineco:
Fixes the 404 error when clicking the link to the _Installation_ page located at the top of the _Getting Started_ page.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -2,7 +2,7 @@
 
 ## Setup
 
-First, [install `ko`](./install).
+First, [install `ko`](../install).
 
 ### Authenticate
 


imjasonh:
Thanks for sending this fix!

This looks right to me, but now I'm worried we have other similar links scattered around.

I wonder if we should configure [`use_directory_urls: false`](https://www.mkdocs.org/user-guide/configuration/#use_directory_urls) so the URLs look like `https://ko.build/get_started.html` instead, and `./install` would work again (maybe only with `.html`?)

I don't think this comment needs to block the PR (thank you again!) I'm just ruminating about how we want URLs to look and work.

I should also run a link checker on the site to see if anything else is broken. 💀

imjasonh (approved):

antoineco:
That could be an option, potentially it could make relative links more intuitive but I have to admit I have only scratched the surface with mkdocs!

PR #866, opened 2022-10-24
bump ggcr and handle breaking change

imjasonh:
Supercedes https://github.com/ko-build/ko/pull/865


jonjohnsonjr (approved):

JohnTitor:
@imjasonh Could you cut a new release including this? It would fix a build issue with a recent build tool.

imjasonh:
> @imjasonh Could you cut a new release including this? It would fix a build issue with a recent build tool. Thanks in advance!

If you depend on `ko` as a library, you should be able to `go get -u github.com/google/ko@main`, or add a `replace` to your `go.mod` to pick up this change. If those don't work for some reason, let me know and I can try to help debug it.

We plan to cut a release in the new year, to account for some importpath changes we have planned.

JohnTitor:
Yeah, that indeed fixed our issue but it'd be great if we have a release to specify a tag instead of a commit SHA.

> We plan to cut a release in the new year, to account for some importpath changes we have planned.

Great, thank you!

PR #868, opened 2022-10-25
Publish an tagged image on release

### LINKED ISSUE #847: Tagged "package" for released version of `ko`
As of today, there is no tag on `ghcr.io/ko-build/ko` package (well oci image). This means that to use a released version (`v0.12` for example) we have to "guess" which digest it correspond to.

cc @imjasonh

vdemeester:
Fixes #847

Signed-off-by: Vincent Demeester <vincent@sbr.pm>


vdemeester:
cc @imjasonh @mattmoor

.github/workflows/release.yml
```diff
@@ -35,6 +35,24 @@ jobs:
...
+        with:
+          go-version: 1.18
+          check-latest: true
+      - uses: imjasonh/setup-ko@v0.6
+      - name: push image on ghcr
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
+          ko push --base-import-paths -t ${tag} .
```
mattmoor, line 54: I think `ko push` is incorrect, and we probably want to adjust this either way:
https://github.com/ko-build/ko/blob/e23aebe18bc2ba0906a500da8592ce60b14668c4/.github/workflows/image.yaml#L33

vdemeester: @mattmoor you are right 🤦🏼 it should be `publish` or `build` indeed ! I'll update 👼🏼

vdemeester: However, I do want to tag it with the tag name (0.12.0) versus the sha (commit id)

.github/workflows/release.yml
```diff
@@ -35,6 +35,24 @@ jobs:
...
+    # outputs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: git fetch --prune --unshallow
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+          check-latest: true
+      - uses: imjasonh/setup-ko@v0.6
```
imjasonh, line 48:
```suggestion
      - uses: imjasonh/setup-ko@v0.6 # This installs the current latest release.
```

(Just to save me some thinking when I read this later 😄)

.github/workflows/release.yml
```diff
@@ -35,6 +35,24 @@ jobs:
           checksum_file=$(echo "$ARTIFACTS" | jq -r '.[] | select (.type=="Checksum") | .path')
           echo "::set-output name=hashes::$(cat $checksum_file | base64 -w0)"
 
+  publish:
+    # outputs:
```
imjasonh, line 39:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -35,6 +35,23 @@ jobs:
...
+        with:
+          go-version: 1.18
+          check-latest: true
+      - uses: imjasonh/setup-ko@v0.6 # This installs the current latest release.
+      - name: push image on ghcr
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
+          ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./
```
imjasonh, line 53:
```suggestion
      - run: |
          tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
          ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./
```

The env is unneeded because setup-ko already sets up auth if you're pushing to GHCR, and in general I just drop the `name` if it's pretty clear what the step does.

.github/workflows/release.yml
mattmoor: Sure, I was thinking we'd commit bash crimes to get a list into `--tags`. 😉 

Really the main thing I'm worried about is that we're not signing these images, which the other flow is doing.

vdemeester: @mattmoor indeed, that would have been my follow-up question 🙃 What would be the flow to sign those images as well ? 👼🏼

.github/workflows/release.yml
```diff
@@ -35,6 +35,20 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - run: git fetch --prune --unshallow
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+          check-latest: true
+      - uses: imjasonh/setup-ko@v0.6 # This installs the current latest release.
+      - run: |
+          tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
+          ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./
```
imjasonh, line 50:
```suggestion
      - uses: imjasonh/setup-ko@v0.6 # This installs the current latest release.
      - uses: sigstore/cosign-installer@v2.8.1
      - run: |
          tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
          img=$(ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./)
          echo "built ${img}"
          cosign sign ${img} \
              -a sha=${{ github.sha }} \
              -a run_id=${{ github.run_id }} \
              -a run_attempt=${{ github.run_attempt }} \
              -a tag=${tag}
```

cribbed from here: https://github.com/ko-build/ko/blob/f158992ccb55617b98b8c580d7a7e25a2aaa3ae8/.github/workflows/image.yaml#L35-L38

imjasonh (approved):

.github/workflows/release.yml
```diff
@@ -35,6 +35,27 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: git fetch --prune --unshallow
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+          check-latest: true
+      - uses: imjasonh/setup-ko@v0.6 # This installs the current latest release.
+      - uses: sigstore/cosign-installer@v2.8.1
```
mattmoor, line 48: Do we need `COSIGN_EXPERIMENTAL=true` or does this grab a version that no longer needs it?

mattmoor (approved):

.github/workflows/release.yml
vdemeester: Not sure I understand the question 🙃

imjasonh: > Not sure I understand the question 🙃

`COSIGN_EXPERIMENTAL` was needed to enter the keyless signing flow, but that was recently removed since keyless signing went GA.

cosign-installer@v2.8.1 installs cosign

PR #870, opened 2022-10-26
Layout omit image.ref.name annotation

### LINKED ISSUE #869: associate oci-image-layout directory with the tag of an image
We (w/@dentrax) noticed that when you build your container image with the `--oci-layout-path` flag enabled, the image tag associated with it is lost, [regctl](https://github.com/regclient/regclient) uses an annotation in the image's manifest (index.json) like the following to associate the tag with the image.

```shell
...
   "annotations": {
        "org.opencontainers.image.ref.name": "latest"
      }
```

Could we do the same one for ko builds, WDYT @imjasonh?

Dentrax:
Fixes #869

Signed-off-by: Furkan <furkan.turkal@trendyol.com>
Co-authored-by: Batuhan <batuhan.apaydin@trendyol.com>

cc @developer-guy

- [ ] tests a bit missing, we will add some assertions to check if all annotations are valid and correct. (if overall idea is OK)

### DEMO

```
$ KO_DOCKER_REPO="foo" LDFLAGS="-s" go run . build -t foo,bar,baz --push=false --oci-layout-path test .

$ cat test/index.json | jq
```

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "size": 1943,
      "digest": "sha256:99a7a7bf1ad3fde0107b1f5909bb9a1df037f8be8b73e73d1779a0fd473cf1ed",
      "annotations": {
        "org.opencontainers.image.ref.name": "foo"
      }
    },
    {
      "size": 1943,
      "digest": "sha256:99a7a7bf1ad3fde0107b1f5909bb9a1df037f8be8b73e73d1779a0fd473cf1ed",
      "annotations": {
        "org.opencontainers.image.ref.name": "bar"
      }
    },
    {
      "size": 1943,
      "digest": "sha256:99a7a7bf1ad3fde0107b1f5909bb9a1df037f8be8b73e73d1779a0fd473cf1ed",
      "annotations": {
        "org.opencontainers.image.ref.name": "baz"
      }
    }
  ]
}
```

### DIFF
--- a/pkg/publish/layout.go
+++ b/pkg/publish/layout.go
@@ -25,14 +25,16 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/layout"
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/google/ko/pkg/build"
+	specsv1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 type LayoutPublisher struct {
-	p layout.Path
+	p    layout.Path
+	tags []string
 }
 
 // NewLayout returns a new publish.Interface that saves images to an OCI Image Layout.
-func NewLayout(path string) (Interface, error) {
+func NewLayout(path string, tags ...string) (Interface, error) {
 	p, err := layout.FromPath(path)
 	if err != nil {
 		p, err = layout.Write(path, empty.Index)
@@ -40,7 +42,10 @@ func NewLayout(path string) (Interface, error) {
 			return nil, err
 		}
 	}
-	return &LayoutPublisher{p}, nil
+	if len(tags) == 0 {
+		tags = []string{latestTag}
+	}
+	return &LayoutPublisher{p, tags}, nil
 }
 
 func (l *LayoutPublisher) writeResult(br build.Result) error {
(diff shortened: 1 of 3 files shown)


developer-guy:
also ccing @imjasonh @jonjohnsonjr 🙈

developer-guy:
kindly ping @imjasonh ☝️

pkg/publish/layout_test.go
```diff
@@ -25,25 +25,52 @@ import (
...
-		t.Fatalf("random.Image() = %v", err)
+	cases := []struct {
+		name string
+		tags []string
+		want error
+	}{
+		{
+			name: "no tags",
+			tags: nil,
+			want: nil,
```
imjasonh, line 36: `want` is `nil` in all of these cases. If none of these cases test the error scenario we can just remove `want` from the struct here.

pkg/publish/layout.go
```diff
@@ -25,22 +25,27 @@ import (
...
 
 type LayoutPublisher struct {
-	p layout.Path
+	p    layout.Path
+	tags []string
 }
 
 // NewLayout returns a new publish.Interface that saves images to an OCI Image Layout.
-func NewLayout(path string) (Interface, error) {
+func NewLayout(path string, tags []string) (Interface, error) {
```
imjasonh, line 37: FYI this changes the method signature, and will break anyone that uses this as a library. I don't think this is a blocker, but if we expect that to be something people do more in the future we should be more careful.

Not a blocker this time though.

imjasonh (commented):
This looks good! Thanks for doing this, and sorry as always for the delay.

pkg/publish/layout.go
developer-guy: We can change it as a variadic function, this does not affect people who might be using the ko project as a library.

developer-guy: it should be good now

developer-guy:
kindly ping @Dentrax, we should fix the tests 🙈

developer-guy:
kindly ping @imjasonh 🙈

developer-guy:
kindly ping @imjasonh 🕺🏻

imjasonh (dismissed):

imjasonh:
@jonjohnsonjr any concerns?

pkg/publish/layout.go
```diff
@@ -25,22 +25,27 @@ import (
...
 	p, err := layout.FromPath(path)
 	if err != nil {
 		p, err = layout.Write(path, empty.Index)
 		if err != nil {
 			return nil, err
 		}
 	}
-	return &LayoutPublisher{p}, nil
+	if len(tags) == 0 {
+		tags = []string{"latest"}
```
jonjohnsonjr, line 46: I'd drop this so it's possible to use this without adding tags?

jonjohnsonjr (changes requested):

pkg/publish/layout.go
developer-guy: @dentrax are you still interested in working on this?

developer-guy:
kindly ping @jonjohnsonjr @imjasonh, it seems that everything is working fine.

developer-guy:
kindly ping @Dentrax

PR #879, opened 2022-11-03
install: fail with 404 instead of gzip error when url was wrong

grosser:
```
       -S, --show-error
              When used with -s, --silent, it makes curl show an error message if it fails.
       -f, --fail
              (HTTP) Fail silently (no output at all) on server errors. This is mostly done to enable scripts etc to better deal with failed attempts. In
              normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and
              more). This flag will prevent curl from outputting that and return error 22.
```

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -3,7 +3,7 @@
 ### Install from [GitHub Releases](https://github.com/ko-build/ko/releases)
 
 ```
-$ VERSION=TODO # choose the latest version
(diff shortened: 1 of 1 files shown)


imjasonh:
Thanks for this! I should also update https://github.com/imjasonh/setup-ko/blob/main/action.yml#L41 accordingly.

imjasonh (approved):

grosser:
merge ?

Issue #883, opened 2022-11-08
Deduplicate tags


bluebrown:
It would be nice if duplicate tags would be removed before publishing to avoid pushing the same tag multiple times.

PR #884, opened 2022-11-08
feat: deduplicate tags

bluebrown:
Signed-off-by: Nico Braun <rainbowstack@gmail.com>

resolve: #883

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -167,6 +167,8 @@ func NewPublisher(po *options.PublishOptions) (publish.Interface, error) {
 }
 
 func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
+	// use each tag only once
+	po.Tags = unique(po.Tags)
 	// Create the publish.Interface that we will use to publish image references
 	// to either a docker daemon or a container image registry.
 	innerPublisher, err := func() (publish.Interface, error) {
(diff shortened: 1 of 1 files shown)


bluebrown:
I decided to move the logic to this place so that, it applies to all publishers.

I wanted to add some tests as well, but I didn't find a good way. This is because the publisher interface doesn't expose a way to get the tags. So it's hard to write a test to compare input and output tags.

imjasonh:
(This isn't commentary on the PR exactly, just research notes I thought it would be useful to share)

The only place we care about the order of the tags is here:

https://github.com/ko-build/ko/blob/140ac8788b9c105eda6402908e940f398c2dc28d/pkg/commands/resolver.go#L241-L246

This is pretty esoteric. This happens if you're not publishing, but specify multiple tags. This might happen if you do:

```
ko build --tarball=foo.tar --push=false --tags=foo --tags=bar
```

In that case the tarball should have the `foo` tag applied consistently, and not `bar`.

While looking into this, I also found and fixed https://github.com/ko-build/ko/pull/885 which was just kinda weird.

imjasonh (dismissed):
This change LGTM, just needs a CLA before we can merge it.

bluebrown:
I have accepted the CLA, and did the rescan :)

pkg/commands/resolver.go
```diff
@@ -470,3 +472,19 @@ func resolveFile(
 
 	return buf.Bytes(), nil
 }
+
+// create a set from the input slice
+// preserving the order of unique elements
+func unique[T comparable](tt []T) []T {
```
imjasonh, line 478: Ehh, I don't know if we need to bother with generics here, since we know we'll only call it with strings.

bluebrown: Yes, I had it only for strings before, but then I thought it may be useful for other stuff too. You mentioned platform, for example. Maybe that's also a string, not sure. I can change it, if you want, though.

imjasonh: Yeah I'd prefer to have it work on strings until/unless we find a generic use for it in the future. YAGNI and all.

bluebrown: ok done. I did a force push. Not sure how this is done here, but I know some project not doing squash merge and thefore prefer a single commit.

imjasonh (approved):

imjasonh (approved):

PR #895, opened 2022-11-22
add KO_DEFAULTBASEIMAGE usage to docs

developer-guy:

/cc @imjasonh


docs/configuration.md
```diff
@@ -20,6 +20,12 @@ You can override this base image in two ways:
 defaultBaseImage: registry.example.com/base/image
 ```
 
+As an alternative way, you can use the `KO_DEFAULTBASEIMAGE` special environment variable to set the default base image to be used in `ko` builds as well:
```
imjasonh, line 23:
```suggestion
You can also use the `KO_DEFAULTBASEIMAGE` environment variable to set the default base image, which overrides the YAML configuration:
```

PR #897, opened 2022-12-01
Add option to configure default platforms

### LINKED ISSUE #521: FR: configure platforms in .ko.yaml
In the new fancy world of Mac M1 Apple silicon, I run into the issue of my dev loop causing me grief because when I target an arch that is not arm64 (like my normal k8s clusters).

What if ko could probe the target cluster and figure out the `--platform` flag?

**edit(jason):** see https://github.com/ko-build/ko/issues/521#issuecomment-1325676486

ReToCode:
Adds an option to configure default platforms either in `.ko.yaml` or using the env variable `KO_DEFAULTPLATFORMS`.

Fixes #521


ReToCode:
@imjasonh something like this? Unfortunately there is no easy way to test this properly without either
* exposing internals of `gobuildOpener` to test this from `resolver_test.go`
* or extending the `build.Interface` interface similar to `QualifyImport` to test it in `resolver_test.go`

WDYT?

imjasonh:
This looks great! This is honestly even less code than I thought it would be. 👍 

The test seems fine. Maybe worth adding another `.ko.yaml` that has `defaultPlatforms: foo,bar` just to make sure it handles splitting on `,` in the config as we expect.

ReToCode:
I added another test case but not with splitting `,` as viper expects a yaml array to be parsed as a string slice. I also changed the docs accordingly.

imjasonh (dismissed):
Nice, that test covers exactly what I was hoping to find out!

jonjohnsonjr (dismissed):

pkg/commands/options/testdata/multiple-platforms/.ko.yaml
```diff
@@ -0,0 +1,4 @@
+defaultBaseImage: alpine
+defaultPlatforms: 
+- linux/arm64
+- linux/amd64
```
imjasonh, line 4:
```suggestion
- linux/amd64

```

imjasonh (dismissed):

ReToCode:
Seems like there was already an empty line in that file which is not really visible in GitHub.

imjasonh (approved):

imjasonh:
Thanks again for this contribution, and for fighting through the lint checks for it! 😅

PR #902, opened 2022-12-12
Fix SLSA link

BenjaminHerbert:
It did not point to an external site and resulted in a 404.


BenjaminHerbert:
I just saw there already is a PR.

PR #904, opened 2022-12-14
use cgr.dev for base img make the const accessible

developer-guy:

According to the comment [here](https://github.com/goreleaser/goreleaser/pull/3490/files#r1036136408)

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -35,6 +35,8 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/spf13/viper"
+
 	"github.com/containerd/stargz-snapshotter/estargz"
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -57,6 +59,14 @@ import (
 const (
 	defaultAppFilename = "ko-app"
 )
+
+// DefaultBaseImage returns the default base image to use
+func DefaultBaseImage() (ref string, err error) {
+	if !viper.IsSet("defaultBaseImage") {
+		return "", errors.New("'defaultBaseImage' is not set")
+	}
+	return viper.GetString("defaultBaseImage"), nil
+}
 
 // GetBase takes an importpath and returns a base image reference and base image (or index).
 type GetBase func(context.Context, string) (name.Reference, Result, error)
--- a/pkg/commands/options/build.go
+++ b/pkg/commands/options/build.go
@@ -31,7 +31,6 @@ import (
 )
 
 const (
-	// configDefaultBaseImage is the default base image if not specified in .ko.yaml.
 	configDefaultBaseImage = "cgr.dev/chainguard/static:latest"
 )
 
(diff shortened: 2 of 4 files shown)


pkg/commands/options/build.go
```diff
@@ -31,8 +31,8 @@ import (
 )
 
 const (
-	// configDefaultBaseImage is the default base image if not specified in .ko.yaml.
-	configDefaultBaseImage = "distroless.dev/static:latest"
+	// ConfigDefaultBaseImage is the default base image if not specified in .ko.yaml.
+	ConfigDefaultBaseImage = "cgr.dev/chainguard/static:latest"
```
imjasonh, line 35: If it's going to be used externally, `pkg/build.DefaultBaseImage` is better? I don't think I want to recommend other Go code depend on `pkg/commands`. Maybe we should make it `internal/commands` instead to prevent that?

pkg/commands/publisher_test.go
```diff
@@ -72,6 +73,16 @@ func TestPublishImages(t *testing.T) {
 		if err != nil {
 			t.Fatalf("%s: MakeBuilder(): %v", test.description, err)
 		}
+
+		defaultBaseImageRef, err := build.DefaultBaseImage()
```
imjasonh, line 77: This doesn't seem to be used later except just to check that it's set? Can we test that it's a specific expected value?

pkg/build/gobuild.go
```diff
@@ -58,6 +60,20 @@ const (
 	defaultAppFilename = "ko-app"
 )
 
+// DefaultBaseImage returns the default base image us
+func DefaultBaseImage() (ref string, err error) {
+	if !viper.IsSet("defaultBaseImage") {
```
imjasonh, line 65: Does this take into account `KO_CONFIG_PATH`?

pkg/commands/publisher_test.go
developer-guy: yep absolutely!

pkg/build/gobuild.go
developer-guy: yep, I've added a test about that.

developer-guy:
This PR will fix the problem created by #910. Can you please check @halvards? PTAL @imjasonh

pkg/build/gobuild.go
```diff
@@ -58,6 +60,20 @@ const (
 	defaultAppFilename = "ko-app"
 )
 
+// DefaultBaseImage returns the default base image us
```
imjasonh, line 63:
```suggestion
// DefaultBaseImage returns the default base image to use
```

pkg/build/gobuild.go
```diff
@@ -58,6 +60,20 @@ const (
...
+func DefaultBaseImage() (ref string, err error) {
+	if !viper.IsSet("defaultBaseImage") {
+		err = errors.New("'defaultBaseImage' is not set")
+		return
+	}
+	ref = viper.GetString("defaultBaseImage")
+	fmt.Printf("Using default base image from config: %s", ref)
```
imjasonh, line 70:
```suggestion
```

pkg/build/gobuild.go
```diff
@@ -58,6 +60,20 @@ const (
...
+	if _, e := name.ParseReference(ref); e != nil {
+		err = fmt.Errorf("'defaultBaseImage': error parsing %q as image reference: %w", ref, e)
+	}
+	return
+}
```
imjasonh, line 75:
```suggestion
func DefaultBaseImage() (string, error) {
	if !viper.IsSet("defaultBaseImage") {
		return "", errors.New("'defaultBaseImage' is not set")
	}
	return viper.GetString("defaultBaseImage"), nil
}
```

developer-guy:
> This PR will fix the problem created by #910. Can you please check @halvards? PTAL @imjasonh

Quick update here, @imjasonh fixed the problem in PR #923, so I've updated my PR to revert changes related to that fix. ☝️

PR #908, opened 2022-12-16
Add --all option to build command

### LINKED ISSUE #643: Idea: build all images defined in .ko.yaml

carlpett:
This PR adds an `--all` flag to the `build` command, allowing to build all targets specified in `.ko.yaml`. 

Fixes #643

### DIFF
--- a/docs/reference/ko_build.md
+++ b/docs/reference/ko_build.md
@@ -42,6 +42,7 @@ ko build IMPORTPATH... [flags]
 ### Options
 
 ```
+      --all                      When given, build all configured targets.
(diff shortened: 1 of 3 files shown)


carlpett:
Hm, it seems I was a bit too quick on this. `BuildOptions` is used in pretty much all commands. I'm not sure `--all` makes sense in many of those contexts, though.
Would it be best to add a new flag set specifically for `build` that isn't reused between commands, or is there another way forward here?

Issue #909, opened 2022-12-18
Nested config doesn't resolve build path as expected


bluebrown:
Hi, I'm struggling with using a nested config. It doesn't resolve the build path as expected and therefore the build options are not used. 

I created a minimal reproducible example here: https://github.com/bluebrown/ko-bug-nested-config

The expected output in this project is: `tag is used` but I didn't manage to get the tag to be used ko.

PR #910, opened 2022-12-19
Explain that KO_CONFIG_PATH should be a directory

halvards:
Related: #909

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,7 +4,8 @@
 
 Aside from `KO_DOCKER_REPO`, you can configure `ko`'s behavior using a
 `.ko.yaml` file. The location of this file can be overridden with
-`KO_CONFIG_PATH`.
+`KO_CONFIG_PATH`, with the value set to the directory that contains the
+`.ko.yaml` file to be used.
 
 ### Overriding Base Images
 


imjasonh:
https://github.com/ko-build/ko/pull/731 changed `KO_CONFIG_PATH` to allow specifying the file. Is there a bug in that logic?

halvards:
It appears so. Repro steps:

```sh
$ ko version
0.12.0

$ git clone https://github.com/bluebrown/ko-bug-nested-config.git
$ cd ko-bug-nested-config
$ git checkout a07be3e7471a7cc2931cb45d2b52000d52a6557c
$ sed -i 's/dir:.*$/dir: ./' nested/.ko.yaml

$ export KO_CONFIG_PATH="$PWD/nested/.ko.yaml"
$ docker run --rm "$(ko build -L -B ./cmd/app/)"
tag is not used

$ export KO_CONFIG_PATH="$PWD/nested"
$ docker run --rm "$(ko build -L -B ./cmd/app/)"
tag is used
```

developer-guy:
> #731 changed `KO_CONFIG_PATH` to allow specifying the file. Is there a bug in that logic?

Yep, there is a bug related to that because it does not consider the file you supply in the `KO_CONFIG_PATH` because of the viper logic you provide as `v.AddConfigPath(override)`; if the value of the `override` variable is a file, it doesn't consider using the content of it while running  `v.ReadInConfig().` I've faced the same issue while working on #904 ☝️

imjasonh:
I believe handling of `KO_CONFIG_PATH` pointing to a file is handled correctly now, so I'm closing this. If folks would like to improve docs or find bugs please feel free to comment here.

PR #913, opened 2022-12-21
fix: possible race condition when applying templates to flags/ldflags

caarlos0:
while working on https://github.com/goreleaser/goreleaser/pull/3653, I got some race conditions sometimes.

Seems to happen when the base image is a image index, in which case it'll build for all platform needed concurrently, which is when we get the concurrent writes.

I don't see a big issue in returning a copy instead, not sure what you think.

PS: Its late here, so please pardon me if I'm doing something dumb 😂

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -699,22 +699,23 @@ func createTemplateData() map[string]interface{} {
 	}
 }
 
-func applyTemplating(list []string, data map[string]interface{}) error {
-	for i, entry := range list {
+func applyTemplating(list []string, data map[string]interface{}) ([]string, error) {
+	result := make([]string, len(list), 0)
+	for _, entry := range list {
 		tmpl, err := template.New("argsTmpl").Option("missingkey=error").Parse(entry)
 		if err != nil {
-			return err
+			return nil, err
 		}
 
 		var buf bytes.Buffer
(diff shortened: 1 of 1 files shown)


caarlos0:
giving it a second though, a question: when the base image is not multi-platform, should ko still try to build for all the given platforms?

imjasonh:
> giving it a second though, a question: when the base image is not multi-platform, should ko still try to build for all the given platforms?

`ko`'s behavior here is kind of weird, since there are a number of scenarios it has to cover.

1. If the base image is a single-platform image, `--platform` basically only describes the platform to set on the output image. `KO_DEFAULTBASEIMAGE=some-amd64-image ko build --platform=linux/arm64` even works, and adds an arm64 Go binary on top of the amd64 base. 🙃 
1. `--platform=all` when building on a single-platform base just means "whatever the base image's platform is", which is nearly always `linux/amd64`, and might just always default to that (I haven't tried; there really aren't a lot of single-platform non-amd64 images).

I think we should keep this behavior (at least `(2)`), because of how it interacts with `ko resolve` where there may be many builds and many base images. You'd want `ko resolve --platform=all` to produce multi-arch images when possible, and single-arch images when that's all that's available in the base.

imjasonh (approved):

caarlos0:
ah cool, thanks for explaining all that, @imjasonh!

Cheers!

PR #917, opened 2022-12-28
s/google/ko-build/g

mattn:


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/google/ko
+module github.com/ko-build/ko
```
imjasonh, line 1: We'd like to take this opportunity to rename the importpath to ko.build, mostly for vanity, but also so we can move the repo in the future without this kind of change.

This means most of this change will need to wait for that, which is currently blocked on transferring the domain to CNCF ownership. (Which is blocked on me starting that process...)

Sorry this wasn't clearer, and thanks for taking the initiative to make this change. It'll happen in time.

mattn: Thanks. Will you fix this your self? If so, I'll close this PR.

PR #929, opened 2023-01-13
add docs for setup-ko to ko.build

imjasonh:


docs/github-action.md
```diff
@@ -0,0 +1,108 @@
...
+
+See [documentation](./configuration.md) to learn more about configuring `ko`.
+
+The action works on Linux and macOS [runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).
+
+## Select `ko` version to install
+
+By default, `setup-ko` installs the [latest release of `ko`](../releases).
+
+You can specify a specific release version:
```
jonjohnsonjr, line 39: I tripped over the wording of "specify a specific"

s/specify/configure/ or s/specific//

docs/github-action.md
```diff
@@ -0,0 +1,108 @@
...
+
+```yaml
+- uses: ko-build/setup-ko@v0.6
+  with:
+    version: tip
+```
+
+## Pushing to other locations
+
+By default, `setup-ko` [configures `ko`](./configuration.md) to push images to [GitHub Container Registry](https://ghcr.io), but you can configure it to push to other registries as well.
```
jonjohnsonjr, line 57: This is the third time we mention this, I'd drop the first or second one maybe?

docs/github-action.md
```diff
@@ -0,0 +1,108 @@
...
+
+If `KO_DOCKER_REPO` is already set when `setup-ko` runs, it will skip logging in to ghcr.io and will propagate `KO_DOCKER_REPO` for subsequent steps.
+
+You will also need to provide credentials to authorize the push. You can use [encrypted secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) to store the authorization token, and pass it to `ko login` before pushing:
```
jonjohnsonjr, line 61:
```suggestion
You will also need to provide credentials to authorize the push. You can use [encrypted secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) to store the authorization token and pass it to `ko login` before pushing:
```

Issue #926, opened 2023-01-12
Make `go` and `kubectl` version configurable


embano1:
Currently, `ko` uses hardcoded values to retrieve the `go` and `kubectl` executables, e.g.

https://github.com/ko-build/ko/blob/453bf803e379696a0b9142c772402ba4599cff34/pkg/build/gobuild.go#L288

The desired behavior is to make those binary locations configurable, e.g. via `KO_GO_PATH` and `KO_KUBECTL_PATH`.
Alternatively, they could be specified in `.ko.yaml` but this might be overkill as setting these values are considered exceptional.

imjasonh:
Let's start with `KO_GO_PATH`. AFAIK nobody's asked to be able to change the kubectl path. The naming is convenient because if we want to make it configurable in `.ko.yaml` we can just add a `go_path` field and Viper will pick up `KO_GO_PATH` for us.

PR #930, opened 2023-01-13
feat: Add KO_GO_PATH env var

embano1:
Makes `go` version used to build image binaries configurable via `KO_GO_PATH`. Optional, defaults to `go` resolved from `$PATH`.

Closes: #926
Signed-off-by: Michael Gasch <15986659+embano1@users.noreply.github.com>


pkg/build/gobuild.go
```diff
@@ -56,6 +56,9 @@ import (
 
 const (
 	defaultAppFilename = "ko-app"
+
```
embano1, line 59: I typically add these for readability. Let me know whether to remove.

pkg/commands/deps.go
```diff
@@ -34,6 +34,11 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
```
embano1, line 37: Had to replicate these to avoid imports.

docs/configuration.md
```diff
@@ -106,6 +104,18 @@ You can also use the `KO_DEFAULTPLATFORMS` environment variable to set the defau
 KO_DEFAULTPLATFORMS=linux/arm64,linux/amd64
 ```
 
+### Environment Variables (advanced)
```
embano1, line 107: Added these to provide more clarity.

pkg/build/gobuild.go
```diff
@@ -285,7 +288,12 @@ func build(ctx context.Context, ip string, dir string, platform v1.Platform, con
 
 	args = append(args, "-o", file)
 	args = append(args, ip)
-	cmd := exec.CommandContext(ctx, "go", args...)
+
+	gobin := defaultGoBin
```
imjasonh, line 292: This is repeated enough maybe should have it be a buildOption, WithGoBinary("gotip")?

pkg/commands/deps.go
imjasonh: This feels like enough reason to finally delete `ko deps`, which has been hidden and deprecated for a while.

embano1: Ah, didn't know! Shall I do this in this PR or follow up?

pkg/build/gobuild.go
embano1: Are you referring to this type def `BuildOptions` or creating a new one and make it a functional options parameter to those two functions so I don't have to change the whole call trees/callers? I would still have to check for set build option in those cases then unless I propagate a default (requires changing the call tree).

Without `pkd/commands.go` we have three copies of this code snippet currently.

pkg/commands/deps.go
embano1: Update: removed

pkg/build/gobuild.go
embano1: I simplified to a helper function, ok?

imjasonh (dismissed):

imjasonh:
Sorry for complicating this PR by including `ko deps`' removal in it. I've split that out into https://github.com/ko-build/ko/pull/937

embano1:
Ah, let me wait until your PR is in and then rebase.

embano1:
Ready after rebase @imjasonh

docs/configuration.md
```diff
@@ -106,6 +104,18 @@ You can also use the `KO_DEFAULTPLATFORMS` environment variable to set the defau
...
+
+| Variable         | Default Value                              | Description                                                                                                                                                                                                                      |
+|------------------|--------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `KO_DOCKER_REPO` | (not set)                                  | Container repository where to push images built with `ko` (required)                                                                                                                                                             |
+| `KO_DATA_PATH`   | `/var/run/ko` (`C:\var\run\ko` on Windows) | By convention, any contents of a directory named `<importpath>/kodata/` will be bundled into the image, and the path where it's available in the image will be identified by the environment variable `KO_DATA_PATH` (optional). |
```
imjasonh, line 115:
```suggestion
```

Let's remove this, since it's a bit different from the others, in that it's only involved at runtime.

imjasonh (approved):

imjasonh (approved):

PR #932, opened 2023-01-16
safeguardings against empty tags

### LINKED ISSUE #931: fix the message in defalt publisher to avoid having misunderstanding
@caarlos0 first noticed this problem from the PR that we are working on integrating ko with the GoReleaser, [here](https://github.com/goreleaser/goreleaser/pull/3653#issuecomment-1383443480). 

[defalt](https://github.com/ko-build/ko/blob/453bf803e379696a0b9142c772402ba4599cff34/pkg/publish/default.go) publisher uses [if condition to push the result by looking at the lengths of the tags](https://github.com/ko-build/ko/blob/453bf803e379696a0b9142c772402ba4599cff34/pkg/publish/default.go#L214) that are going to be published, but if there are no tags defined, it does not push anything, but prints a log message stating that it was successfully published.

developer-guy:

Fixes #931

### DIFF
--- a/pkg/publish/default.go
+++ b/pkg/publish/default.go
@@ -204,6 +204,12 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
+	// safeguarding against the case where the user has specified an empty tag with publish.WithTags([]string{""}) or publish.WithTags([]string{})
+	if len(d.tags) == 0 || (len(d.tags) == 1 && d.tags[0] == "") {
+		log.Printf("No tags specified, using default tag: %s", latestTag)
+		d.tags = []string{latestTag}
+	}
+
 	ro := []remote.Option{remote.WithAuthFromKeychain(d.keychain), remote.WithTransport(d.t), remote.WithContext(ctx), remote.WithUserAgent(d.userAgent)}
 
 	no := []name.Option{}
--- a/pkg/publish/default_test.go
+++ b/pkg/publish/default_test.go
@@ -85,6 +85,49 @@ func TestDefault(t *testing.T) {
 	}
 }
 
+func TestDefaultWithEmptyTag(t *testing.T) {
+	f, err := static.NewFile([]byte("da bom"))
+	if err != nil {
+		t.Fatalf("static.NewFile() = %v", err)
+	}
+	si, err := ocimutate.AttachFileToImage(signed.Image(img), "sbom", f)
+	if err != nil {
+		t.Fatalf("ocimutate.AttachFileToImage() = %v", err)
+	}
+	sii, err := ocimutate.AttachFileToImageIndex(signed.ImageIndex(idx), "sbom", f)
+	if err != nil {
+		t.Fatalf("ocimutate.AttachFileToImageIndex() = %v", err)
+	}
+
+	for _, br := range []build.Result{img, idx, si, sii} {
+		base := "blah"
+		importpath := "github.com/Google/go-containerregistry/cmd/crane"
+		expectedRepo := fmt.Sprintf("%s/%s", base, strings.ToLower(importpath))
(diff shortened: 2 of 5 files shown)


caarlos0 (dismissed):

pkg/publish/default_test.go
```diff
@@ -85,6 +85,49 @@ func TestDefault(t *testing.T) {
...
+		repoName := fmt.Sprintf("%s/%s", u.Host, base)
+		def, err := publish.NewDefault(repoName, publish.WithTags([]string{}))
+		if err != nil {
+			t.Errorf("NewDefault() = %v", err)
+		}
+		if d, err := def.Publish(context.Background(), br, build.StrictScheme+importpath); err != nil {
+			t.Errorf("Publish() = %v", err)
+		} else if !strings.HasPrefix(d.String(), tag.Repository.String()) {
+			t.Errorf("Publish() = %v, wanted prefix %v", d, tag.Repository)
+		}
```
imjasonh, line 127: This only checks that the repository component is as expected, and doesn't check that the image is tagged with `:latest` when `WithTags` is passed this way.

pkg/publish/tarball.go
```diff
@@ -51,6 +51,12 @@ func (t *tar) Publish(_ context.Context, br build.Result, s string) (name.Refere
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
+	// safeguarding against the case where the user has specified an empty tag with publish.WithTags([]string{""})
+	if len(t.tags) == 0 {
+		log.Printf("No tags specified, using default tag: %s", latestTag)
+		t.tags = []string{latestTag}
+	}
```
imjasonh, line 58: The comment here and elsewhere doesn't match the code. `[]string{""}` is different from `[]string{}`; the first has `len(tags) == 1`, the second has `len == 0` and matches the code.

I'm not sure what happens if passed `WithTags([]string{""})` or for that matter `WithTags([]string{"", "", "", "latest", ""})`. It's probably worth finding out and fixing any confusion that might cause. Doesn't need to block this PR though.

pkg/publish/default_test.go
developer-guy: The following code creates a tag with `latest`:

```go
tag, err := name.NewTag(fmt.Sprintf("%s/%s:latest", u.Host, expectedRepo))
```

And in the `else if` block, we check if the tag of the published image is equal to the tag created by giving `latest.`. 

Am I wrong?

pkg/publish/tarball.go
developer-guy: > The comment here and elsewhere doesn't match the code. []string{""} is different from []string{}; the first has len(tags) == 1, the second has len == 0 and matches the code.

My bad 🤦 I've fixed it.

pkg/publish/default_test.go
developer-guy: kindly ping @imjasonh

pkg/publish/default.go
```diff
@@ -204,6 +204,12 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 	// https://github.com/google/go-containerregistry/issues/212
 	s = strings.ToLower(s)
 
+	// safeguarding against the case where the user has specified an empty tag with publish.WithTags([]string{""}) or publish.WithTags([]string{})
+	if len(d.tags) == 0 || (len(d.tags) == 1 && d.tags[0] == "") {
+		log.Printf("No tags specified, using default tag: %s", latestTag)
+		d.tags = []string{latestTag}
+	}
```
imjasonh, line 211: What do we do today if someone manages to pass empty tags? It might be expected that we publish images by digest and don't tag anything.

If we don't do that today then I'm fine changing that so that callers (presumably via the Go package, and not the CLI) default to using the `:latest` tag. If there's any way to call the CLI so that things end up published but untagged, we should probably retain that behavior. And document it.

PR #964, opened 2023-02-23
enable preparer interface to pre operations

### LINKED ISSUE #539: Check for push permissions before building
If I'm `ko publish`ing to a registry/repository that I don't have push permissions to (typo, misconfigured auth, incorrect active `gcloud` user), `ko` can check that and fail before building, using ggcr's [`CheckPushPermission`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/remote#CheckPushPermission).

That has a `TODO`(https://github.com/google/go-containerregistry/issues/412) to make it unnecessary, but I think this is actually a case where it's useful, since it can avoid work and save time before `remote.Write`ing.

developer-guy:
fixes #539 also duplicates #951 

PTAL @imjasonh

### DIFF
--- a/pkg/publish/default.go
+++ b/pkg/publish/default.go
@@ -260,3 +260,25 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
 func (d *defalt) Close() error {
 	return nil
 }
+
+// Prepare implements publish.Preparer
+func (d *defalt) Prepare(ctx context.Context, s string) error {
+	var no []name.Option
+	if d.insecure {
+		no = append(no, name.Insecure)
+	}
+
+	for _, tagName := range d.tags {
+		tag, err := name.NewTag(fmt.Sprintf("%s:%s", d.namer(d.base, s), tagName), no...)
+		if err != nil {
+			return err
+		}
+
+		if err := remote.CheckPushPermission(tag, d.keychain, d.t); err != nil {
+			return err
+		}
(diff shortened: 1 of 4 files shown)


pkg/publish/default.go
```diff
@@ -260,3 +260,25 @@ func (d *defalt) Publish(ctx context.Context, br build.Result, s string) (name.R
...
+	for _, tagName := range d.tags {
+		tag, err := name.NewTag(fmt.Sprintf("%s:%s", d.namer(d.base, s), tagName), no...)
+		if err != nil {
+			return err
+		}
+
+		if err := remote.CheckPushPermission(tag, d.keychain, d.t); err != nil {
+			return err
+		}
+	}
```
imjasonh, line 280: Auth is generally configured at the repo level, and all these tags will have the same repo, so we can just check the first one, or `:latest`.

It looks like `CheckPushPermission` should have just taken a `name.Repository` anyway, it doesn't use the tag: https://github.com/google/go-containerregistry/blob/v0.13.0/pkg/v1/remote/check.go#L34

So let's just check the first tag and not bother checking each one.

imjasonh (commented):
Looks good so far! Honestly this is a smaller change than I expected! 🎉

pkg/publish/default.go
developer-guy: that makes sense.

PR #974, opened 2023-03-07
fix file extension for cyclonedx

developer-guy:
PTAL @imjasonh

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -412,7 +412,7 @@ func cycloneDX() sbomber {
 				return nil, "", err
 			}
 
-			if err := writeSBOM(b, appFileName, dir, "cyclone.json"); err != nil {
+			if err := writeSBOM(b, appFileName, dir, "cyclonedx.json"); err != nil {
 				return nil, "", err
 			}
 


imjasonh:
Is this incorrect as specified today? Is there some code that expects it to be named cyclonedx.json specifically?

developer-guy:
> Is this incorrect as specified today? 
yes, this is a typo.

> Is there some code that expects it to be named cyclonedx.json specifically?
nope, this is just a convention we define to specify the SBOM type within the file

imjasonh:
I guess the problem is, now, someone somewhere might be depending on the `cyclone.json` that `ko` produces, and changing it, even if it's to match the convention, might break them. There's no way to know.

I'm fine making the change, to be clear. `ko` is pre-1.0, and makes no stability guarantees. Probably nobody uses cyclonedx SBOMs written to disk anyway, since that's not the default behavior. But we should take care not to risk breaking folks unless there's a clear benefit. Maybe matching the convention is enough benefit?

developer-guy:
> Probably nobody uses cyclonedx SBOMs written to disk anyway,

I agree because we didn't release that support yet either, this is why I'm not concerned about breaking anything unless someone uses ko from its main.

> Maybe matching the convention is enough benefit?

yes because this is not the correct naming for the file, we made a typo and we should fix this.

Btw, thanks for bringing these concerns to my attention!

imjasonh (approved):

PR #975, opened 2023-03-08
fix: handle docker's unknown/unknown platform in index manifests

imjasonh:
Docker decided to include provenance attestations in index manifests, with descriptors having the platform `unknown/unknown`: https://docs.docker.com/build/attestations/attestation-storage/#image-index-sha25694acc2ca70c40f3f6291681f37ce9c767e3d251ce01c7e4e9b98ccf148c26260

This causes problems for `ko build --platform=all` since `unknown/unknown` is not a, well, known GOOS/GOARCH, in accordance with the [OCI image spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md#image-index-property-descriptions).

This change ignores `unknown/unknown` when trying to match a multi-arch base image's platforms, even when using `--platform=all`.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -1171,6 +1171,13 @@ func parseSpec(spec []string) (*platformMatcher, error) {
 }
 
 func (pm *platformMatcher) matches(base *v1.Platform) bool {
+	// Strip out manifests with "unknown/unknown" platform, which Docker uses
+	// to store provenance attestations.
+	if base != nil &&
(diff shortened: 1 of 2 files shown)


jonjohnsonjr (approved):
This LGTM, but we should consider looking at what the go tool thinks are valid GOOS and GOARCH combinations per [the spec](https://github.com/opencontainers/image-spec/blob/main/image-index.md#:~:text=MUST%20be%20ignored.-,platform,-object).

developer-guy:
this is so cool!

PR #977, opened 2023-03-10
fix release workflow

imjasonh:
This breaks because the release doesn't have a built `ko` binary yet.

This only runs `publish` after the `goreleaser` job has finished and binaries are available.

I think we could also just remember not to check "this is the latest release" when cutting a new release, but I will never remember to do that.

For this release I'm just going to retry the failed job after `goreleaser` finishes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
           tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
           img=$(ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./)
           echo "built ${img}"
-          cosign sign ${img} \
+          cosign sign ${img} --yes \
               -a sha=${{ github.sha }} \
               -a run_id=${{ github.run_id }} \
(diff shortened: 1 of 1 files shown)


developer-guy (approved):
lgtm

.github/workflows/release.yml
```diff
@@ -50,7 +51,7 @@ jobs:
           tag=$(echo ${{ github.ref }} | cut -c11-) # get tag name without tags/refs/ prefix.
           img=$(ko build --bare --platform=all -t latest -t ${{ github.sha }} -t ${tag} ./)
           echo "built ${img}"
-          cosign sign ${img} \
+          cosign sign ${img} --yes \
```
developer-guy, line 54: what do you think about using the `COSIGN_YES` variable rather than using the `--yes` flag? 🤔

imjasonh: Is there a reason to prefer one over the other? I suppose if we add a second `cosign` invocation here we'd also need to remember to pass `--yes` to that as well.

cpanato (approved):

jonjohnsonjr (approved):

PR #985, opened 2023-03-14
add required flag when using cosign v2

cpanato:
add required flag when using cosign v2


cpanato:
closing in favor of https://github.com/ko-build/ko/pull/977

PR #1015, opened 2023-04-19
stop estargz-optimizing

imjasonh:
This didn't do anything unless `GGCR_EXPERIMENTAL_ESTARGZ` was also set, but if it was it would tell estargz to prioritize fetching the app binary.

In light of https://github.com/google/go-containerregistry/pull/1660 we don't want to use this deprecated method.


evankanderson (approved):

imjasonh:
Fixed in #1142

PR #1027, opened 2023-04-27
Fix kind image names with --bare

aidy:
Make the functionality match between:
KO_DOCKER_REPO=ko.local/my-image ko build --bare

And:
KO_DOCKER_REPO=kind.local/my-image ko build --bare

As it stands, `--bare` is broken with `kind.local/` as it just gets `kind.local` as the image name, and then it fails to tag because the registry is missing.

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -199,7 +199,7 @@ func makePublisher(po *options.PublishOptions) (publish.Interface, error) {
 			)
 		}
 		if strings.HasPrefix(repoName, publish.KindDomain) {
-			return publish.NewKindPublisher(namer, po.Tags), nil
+			return publish.NewKindPublisher(repoName, namer, po.Tags), nil
 		}
 
 		if repoName == "" && po.Push {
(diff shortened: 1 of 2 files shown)


nitinmohan87:
@imjasonh Any chances this PR can be reviewed/merged?

cpanato (commented):
is it possible to add some tests?

aidy:
@cpanato  It's been nearly a year since I raised this, but I'll try to remember how it hangs together.

aidy:
I don't think it's possible to write a useful test for this without either instantiating a new kind cluster (making tests much more heavyweight), or by doing some fairly hefty refactoring to expose some of the kind publisher internals to allow mock kinds to be injected from other packages.

aidy:
> or by doing some fairly hefty refactoring to expose some of the kind publisher internals to allow mock kinds to be injected from other packages.

Would look something like: https://github.com/aidy/ko/compare/fix_kind_image_names...aidy:ko:fix_kind_image_names_test

Can add to this PR if you'd like, but obviously it does open up some internals.

nitinmohan87:
Thanks for reopening the PR and getting it ready @aidy  and @cpanato ! Any thoughts on getting this merged?

imjasonh (approved):

PR #1038, opened 2023-05-05
support building from a scratch image

### LINKED ISSUE #207: support "scratch" as a special base image
Something like `ko.yaml`:
```
defaultBaseImage: scratch
```

tzneal:
Fixes #207

### DIFF
--- a/pkg/build/scratch.go
+++ b/pkg/build/scratch.go
@@ -1,0 +1,62 @@
+// Copyright 2023 ko Build Authors All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package build
+
+import (
+	"fmt"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+// ScratchImage returns a scratch image manifest with scratch images for each of the specified platforms
+func ScratchImage(platforms []string) (Result, error) {
+	var manifests []mutate.IndexAddendum
+	for _, pf := range platforms {
+		if pf == "all" {
+			return nil, fmt.Errorf("'all' is not supported for building a scratch image, the platform list must be provided")
+		}
+		p, err := v1.ParsePlatform(pf)
+		if err != nil {
+			return nil, err
+		}
+		img, err := mutate.ConfigFile(empty.Image, &v1.ConfigFile{
+			RootFS: v1.RootFS{
+				// Some clients check this.
+				Type: "layers",
+			},
+			Architecture: p.Architecture,
+			OS:           p.OS,
+			Variant:      p.Variant,
+			OSVersion:    p.OSVersion,
+			OSFeatures:   p.OSFeatures,
+		},
+		)
+		if err != nil {
+			return nil, fmt.Errorf("setting config file on empty image, %w", err)
+		}
+		manifests = append(manifests, mutate.IndexAddendum{
+			Add: img,
+			Descriptor: v1.Descriptor{
+				Platform: p,
+			},
+		})
+	}
+	idx := mutate.IndexMediaType(empty.Index, types.OCIImageIndex)
+	idx = mutate.AppendManifests(idx, manifests...)
+	return idx, nil
+}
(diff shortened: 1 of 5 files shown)


pkg/build/scratch.go
```diff
@@ -0,0 +1,168 @@
...
+
+var _ partial.UncompressedImageCore = (*scratchImageContents)(nil)
+var _ partial.WithConfigFile = (*scratchImageContents)(nil)
+var _ v1.ImageIndex = (*scratchImageIndex)(nil)
+
+// ScratchImage returns a scratch image manifest with scratch images for each of the specified platforms
+func ScratchImage(platforms []string) (Result, error) {
+	var pfs []v1.Platform
+	for _, pf := range platforms {
+		fields := strings.Split(pf, "/")
```
imjasonh, line 39: You can use [`ParsePlatform`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1#ParsePlatform) to do this for you.

pkg/build/scratch.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/partial"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+)
+
+var _ partial.UncompressedImageCore = (*scratchImageContents)(nil)
+var _ partial.WithConfigFile = (*scratchImageContents)(nil)
+var _ v1.ImageIndex = (*scratchImageIndex)(nil)
+
+// ScratchImage returns a scratch image manifest with scratch images for each of the specified platforms
+func ScratchImage(platforms []string) (Result, error) {
```
imjasonh, line 36:
```suggestion
func scratchImage(platforms []string) (Result, error) {
```

I don't think we should export this method so that external projects could use it.

imjasonh (commented):
You might have a better time using [`pkg/v1/empty`](https://pkg.go.dev/github.com/google/go-containerregistry/pkg/v1/empty) to do this:

```go
idx := mutate.IndexMediaType(empty.Index, types.OCIImageIndex)
for _, p := range platforms {
  plat, err := v1.ParsePlatform(p)
  // handle err
  img := mutate.MediaType(
    mutate.ConfigMediaType(empty.Image, types.OCIConfigJSON),
    types.OCIManifestSchema1,
  )
  idx = mutate.AppendManifests(idx, mutate.IndexAddendum{
    Add: img,
    Descriptor: v1.Descriptor{
      Platform: plat,
    },
  })
}
return idx, nil
```

This will handle most of the manual hash handling for you, and should be usable has a multi-arch zero-layer base image.

I wonder what we should do when the base is `scratch` and `--platform=all` 🤔

pkg/build/scratch.go
tzneal: Nice, I've never looked at this codebase before so I'm just fumbling around 😆

pkg/build/scratch.go
tzneal: I'd need to move it to the commands package if its not exported. WDYT?

tzneal:
All failed to parse with the platform parsing code, I added an explicit error for it now.  I concluded it didn't make sense for a scratch image case, and you can work around it by just providing all platforms if you really do want them.

pkg/build/scratch.go
jonjohnsonjr: That's fine. If we need it in more than one package, we can move it to `pkg/internal/something`.

pkg/commands/scratch_test.go
```diff
@@ -0,0 +1,66 @@
...
+		t.Fatalf("expected to have an image index")
+	}
+	mf, err := imgIdx.IndexManifest()
+	if mt != expMT {
+		t.Errorf("expected a manifest, got %s", err)
+	}
+	if len(mf.Manifests) != 2 {
+		t.Fatalf("expected two manifests, got %d", len(mf.Manifests))
+	}
+	for _, m := range mf.Manifests {
```
imjasonh, line 51: Could we also check that the images have 2 layers, both added by us?

```
img, err := imgIdx.Image(m)
ls, err := img.Layers()
if len(ls) != 2 {
  t.Errorf(...)
}
```

imjasonh (commented):
This looks really good! Thanks for doing this.

pkg/commands/scratch_test.go
tzneal: Sounds good, I ended up making `ScratchImage` exported so I could build the image and verify that it only has two base layers.  I also ensure that the scratch base image has zero layers now.

tzneal:
@imjasonh Was there something else remaining for this?

imjasonh:
Sorry for taking so long to get back to this. This change still looks good, and thanks for adding a docs section as well.

I think it would be good to have an e2e test in https://github.com/ko-build/ko/blob/main/.github/workflows/e2e.yaml that does `KO_DEFAULTBASEIMAGE=scratch ko build --platform=linux/amd64,linux/s390x` and checks that the resulting image is as expected (using `crane` and `jq` probably). I'd be happy to help with that if you're still interested. If not I can do it as a followup.

Thanks again for making this change, and sorry again for the delay.

tzneal:
>   If not I can do it as a followup.

I'd be happy for you to take over the e2e test part.  I'm swamped with some other work and certain that you can do it faster than me :laughing:

cprivitere:
Can this be re-opened? I think it was only closed because it was stale, not because nobody wants it.

cprivitere:
Dunno what your current situation is @imjasonh  maybe you can do the e2e stuff if you have time now?

PR #1048, opened 2023-05-15
Minor cleanup with `ldflags` quoting

briantopping:
When there are multiple `ldflags` defined, they will not be properly handled without a set of quotes around all of them.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -757,7 +757,13 @@ func createBuildArgs(buildCfg Config) ([]string, error) {
 			return nil, err
 		}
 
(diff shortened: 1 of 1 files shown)


briantopping:
CC @imjasonh

imjasonh:
This change looks correct, but I'm wondering if you can provide an example of how this worked before this change, and what it's intended to prevent? As I understand it if you pass ldflags this way, `ko` would pass them to `go`, which would likely fail. Is the idea that we can detect this case and provide a better error message than `go` does?

briantopping:
Thanks @imjasonh. I believe the existing code would work properly with`ldflags` that did not include spaces. In other words, a single flag. Alternately, only the first flag might be interpreted and the rest ignored. I haven't exhaustively checked it.

Issue #1096, opened 2023-07-16
Documentation mistake


4lch4:
I'm new to Go overall (started learning it about a month ago) and have come across Ko while looking into Go Releaser. While going through the documentation I noticed one of the steps are no longer correct, specifically the bit [on attestations][0]:

```bash
$ curl -sSfL "https://github.com/ko-build/ko/releases/download/v${VERSION}/ko_${VERSION}_${OS}_${ARCH}.tar.gz" > ko.tar.gz
$ curl -sSfL https://github.com/ko-build/ko/releases/download/v${VERSION}/attestation.intoto.jsonl > provenance.intoto.jsonl
$ slsa-verifier -artifact-path ko.tar.gz -provenance provenance.intoto.jsonl -source github.com/google/ko -tag "v${VERSION}"
  PASSED: Verified SLSA provenance
```

The 2nd line specifies an `attestation.intoto.jsonl` file but it no longer exists as of two releases ago, so these lines fail during installation. Here's the output from my attempt as an example:

```bash
dleaman at Devins-MBP in ~/sw 
$ VERSION=0.14.1

dleaman at Devins-MBP in ~/sw 
$ OS=Darwin                                    

dleaman at Devins-MBP in ~/sw 
$ ARCH=arm64                                                                                                                                                                                       

dleaman at Devins-MBP in ~/sw 
$ curl -sSfL "https://github.com/ko-build/ko/releases/download/v${VERSION}/ko_${VERSION}_${OS}_${ARCH}.tar.gz" > ko.tar.gz

dleaman at Devins-MBP in ~/sw 
$ curl -sSfL "https://github.com/ko-build/ko/releases/download/v${VERSION}/attestation.intoto.jsonl" > provenance.intoto.jsonl
curl: (22) The requested URL returned error: 404
```

This caused me to look at the releases and I noticed the `attestation.intoto.jsonl` file was renamed to `multiple.intoto.jsonl` for the [`v0.13.0` release][1].

I'd be happy to help update this if needed, I just wanted to submit an issue before modifying any files.

[0]: https://ko.build/install/#install-from-github-releases
[1]: https://github.com/ko-build/ko/releases/tag/v0.13.0

cpanato:
thanks for the issue, yes that is correct the name changed to `multiple.intoto.jsonl` so we can update the docs to say in version `0.13.x` is called `attestation.intoto.jsonl` and `>0.14` is `multiple.intoto.jsonl`


[...]

PR #1099, opened 2023-07-18
refactor: fixed install docs attestation

4lch4:
This PR was created as a result of #1096. @cpanato asked me to submit a PR that would fix the small documentation error I uncovered so here it is.

This PR updates the installation docs (`docs/install.md`) on how to install using GitHub Releases. Previously, the steps would encounter a 404 error when attempting to download the `attestation.intoto.jsonl` because the file was renamed to `multiple.intoto.jsonl` in the `v0.13.0` release. Now, the documentation mentions that if you're using any version before `v0.13.0` then use the previous filename but any version after (and including) `v0.13.0` must use the `multiple.intoto.jsonl` filename instead.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -8,8 +8,23 @@ $ OS=Linux     # or Darwin
 $ ARCH=x86_64  # or arm64, i386, s390x
 ```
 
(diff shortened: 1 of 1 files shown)


cpanato (commented):
lgtm

thank you

cpanato (approved):

PR #1102, opened 2023-07-27
implement dumb cache for images

jonjohnsonjr:
This speeds up repeated builds by caching manifests and config files locally under `${KOCACHE}/img`.
Before:

```
ko build .  1.09s user 2.37s system 148% cpu 2.326 total
```

After:

```
ko build .  1.05s user 2.29s system 244% cpu 1.366 total
```

(Most of the time difference comes from not having to do the TLS handshake with DockerHub to fetch the golang stuff.)

This has a nice effect of allowing offline builds with `ko` if your base images are referenced by digest _and_ you pre-pull the images into `${KOCACHE}/img`.

This does not cache base image layers by default because they are often present in the destination registry and would slow things down and take up space.

It might make sense to add a command or flag to completely cache the base image, but we could also just add some docs about how to pull it off with crane, e.g. for ko:

```
$ crane pull --format oci golang:1.20 $KOCACHE/img
```

Then modifying the `.ko.yaml` file to be:

```
baseImageOverrides:
  github.com/google/ko: golang:1.20@sha256:cfc9d1b07b1ef4f7a4571f0b60a99646a92ef76adb7d9943f4cb7b606c6554e2
```

If we `--push=false`, we don't hit the network at all:

```
time ko build -v . --push=false
2023/07/27 14:37:32 maxprocs: Leaving GOMAXPROCS=12: CPU quota undefined
2023/07/27 14:37:32 cache hit: golang:1.20@sha256:cfc9d1b07b1ef4f7a4571f0b60a99646a92ef76adb7d9943f4cb7b606c6554e2
2023/07/27 14:37:32 Using base golang:1.20@sha256:cfc9d1b07b1ef4f7a4571f0b60a99646a92ef76adb7d9943f4cb7b606c6554e2 for github.com/google/ko
2023/07/27 14:37:32 cache hit: index.docker.io/library/golang@sha256:832f2f74baa3e2b00ace688cb2fa934dffeade39f5b4c0cc8b1cff8d3fb084a0
2023/07/27 14:37:32 Using build config ko for github.com/google/ko
2023/07/27 14:37:32 Building github.com/google/ko for linux/amd64
us-central1-docker.pkg.dev/jonjohnson-chainguard/ko/ko-98b8c7facdad74510a7cae0cd368eb4e:latest@sha256:3609cb6560c07a6f2cff054efbf962860e35e89cfa495a85cbf71e71c9e99445
ko build -v . --push=false  0.96s user 2.17s system 573% cpu 0.546 total
```

### DIFF
--- a/pkg/commands/cache.go
+++ b/pkg/commands/cache.go
@@ -1,0 +1,464 @@
+// Copyright 2023 ko Build Authors All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package commands
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/google/go-containerregistry/pkg/logs"
+	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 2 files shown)


cpanato (approved):
wow

PR #1103, opened 2023-07-28
Allow using go test -c as build command

y-tajiri:
https://github.com/ko-build/ko/issues/96
This isssue has been stopped, but I made a preliminary draft.

```
ko build --test import-path
```
will use go test -c instead of go build.

### DIFF
--- a/pkg/build/gobuilds.go
+++ b/pkg/build/gobuilds.go
@@ -32,6 +32,9 @@ type gobuilds struct {
 
 	// workingDirectory is typically ".", but it may be a different value if ko is embedded as a library.
 	workingDirectory string
+
+	// ignore supported reference
+	ignoreSupportedReference bool
 }
 
 // builderWithConfig is not an imaginative name.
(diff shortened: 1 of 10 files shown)


hermanbanken:
Would be very interested in using this. I was just searching the docs of Ko for this, and also searching whether I can get the same kind of output with `go build` as I could with `go test -c`, which apparently is not possible. So having `ko build --test` would REALLY be great!

pkg/build/gobuilds.go
```diff
@@ -32,6 +32,9 @@ type gobuilds struct {
 
 	// workingDirectory is typically ".", but it may be a different value if ko is embedded as a library.
 	workingDirectory string
+
+	// ignore supported reference
+	ignoreSupportedReference bool
```
imjasonh, line 37: This seems to be unused.

imjasonh (commented):
This looks pretty good so far. I'd like to think about whether this warrants a separate `ko test ./pkg/foo` surface. I'm still not sure I understand the use case well enough to have a strong opinion, but I'd love to learn.

Aside from that I think we'll need an e2e test that exercises this path, probably starting by having the `./test` package have a `foo_test.go` in it, and doing a `ko build --test ./test` that runs the test.

We'll also want some docs to describe how this works, why you'd want to use it, etc.

This change as-is looks like a good start, thanks for picking it up so long after the issue went cold.

PR #1109, opened 2023-08-08
pkg/commands: fix dropped errors

alrs:
This picks up two dropped `err` variables in `pkg/commands`.

### DIFF
--- a/pkg/commands/cache.go
+++ b/pkg/commands/cache.go
@@ -124,15 +124,18 @@ func (i *imageCache) get(ctx context.Context, ref name.Reference, missFunc baseF
 					return nil, err
 				}
 				br, err = i.newLazyIndex(ref, idx, missFunc)
+				if err != nil {
+					return nil, err
+				}
 			} else {
 				img, err := ii.Image(h)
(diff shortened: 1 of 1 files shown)


pkg/commands/cache.go
```diff
@@ -124,12 +124,18 @@ func (i *imageCache) get(ctx context.Context, ref name.Reference, missFunc baseF
 					return nil, err
 				}
 				br, err = i.newLazyIndex(ref, idx, missFunc)
+				if err != nil {
```
imjasonh, line 127: I think the intention was that `err` would be checked below on line 140, but this doesn't happen because lines 122 and 131 use `img, err := ` which creates a new scoped `err`. The `err` checked on line 140 is the one we get from line 112.

If we proceed with this approach (I do think it's more readable, and less clever), we should also remove the redundant check on line 140, which doesn't do anything today or with your change.

Thanks for catching this!

alrs: @imjasonh I've removed the redundant check on 140.

imjasonh (approved):

PR #1128, opened 2023-09-08
Allow specifying binary path in image

justinsb:
Previously this was hard-coded to `/ko-app/<app-name>`.

go-releaser allows specifying the binary (including directory prefix);
we now support the same.

### DIFF
--- a/pkg/commands/options/build.go
+++ b/pkg/commands/options/build.go
@@ -78,6 +78,9 @@ type BuildOptions struct {
 	// `AddBuildOptions()` defaults this field to `true`.
 	Trimpath bool
 
+	// BinaryPath overrides the default path for the binary in the output image.
+	BinaryPath string
+
 	// BuildConfigs stores the per-image build config from `.ko.yaml`.
 	BuildConfigs map[string]build.Config
 }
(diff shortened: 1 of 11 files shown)


justinsb:
Opening for discussion, I expect we'll want e.g. some test coverage and docs.

imjasonh:
This change looks pretty good to me. I think we should also support it as a flag in `ko build` (flag name ideas welcome!)

hakman:
@justinsb I see `ko` has `KO_DATA_PATH` so maybe split this into `KO_APP_PATH` and `KO_APP_NAME`? (just a thought to make things more consistent)

justinsb:
> I see ko has KO_DATA_PATH so maybe split this into KO_APP_PATH and KO_APP_NAME? (just a thought to make things more consistent)
I was trying to follow go-releaser here in terms of how this value was provided by the user.  It's not a perfect match because goreleaser also adds goos and goarch, but I think that's primarily to keep things separated in the artifacts directory, and the container image does that naturally.

It sounds like we do want to export the path to the binary for e.g. dlv, which we can add (not sure if a separate PR)

> I think we should also support it as a flag in ko build (flag name ideas welcome!)

I added a flag `--binary`, just as a strawman.  Happy to change the name.

I also added some tests (and note: I think the app layer test was actually checking the data layer, so I tweaked that)

justinsb:
Not stale, needs review.

abh:
Still needs review ...

pkg/commands/options/build.go
```diff
@@ -93,6 +96,8 @@ func AddBuildOptions(cmd *cobra.Command, bo *BuildOptions) {
 		"Path to file where the SBOM will be written.")
 	cmd.Flags().StringSliceVar(&bo.Platforms, "platform", []string{},
 		"Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]][,platform]*")
+	cmd.Flags().StringVar(&bo.BinaryPath, "binary", "",
```
imjasonh, line 99: Could we rename this to `--binary-path` instead?

imjasonh (commented):
I think we should finally bite the bullet and accept a change for this, possibly even this change.

I have one comment about the flag name, but otherwise it looks pretty good. An end-to-end test that exercised this behavior would be great, especially to get confidence about the Windows case.

The other open PR to address this request has a few more tests, if you feel inclined to port those over to this change as well.

PR #1138, opened 2023-09-11
Introduce `--disable-entrypoint-overwrite` option to allow to disable entrypoint overwrite

creydr:
Currently the entrypoint of the image is always set to `/ko-app/<app-name>`. This makes it hard to have some kind of wrapper around the app (e.g. in case you want to remote debug the app via Delve, which invokes the app e.g. via `dlv exec <path-to-app>`).

This PR addresses it and introduces a new option (`--disable-entrypoint-overwrite`), which allows to use the entrypoint of the base image. To be able to get the path to the apps binary, this PR also introduce the `KO_APP_PATH` var, which points to the apps binary.
That way a developer who wants to remote debug their app (without to adjust yamls with an updated command & arg) could simply use a base image with a pre setup entrypoint, like the following:

```
ENTRYPOINT "/usr/bin/dlv exec --listen=:40000 --headless=true --log=true --accept-multiclient --api-version=2 $APP_PATH"
```
and build/apply via:
```
ko apply apply --disable-entrypoint-overwrite --disable-optimizations -f <path-to-yaml>
```

### DIFF
--- a/docs/reference/ko_apply.md
+++ b/docs/reference/ko_apply.md
@@ -45,27 +45,28 @@ ko apply -f FILENAME [flags]
 ### Options
 
 ```
-      --bare                     Whether to just use KO_DOCKER_REPO without additional context (may not work properly with --tags).
-  -B, --base-import-paths        Whether to use the base path without MD5 hash after KO_DOCKER_REPO (may not work properly with --tags).
-      --disable-optimizations    Disable optimizations when building Go code. Useful when you want to interactively debug the created container.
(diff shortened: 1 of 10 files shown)


imjasonh:
...And in fact that seems to be what https://github.com/ko-build/ko/pull/1128 starts to enable! 🎉 

If that seems promising we can push on that PR and see if that helps with your use case as well.

creydr:
Putting this on hold, as #1148 might be a better approach
/hold

PR #1142, opened 2023-09-11
handle newfound lint errors

imjasonh:
depguard is disabled because its new default config is to only allow stdlib deps (i.e., make you explicitly list each dep in your linter, for some reason) https://github.com/golangci/golangci-lint/issues/3877

deadcode is disabled because it's unmaintained and warns when you use it.

The rest of the changes fix findings reported by newer versions of golangci-lint

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -10,8 +10,6 @@ issues:
 linters:
   enable:
   - asciicheck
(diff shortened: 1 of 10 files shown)


cpanato (approved):
thanks!

PR #1176, opened 2023-11-06
Add more locking around on-disk image cache

jonjohnsonjr:
Concurrent writes/reads to the cache might result in partially written files, which we'd like to avoid.

On top of that, we generally don't want builds to fail if something goes wrong with the cache, so now we just log any cache errors and do a fallback.


imjasonh (approved):

cpanato (approved):
Thanks

PR #1180, opened 2023-11-08
docs: add MacPorts install info

herbygillot:
Add details on how to install `ko` on macOS using MacPorts: https://ports.macports.org/port/ko/


cpanato (approved):
thanks

PR #1198, opened 2023-12-10
annotation support

eugene-eeo:
same use case as https://github.com/ko-build/ko/issues/1090, pushing to ghcr registry. this PR adds ability to set annotations for both single/multi-arch builds, so they are nicely labelled on ghcr.io:

```
ko build --tags=v0.0.1 \
    --sbom=none --platform linux/amd64,linux/arm64 \
    --image-annotation org.opencontainers.image.source='https://github.com/foo/bar' \
    --image-annotation org.opencontainers.image.description='something about the package' \
    --push=true --bare \
    .
```

references:
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#labelling-container-images
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#adding-a-description-to-multi-arch-images

### DIFF
--- a/docs/reference/ko_apply.md
+++ b/docs/reference/ko_apply.md
@@ -45,27 +45,28 @@ ko apply -f FILENAME [flags]
 ### Options
 
 ```
-      --bare                     Whether to just use KO_DOCKER_REPO without additional context (may not work properly with --tags).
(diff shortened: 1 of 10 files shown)


Qup42:
It would be really great to have this feature. I would love to build properly annotated multi-arch images with a downstream tool (goreleaser).

PR #1208, opened 2023-12-28
Fix the KO_CONFIG_PATH default value in the documentation

amirh:
The default value listed for KO_CONFIG_PATH in configuration.md was `./ko.yaml` while in practice it is `/.ko.yaml` (defined in https://github.com/ko-build/ko/blob/9ae56937a80451e5b3f366828480a667643a9264/pkg/commands/options/build.go#L98)


cpanato (approved):
thank you

PR #1247, opened 2024-02-27
Fix fly.io deployment docs

imjasonh:
Fixes https://github.com/ko-build/ko/issues/1186


brunoluiz:
Even after logged in, as suggested in the documentation, I really had to do this:

```
fly deploy --image $(KO_DOCKER_REPO=registry.fly.io/<app> ko build <path-of-cmd> --bare)
```

1. It is not detecting the repository by itself, so I need to nudge it using `KO_DOCKER_REPO`
2. It is impossible to push without `--bare` (probably a limitation on how they store it)

PR #1261, opened 2024-03-18
feat: add Docker workingdir option

### LINKED ISSUE #1260: feature request: make it possible to set the WORKDIR
In the ["migrating from dockerfile" example](https://github.com/ko-build/ko/blob/7067ad211ba58ff19ca705a177edff20169d8872/docs/advanced/migrating-from-dockerfile.md), the base `Dockerfile` uses the `WORKDIR` statement https://docs.docker.com/reference/dockerfile/#workdir

However, I don't see a way to set the workdir in the configuration at https://github.com/ko-build/ko/blob/7067ad211ba58ff19ca705a177edff20169d8872/pkg/commands/options/build.go#L91-L172

As such, the workdir isn't set, e.g. `docker image inspect -f '{{.Config.WorkingDir}}' something_made_with_ko:latest` will return nothing.

`ko` inherits Docker config from a base container, but the default working directory of a Docker container is `/`, and if it's not set in the base, it will still default to `/`.

Docker containers that are designed to process data have a location where you'd expect to mount in a volume, e.g. `/data`. Without being able to set the detfault `workingdir`, you have to add an extra `-w` parameter to set it on each use.

```bash
docker run -w=/data -v `pwd`:/data  container_name:latest
```

Whereas, with the workingdir set, you could do:

```bash
docker run -v `pwd`:/data  container_name:latest
```

So, adding the ability to set the `workingdir` would be useful.

a-h:
Fixes #1260

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -84,6 +84,7 @@ type gobuild struct {
 	dir                  string
 	labels               map[string]string
 	semaphore            *semaphore.Weighted
+	workingDir           string
 
 	cache *layerCache
 }
@@ -106,6 +107,7 @@ type gobuildOpener struct {
 	labels               map[string]string
 	dir                  string
 	jobs                 int
+	workingDir           string
 }
 
 func (gbo *gobuildOpener) Open() (Interface, error) {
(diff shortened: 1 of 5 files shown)


a-h:
Uncool, bot.

PR #1281, opened 2024-04-18
Fix AWS Lambda advanced docs

mattn:


cpanato:
please sign the DCO

thanks for the PR!

docs/advanced/lambda.md
```diff
@@ -23,7 +23,7 @@ type Event struct {
 func main() {
     lambda.Start(func(ctx context.Context, event Event) (string, error) {
         return fmt.Sprintf("Hello %s!", event.Name), nil
-    }
+    })
```
cpanato, line 26: actually the correct is

```
func main() {
	lambda.Start(func(ctx context.Context, event Event) (string, error) {
		return fmt.Sprintf("Hello %s!", event.Name), nil
	})
}
```

cpanato (changes requested):

docs/advanced/lambda.md
mattn: What do you mean? I did fix it already. Do you mean replacing 4 spaces into tabs?

imjasonh (approved):

docs/advanced/lambda.md
imjasonh: I don't feel strongly about the tabs vs spaces, folks who use the code will have it gofmt'ed.

cpanato (approved):

mattn:
Ooops, race condition of commit & you comment. :(
I updated the Go code. (replaced 4 spaces into tabs)

mattn:
BTW, I'm trying ko with AWS Lambda, but it seems that `ko build` require `--bare` for AWS ECR. The docs is correctly?

PR #1307, opened 2024-05-11
feat: Add template params for git

### LINKED ISSUE #493: Define common build tags to consume in `.ko.yaml`
Since `.ko.yaml` is likely going to be used for more CI automation and releasing artifacts, it would be great to have access to common build variables, e.g. `date/time` (RFC3339), `commit` (short/long) and perhaps (git) `tag` instead of having to manually create them in the environment to reference them in `.ko.yaml`.

I know this might be out of scope for `ko` due to reasons (e.g. relying on `git`) but just wanted to throw this out so people can find this issue and an answer why this won't be implemented or if this is something you think is useful for the project.

Since `.ko.yaml` is inspired by `goreleaser`, we could reuse some of the standard build-related variables/templates (full list [here](https://goreleaser.com/customization/templates/)).

Example `.ko.yaml` idea:

```yaml
  ldflags:
  - -s -w
  - -extldflags "-static"
  - -X main.commit={{.Env.KO_COMMIT_LONG}}
  - -X main.shortCommit={{.Env.KO_COMMIT_SHORT}}
  - -X main.buildDate={{.Env.KO_DATETIME}}
```

nmittler:
This includes a number of template parameters supported by [goreleaser](https://goreleaser.com/customization/templates/). Specifically, the build date information and the majority of the Git params.

Majority of the code is copied from goreleaser. I've added the MIT license from goreleaser at the top of the files.

Fixes #493

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -31,6 +31,7 @@ import (
 	"strconv"
 	"strings"
 	"text/template"
+	"time"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -40,6 +41,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/types"
 	"github.com/google/ko/internal/sbom"
 	"github.com/google/ko/pkg/caps"
+	"github.com/google/ko/pkg/internal/git"
 	specsv1 "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/sigstore/cosign/v2/pkg/oci"
 	ocimutate "github.com/sigstore/cosign/v2/pkg/oci/mutate"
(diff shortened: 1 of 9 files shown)


nmittler:
It seems that the vendoring is causing issues with including dependencies like testify. I've reverted the mod/vendor changes in favor of just hand-rolling `requireXXX` methods.

pkg/build/gobuild.go
```diff
@@ -56,18 +58,24 @@ const (
 
 	defaultGoBin = "go"         // defaults to first go binary found in PATH
 	goBinPathEnv = "KO_GO_PATH" // env lookup for optional relative or full go binary path
+
+	envTemplateKey       = "Env"
+	gitTemplateKey       = "Git"
+	dateTemplateKey      = "Date"
+	timestampTemplateKey = "Timestamp"
```
imjasonh, line 65: Probably just personal style, but if these consts are only used once as map keys below, I think it would be easier to read by just inlining them down there. Otherwise when reading the code below you have to scroll back up to here to tell what `envTemplateKey` actually is.

imjasonh (commented):
Copying [Slack thread comment](https://kubernetes.slack.com/archives/C01T7DTP65S/p1715780250324799) for posterity:

> Do you intend to use all of these template features? I'm a little wary of speculatively adding features that we don't intend to use and get value from. If it's to keep compatibility with goreleaser I can understand, but that's also a moving target. :confused:
> I don't tend to use the goreleaser features myself to be honest, so I may be especially biased against this. If you or others do get value from that though I'm happy to approve.

imjasonh (approved):

PR #1310, opened 2024-05-14
feat: Add global env

### LINKED ISSUE #1305: feat: Add global env
If there are common environment variables used in each build in `.ko.yaml`, this can creates a lot of boilerplate. It would be better to allow a set of global variables at the top, that will automatically available to each build as its generated.

I suspect the order they should appear in the aggregate list is as follows:

- `os.Eniviron`
- global `env`
- per-build `env`

Happy to do the work, if folks think this is a good idea.

nmittler:
Supports configuring global `env` variables that will be applied to all builds.

Modifies the `builder` function to accept a `buildContext` structure. This will simplify similar modifications in the future.

Fixes #1305

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -104,6 +104,31 @@ You can also use the `KO_DEFAULTPLATFORMS` environment variable to set the defau
 KO_DEFAULTPLATFORMS=linux/arm64,linux/amd64
 ```
 
+### Setting build environment variables
+
+By default, `ko` builds use the ambient environment from the system (i.e. `os.Environ()`).
+These values can be overridden globally or per-build (or both).
+
+```yaml
+env:
+- FOO=foo
+builds:
+- id: foo
+  dir: .
+  main: ./foobar/foo
+  env:
+  - FOO=bar # Overrides the global value.
+- id: bar
+  dir: ./bar
+  main: .
+```
+
+For a given build, the environment variables are merged in the following order:
+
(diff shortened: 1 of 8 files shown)


pkg/commands/options/testdata/config/.ko.yaml
```diff
@@ -1,2 +1,3 @@
 defaultBaseImage: alpine
 defaultPlatforms: all
+env: CGO_ENABLED=0
```
imjasonh, line 3: Ko already does some weirdness with `CGO_ENABLED` by default, can you make this something more obviously for testing, like `FOO=bar`, and check that this is set in the build env?

imjasonh (changes requested):

imjasonh (dismissed):

docs/configuration.md
```diff
@@ -104,6 +104,31 @@ You can also use the `KO_DEFAULTPLATFORMS` environment variable to set the defau
...
 ```
 
+### Setting build environment variables
+
+By default, `ko` builds use the ambient environment from the system (i.e. `os.Environ()`).
+These values can be overridden globally or per-build (or both).
+
+```yaml
+env:
+- FOO=foo
```
imjasonh, line 114:
```suggestion
- FOO=foo
```

imjasonh (approved):

cpanato (approved):
thanks!

PR #1312, opened 2024-05-15
Remove vendor directory

nmittler:
Also adding dependency on testify to replace some hand-rolled `requireXXX` methods.


imjasonh (approved):
Congrats on being the ko contributor to have deleted the most lines. 🚀

PR #1318, opened 2024-05-17
Refactor global values to be defaults

### LINKED ISSUE #1317: 0.15.3 has breaking issues error resolving image references: build: go build: exit status 2: flag provided but not defined: -platform
0.15.3 has breaking issues

https://github.com/knative/serving/actions/runs/9127229170/job/25097079749?pr=15219

```
2024/05/17 11:05:55 Building knative.dev/serving/cmd/webhook for linux/amd64
Error: error processing import paths in "config/core/999-cache.yaml": error resolving image references: build: go build: exit status 2: flag provided but not defined: -platform
usage: go build [-o output] [build flags] [packages]
Run 'go help build' for details.
```

nmittler:
There was recent work to add global values for `env`, `flags`, and `ldflags`. The global values would be merged with per-build values to generate the value used for the builds.

There are a couple issues with this:

- It's inconsistent with the existing code, which only has `default` values declared globally (there is no merging today).
- The name of the `flag` variable, caused a conflict with knative's `KO_FLAGS` environment variable (see https://github.com/ko-build/ko/issues/1317)

This PR does the following:

- Refactors the logic to use `defaultEnv`, `defaultFlags`, and `defaultLdflags`. This resolves both issues described above.
- Updates documentation

Fixes https://github.com/ko-build/ko/issues/1317

### DIFF
--- a/integration_test.sh
+++ b/integration_test.sh
@@ -44,6 +44,7 @@ else
 fi
 
 echo "2. Test knative 'KO_FLAGS' variable is ignored."
+# https://github.com/ko-build/ko/issues/1317
 RESULT="$(KO_FLAGS="--platform=badvalue" ./ko build --local --platform="linux/$GOARCH" "$GOPATH/src/github.com/google/ko/test" | grep "$FILTER" | xargs -I% docker run %)"
 if [[ "$RESULT" != *"Hello there"* ]]; then
   echo "Test FAILED. Saw $RESULT" && exit 1
--- a/integration_test.sh
+++ b/integration_test.sh
@@ -43,7 +43,15 @@ else
   echo "Test PASSED"
 fi
 
-echo "2. Linux capabilities."
+echo "2. Test knative 'KO_FLAGS' variable is ignored."
+RESULT="$(KO_FLAGS="--platform=badvalue" ./ko build --local --platform="linux/$GOARCH" "$GOPATH/src/github.com/google/ko/test" | grep "$FILTER" | xargs -I% docker run %)"
+if [[ "$RESULT" != *"Hello there"* ]]; then
+  echo "Test FAILED. Saw $RESULT" && exit 1
+else
+  echo "Test PASSED"
+fi
+
+echo "3. Linux capabilities."
 pushd test/build-configs || exit 1
 # run as non-root user with net_bind_service cap granted
 docker_run_opts="--user 1 --cap-add=net_bind_service"
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -125,18 +125,18 @@ KO_DEFAULTPLATFORMS=linux/arm64,linux/amd64
 ### Setting build environment variables
 
 By default, `ko` builds use the ambient environment from the system (i.e. `os.Environ()`).
-These values can be overridden globally or per-build (or both).
+These values can be overridden for your build.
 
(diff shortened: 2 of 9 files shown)


nmittler:
@dprotaso @imjasonh  PTAL

nmittler:
The integration test I added does seem to indicate that the original error is gone.

However, it fails on my m1 macbook. with:

```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
exec /ko-app/test: exec format error
```

The test should be using the `GOARCH` of the local platform, but it apparently doesn't. This makes me think it's not actually using the value passed in by `KO_FLAGS`.

I can't see how my recent changes could be related, however.

dprotaso:
> This makes me think it's not actually using the value passed in by KO_FLAGS.

It's not - so knative uses this in our scripts and we invoke `ko` like `ko resolve ${KO_FLAGS}` so it was sorta a coincidence your change started reading that flag.

Hence why I think it might be better to be more descriptive with the flags.

eg. `defaultFlags` => `goBuildFlags` then that could map to `KO_GO_BUILD_FLAGS` or something

nmittler:
@dprotaso

> It's not - so knative uses this in our scripts and we invoke ko like ko resolve ${KO_FLAGS} so it was sorta a coincidence your change started reading that flag.

Ah, that makes sense. Up until now I was under the impression that this was something viper was doing under the covers. 

But given that ko is effectively reserving the `KO_` prefix through viper, isn't the real issue that knative is using that prefix in the first place? There will always be a chance of collision going forward otherwise.

I'm a little concerned with `goBuildXXX` names with respect to consistency:

```yaml
defaultBaseImage: ...
defaultPlatforms:
- linux/arm64
- linux/amd64
goBuildEnv:
- FOO=bar
goBuildFlags:
- -v
goBuildLdflags:
- -s
builds:
- id: foo
  dir: .
  main: ./foobar/foo
  flags:
  - -trimpath
  ldflags:
  - -w
...
```

This seems to break the naming convention that was established with `defaultXXX`. No other variable uses `go` (which I suppose is just assumed since ko only builds go applications) or `build` (all of the config is effectively build-related).

From the perspective of ko, just sticking with the `defaultXXX` pattern would seem to make sense both in name and behavior.

Thoughts?

nmittler:
@imjasonh @cpanato 

Any concerns about the breaking changes in this PR? I think regardless which way we go, something will break between 0.15.3 and 0.15.4.

imjasonh:
I'm comfortable breaking behavior from 0.15.3 -> 0.15.4, since it's very unlikely anybody has come to rely on it during that short window. (If you're here because you came to rely on it during that short window, I'm sorry, I owe you a beer! 🍻 )

integration_test.sh
```diff
@@ -43,7 +43,15 @@ else
   echo "Test PASSED"
 fi
 
-echo "2. Linux capabilities."
+echo "2. Test knative 'KO_FLAGS' variable is ignored."
```
imjasonh, line 46:
```suggestion
echo "2. Test knative 'KO_FLAGS' variable is ignored."
# https://github.com/ko-build/ko/issues/1317
```

imjasonh (approved):

PR #1148, opened 2023-09-13
Add `--debug` build flag to include Delve debugger

creydr:
Currently the entrypoint of the image is always set to `/ko-app/<app-name>`. This makes it hard to remote debug the app via Delve, which invokes the app e.g. via `dlv exec <path-to-app>`).

This PR adds a new build flag (`--debug`), which adds the Delve debugger to the image and sets the entrypoint to invoke delve instead of the ko-app directly.

**Hint:** 
* This idea came out of https://github.com/ko-build/ko/pull/1138#issuecomment-1713834074.
* I have only tested it on Linux. Tests on MacOS & Windows are welcome :) 
* To allow a more customized exec configuration of delve (e.g. specify port, etc.), we could allow more configuration flags, to override the default delve args. But I would see this as a follow up, to not blow up this initial PR. (IMO)

**How to test this PR:**
1. Apply a manifests with the `debug` flag. e.g. `ko apply --debug -f config/controller.yaml`
2. Start a port forwarding for Delve: `kubectl port-forward deploy/my-controller 40000:40000`
3. Connect from your IDE to the remote debugger. E.g. use the following VSCode launch config: 
      ````
      {
          "version": "0.2.0",
          "configurations": [{
                  "name": "Attach to remote",
                  "type": "go",
                  "request": "attach",
                  "mode": "remote",
                  "port": 40000,
                  "host": "0.0.0.0",
                  "showLog": true,
                  "cwd": "${workspaceFolder}"
              }
          ]
      }
      ````

PR #1320, opened 2024-05-21
Add debug flag to include delve

luhring:
Picking up the awesome work from https://github.com/ko-build/ko/pull/1148, just resolving current conflicts and seeing if we can get this merged.

I've tested this running the built container in Docker Desktop on macOS, using GoLand (running natively on macOS), and it appears to work great.

If there's anything further this PR needs, I'm happy to help out. I'd love to see this `--debug` flag in `ko`! 😃 

cc: @imjasonh

### DIFF
--- a/docs/features/debugging.md
+++ b/docs/features/debugging.md
@@ -1,0 +1,29 @@
+# Debugging
+
+Sometimes it's challenging to track down the cause of unexpected behavior in an app. Because `ko` makes it simple to make tweaks to your app and immediately rebuild your image, it's possible to iteratively explore various aspects of your app, such as by adding log lines that print variable values.
+
+But to help you solve the problem _as fast as possible_, `ko` supports debugging your Go app with [delve](https://github.com/go-delve/delve).
+
+To use this feature, just add the `--debug` flag to your `ko build` command. This adjusts how the image is built:
+
+- It installs `delve` in the image (in addition to your own app).
+- It sets the image's `ENTRYPOINT` to a `delve exec ...` command that runs the Go app in debug-mode, listening on port `40000` for a debugger client.
+- It ensures your compiled Go app includes debug symbols needed to enable debugging.
+
+**Note:** This feature is geared toward development workflows. It **should not** be used in production.
+
+### How it works
+
+Build the image using the debug feature.
+
+```plaintext
+ko build . --debug
+```
+
+Run the container, ensuring that the debug port (`40000`) is exposed to allow clients to connect to it.
+
+```plaintext
+docker run -p 40000:40000 <img>
+```
+
+This sets up your app to be waiting to run the command you've specified. All that's needed now is to connect your debugger client to the running container!
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -309,11 +309,11 @@ func getDelve(ctx context.Context, platform v1.Platform) (string, error) {
 	cloneDir := filepath.Join(tmpInstallDir, "delve")
 	err = os.MkdirAll(cloneDir, 0755)
 	if err != nil {
-		return "", fmt.Errorf("making dir for delve clone: %v", err)
+		return "", fmt.Errorf("making dir for delve clone: %w", err)
 	}
 	err = git.Clone(ctx, cloneDir, delveCloneURL)
 	if err != nil {
-		return "", fmt.Errorf("cloning delve repo: %v", err)
+		return "", fmt.Errorf("cloning delve repo: %w", err)
 	}
 	osArchDir := fmt.Sprintf("%s_%s", platform.OS, platform.Architecture)
 	delveBinaryPath := filepath.Join(tmpInstallDir, "bin", osArchDir, "dlv")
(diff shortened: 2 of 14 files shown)


pkg/build/gobuild.go
```diff
@@ -274,6 +278,63 @@ func getGoBinary() string {
...
+			return err
+		}
+
+		if !d.IsDir() && strings.Contains(d.Name(), "dlv") {
+			delveBinary = path
+		}
+
+		return nil
+	})
+	if err != nil {
```
imjasonh, line 327:
```suggestion
	if err := filepath.WalkDir(filepath.Join(tmpInstallDir, "bin"), func(path string, d fs.DirEntry, err error) error {
		if err != nil {
			return err
		}

		if !d.IsDir() && strings.Contains(d.Name(), "dlv") {
			delveBinary = path
		}

		return nil
	}); err != nil {
```

pkg/build/gobuild.go
```diff
@@ -274,6 +278,63 @@ func getGoBinary() string {
...
+
+	log.Printf("Building Delve for %s", platform)
+	if err := cmd.Run(); err != nil {
+		os.RemoveAll(tmpInstallDir)
+		return "", fmt.Errorf("go build Delve: %w: %s", err, output.String())
+	}
+
+	// find the delve binary in tmpInstallDir/bin/
+	delveBinary := ""
+	err = filepath.WalkDir(filepath.Join(tmpInstallDir, "bin"), func(path string, d fs.DirEntry, err error) error {
```
imjasonh, line 316: Why do we have to walk? Don't we know it'll be at `tmpInstallDir/bin/dlv`?

pkg/build/gobuild.go
```diff
@@ -1022,6 +1083,45 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
...
+	delvePath := "" // path for delve in image
+	if g.debug {
+		// get delve locally
+		delveBinary, err := getDelve(ctx, *platform)
+		if err != nil {
+			return nil, fmt.Errorf("building Delve: %w", err)
+		}
+		defer os.RemoveAll(filepath.Dir(delveBinary))
+
+		delvePath = "/usr/bin/" + filepath.Base(delveBinary)
```
imjasonh, line 1095: It might make sense to put it in `/ko-app/dlv` just so it's clear it's owned by `ko` and not by the base image, in case it's already available there.

imjasonh (commented):
This looks good, thank you @creydr especially for your work and patience, and for putting up with us.

In addition to the CLI docs it'd be great to have some usage docs for the website, but that can happen later, when this is released.

pkg/build/gobuild.go
luhring: That's a good point... We probably want to confirm the end state after the "go install", but IIUC we could do that more directly with a stat call with the expected path.

pkg/build/gobuild.go
luhring: That makes sense to me.

pkg/build/gobuild.go
luhring: Ah I just hit a snag here that I'm guessing motivated this design: testing again locally, the binary was at `tmpInstallDir/bin/linux_arm64/dlv`.

pkg/build/gobuild.go
luhring: Pushed an update!

pkg/build/gobuild.go
creydr: Hey @imjasonh,
the motivation behind this is as @luhring said in https://github.com/ko-build/ko/pull/1320#discussion_r1607491196: for other archs it can be located at different locations/folders.

creydr:
Awesome @luhring,
thanks for fixing the merge conflicts (last week I didn't see any when I rebased #1148 :see_no_evil: ) and giving this some push again.

pkg/build/gobuild.go
imjasonh: That makes sense! In that case, should we only look at `tmpInstallDir/bin/$GOOS_GOARCH/dlv`? I'm worried the walking might open us up to a bug where `dlv` gets installed in the image for the wrong arch.

luhring: Pushed e1438e2 to address this, please let me know if that's looking any better!

docs/features/debugging.md
luhring, line 1: @imjasonh I started this draft of a docs page. Very open to feedback!

I'm also wondering if it'd be nice to talk through how to set up VS Code or GoLand to leverage the now-debuggable container image. Let me know if you think we should add that, or if that's out of scope for ko's docs.

pkg/build/gobuild.go
```diff
@@ -969,6 +979,9 @@ func (g *gobuild) buildOne(ctx context.Context, refStr string, base v1.Image, pl
 			OSVersion:    cf.OSVersion,
 		}
 	}
+	if g.debug && !doesPlatformSupportDebugging(*platform) {
+		return nil, fmt.Errorf("debugging is not supported for %s", platform)
```
imjasonh, line 983: I'd log and skip in this case, in case someone wants to build for all platforms but only intends to debug for certain platforms.

luhring: Okay, gotcha. I tried to understand how "skip" would work, but it wasn't immediately obviously to me. Would we return a nil `oci.SignedImage` and a nil `error`?

pkg/build/gobuild.go
```diff
@@ -277,6 +277,16 @@ func getGoBinary() string {
 	return defaultGoBin
 }
 
+func doesPlatformSupportDebugging(platform v1.Platform) bool {
+	// Here's the list of supported platforms by Delve:
+	//
+	// https://github.com/go-delve/delve/blob/master/Documentation/faq.md#unsupportedplatforms
+	//
+	// For the time being, we'll support only linux/amd64 and linux/arm64.
```
creydr, line 285: What is the reason for not supporting Darwin (macOS) / amd64?

pkg/build/gobuild.go
creydr: @imjasonh can you check on this?

imjasonh (dismissed):

imjasonh (approved):

PR #1323, opened 2024-05-28
Fix a typo in terraform.md

chaodaiG:
The field name of `image` does not exist per https://registry.terraform.io/providers/ko-build/ko/latest/docs/resources/build#read-only.

Found this problem while trying to use it, and confirmed it working with this fix.


chaodaiG:
@imjasonh

cpanato (approved):
thanks

PR #1333, opened 2024-06-10
remove support for CycloneDX SBOMs

imjasonh:


caarlos0:
sorry for being out of the loop, but, can I ask why its been removed?

caarlos0:
also, this removes go modules sboms as well 🤔

imjasonh:
> also, this removes go modules sboms as well 🤔

That option was never really supported or documented.

As for cyclonedx, I'm not aware of any user that used it, and there were more than a few bugs in it. It seemed not worth the effort to fix them, based on usage.

If you're aware of users of either of these features, let me know, we can talk about next steps for them.

caarlos0:
ah, that's fair! 

it was allowed on goreleaser's ko integration, but no idea if anyone used it (I have never seen).

fwiw, i removed both options from goreleaser too:

- https://github.com/goreleaser/goreleaser/pull/5063
- https://goreleaser.com/deprecations/#kossbom

chipzoller:
We were using it in our Kubecost disk-autoscaler program and this has broken our releaser workflows :) I guess we'll have to use SPDX from now on.

Issue #1378, opened 2024-08-22
Using --image-refs in ko build results in dirty git


jeffmendoza:
Run `ko build` with `--image-refs=refs.txt`
See:
```
2024/08/22 11:18:46 git is in a dirty state
Please check in your pipeline what can be changing the following files:
?? refs.txt
```
It would be nice if this didn't create the file until after the check.

jeffmendoza:
Looks like`--oci-layout-path` does the same thing. I can fix that as well.

PR #1385, opened 2024-09-04
Change OCI Layout publisher to lazy create layout after build.

jeffmendoza:
When using ko with OCI Layout, Git status will always be dirty because the layout is created before this build. This changes the layout to be created after the build so that git will be clean. The means that any error creating the layout won't be seen until after the build.


jeffmendoza:
As noted in #1378

imjasonh (approved):

Issue #944, opened 2023-01-26
Renameable binaries / path


leongross:
`ko` defaults to placing binaries into the path `/ko-app/<app-name>`.
As for now it seems as if there is no way to set this name during the build process.

This is especially important if the containers are used in a Kubernetes environment where the `entrypoint` is ignored and the absolute path of the binary is important.

Although this path might be known it complicates the drop-in integration of `ko` containers in a production environment, since the deployments have to be adjusted to use the `ko` specific path.

The lines responsible for the naming seem to be [these](https://github.com/ko-build/ko/blob/96d25186014baaf546c2c7a960f75ba1687bc7e3/pkg/build/gobuild.go#L861).

It would be great if a new flag such as `-o` / `--out-path` or something similar could be introduced that let the user set the output path to their needs.

samox73:
Is there any update on this? I'd like to switch over to ko from docker's buildx, but I'm also facing the problem of using ko with pre-existing projects where the entrypoint may not change.

imjasonh:
Sorry for not responding earlier.

I'm not opposed to this in general, but I'd like to understand a bit more what requirements are driving it. Is the problem that a runtime config might be specifying the image's entrypoint, and it breaks when migrating to build with `ko`?

Is there a reason the image's entrypoint can't be used directly, so that both the old buildx image and the new `ko` image both work?

`ko` in general aims to make details of the image like this invisible to users, as as UX feature. I understand if sometimes this is too optimistic, and folks need this low-level control.

unmarshall:
If you use docker to build the images (multi-stage builds) then you build the binary in the builder step and copy the built binary to a specific location which also becomes the entry point. One has complete control over it. This is also then used to set the command in the k8s deployment yaml file. Now when moving to KO this changes as it decides to prefix the entry point with `ko-app` and has no way to overwrite it. This creates an issue for all existing deployment yaml files.

samox73:
As @unmarshall explained, having a fixed entrypoint poses the issue of breaking working systems and using ko as a drop-in replacement is not possible. Having the ability to change the default  entrypoint would make adoption of ko into existing project much easier.

In my particular case, I am writing a k8s operator with the [operator framework](https://sdk.operatorframework.io/) and when building the bundle manifests it is not documented how to change the k8s `command` of the container. I found that this is defined in `/config/manager.yaml` and could (as @unmarshall said) work my way around it, albeit somewhat clumsily.

imjasonh:
[...]

PR #1403, opened 2024-09-20
Allow configuring app directory/path binaries are placed into

UnseenWizzard:
This PR proposes a way to allow optionally modifying under which path binaries are placed inside containers built with ko - as raised in issue #944 

Without deeper insight into the code base, this seems like a straightforward addition with limited impact on future maintainability to me - while allowing users migrating to ko to still place binaries in the same location as their previous Docker containers.

Looking forward to your input and any suggestions!

### DIFF
--- a/pkg/build/gobuild_test.go
+++ b/pkg/build/gobuild_test.go
@@ -640,7 +640,13 @@ func TestGoBuildNoKoData(t *testing.T) {
 	})
 }
 
-func validateImage(t *testing.T, img oci.SignedImage, baseLayers int64, creationTime v1.Time, checkAnnotations bool, expectSBOM bool) {
+type validationOptions struct {
+	checkAnnotations   bool
+	expectSBOM         bool
+	entryPointBasePath string
+}
+
+func validateImage(t *testing.T, img oci.SignedImage, baseLayers int64, creationTime v1.Time, opts validationOptions) {
 	t.Helper()
 
 	ls, err := img.Layers()
@@ -722,7 +728,11 @@ func validateImage(t *testing.T, img oci.SignedImage, baseLayers int64, creation
 			t.Errorf("len(entrypoint) = %v, want %v", got, want)
 		}
 
-		if got, want := entrypoint[0], "/ko-app/test"; got != want {
+		want := "/ko-app/test"
+		if opts.entryPointBasePath != "" {
+			want = path.Join(opts.entryPointBasePath, "test")
+		}
+		if got := entrypoint[0]; got != want {
 			t.Errorf("entrypoint = %v, want %v", got, want)
 		}
 	})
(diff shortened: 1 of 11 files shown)


pkg/build/gobuild_test.go
```diff
@@ -640,7 +640,13 @@ func TestGoBuildNoKoData(t *testing.T) {
...
 }
 
-func validateImage(t *testing.T, img oci.SignedImage, baseLayers int64, creationTime v1.Time, checkAnnotations bool, expectSBOM bool) {
+type validationOptions struct {
+	checkAnnotations   bool
+	expectSBOM         bool
+	entryPointBasePath string
+}
+
+func validateImage(t *testing.T, img oci.SignedImage, baseLayers int64, creationTime v1.Time, opts validationOptions) {
```
UnseenWizzard, line 649: note that `baseLayers` being 3 in all test uses triggers the linter - as it's unrelated to the change, I didn't want to change the function to appease the linter. lmk if I should

maboehm:
@UnseenWizzard @imjasonh 
Really a bummer to see this PR get stale and closed. We would really find this feature helpful and that change (while adding some new functionality to maintain) is non-intrusive and non-breaking.

Could we re-open this PR?

UnseenWizzard:
From my perspective I'm very happy to reopen this PR and make any changes need to get in! Just let me know 

I just honestly got the impression from the issue discussions, that a feature like this, while relevant to a part of the user base, is not of interest to maintainers. Which is a pity.

cpanato:
reopened, i also agree with that would be a nice feature

Laubi:
This feature would also be really nice for us, as we don't want to introduce breaking changes into existing images. What is needed to merge this feature?

UnseenWizzard:
@cpanato change is rebased and conflicts (in docs/ only) resolved :) 
CI ran successfully on my fork - note the comment about linting above, lmk if and how you want that addressed with this PR.

I'd also like to add an E2E test for this change - should that simply be added as another part of the `Build and run [...]` step in the `e2e.yaml` action? Or would it make sense to split that action into more than one step?

(Btw I can't view why the license compliance check fails here but it seems to do the same on main?)

rfranzke:
We would also very much like to have this feature :) Can we do something to move this PR forward?

hakman:
This would be very useful to lots of projects.
@justinsb had a similar PR, but I think he missed the feedback from @imjasonh. https://github.com/ko-build/ko/pull/1128#pullrequestreview-2405708775
@cpanato @imjasonh could you take another look? Thanks!

pkg/build/gobuild_test.go
den-vasyliev: Thanks, just use your PR locally. Appreciate for the project, but not sure why we need branded entrypoints (Makefile, workflows already has references)...

PR #1426, opened 2024-10-15
allow setting annotations

### LINKED ISSUE #1090: option to set image annotations
Currently we can set image labels with `--image-label`, but when working with multi platform images, there's no canonical image to look at for metadata.
Registries like github's ghcr instead look at the annotations on the image/index.
I'd like a way to set those.

https://kubernetes.slack.com/archives/C01T7DTP65S/p1687637484368199

### LINKED ISSUE #1231: Chainguard annotations are, but should not be, propagated to multi-platform manifest
When I `ko build` a multi-platform manifest, _my_ manifest ends up with annotations attributing it to chainguard. Here is what I found:

```json
{"org.opencontainers.image.authors": "Chainguard Team https://www.chainguard.dev/",
 "org.opencontainers.image.base.digest": "sha256:fce1fdce595332afe7f339303ae288c349c4e4139d926f8c7ccf4b7ca2911553",
 "org.opencontainers.image.base.name": "cgr.dev/chainguard/static:latest",
 "org.opencontainers.image.source": "https://github.com/chainguard-images/images/tree/main/images/static",
 "org.opencontainers.image.url": "https://edu.chainguard.dev/chainguard/chainguard-images/reference/static/"}
```

### LINKED ISSUE #1235: ko build does not give me a way to put annotations in a multi-platform manifest
As far as I can tell, there is not way that a `ko build` user, when building a multi-platform manifest, can specify annotations that go into that manifest. It would be good if there were a way.

### LINKED ISSUE #1395: Support setting OCI annotations in images built by ko - and stop inheriting all annotations
[OCI annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md) can be set on manifests, to provide metadata about the images.

`ko` today supports setting labels (introduced by docker) but not setting annotations (introduced by the OCI spec). 
It does set annotations related to the base image used, which is great, but it also sets additional annotations, inherited from the base image, which do not apply to the target image, and cannot be overwritten using `ko`.

This feature request stems from something I noticed on Tekton container images.
Since Tekton uses Chainguard's base images and `ko` to build its images, the annotations resulting in Tekton images look like:

```
crane manifest gcr.io/tekton-releases/github.com/tektoncd/pipeline/cmd/resolvers:v0.63.0@sha256:e3d170eb9bc04c6ee69306529c2f82e84296a52414e136ef16995ac179abcd12 | jq .annotations
{
  "org.opencontainers.image.authors": "Chainguard Team https://www.chainguard.dev/",
  "org.opencontainers.image.base.digest": "sha256:67a1b00e0134e2b3a614c7198a26f7deed9d11b7acad4d52c79c0cfd47a2eae7",
  "org.opencontainers.image.base.name": "cgr.dev/chainguard/static@sha256:67a1b00e0134e2b3a614c7198a26f7deed9d11b7acad4d52c79c0cfd47a2eae7",
  "org.opencontainers.image.source": "https://github.com/chainguard-images/images/tree/main/images/static",
  "org.opencontainers.image.url": "https://edu.chainguard.dev/chainguard/chainguard-images/reference/static/"
}
```

While the `org.opencontainers.image.base` part is correct, the values author, source and URL do not apply to Tekton images.

seankhliao:
This adds a new build flag for setting annotations on image indexes and manifests.
Annotations are no longer copied from the base image. `org.opencontainers.image.base.digest` and
`org.opencontainers.image.base.name`
are always set to the resolved values.

Usage example:

```sh
ko build --image-annotation foo=bar,fizz=buzz .
```

Fixes #1090
Fixes #1090
Fixes #1231
Fixes #1235
Fixes #1395

### DIFF
--- a/docs/reference/ko_apply.md
+++ b/docs/reference/ko_apply.md
@@ -45,28 +45,29 @@ ko apply -f FILENAME [flags]
 ### Options
 
 ```
-      --bare                     Whether to just use KO_DOCKER_REPO without additional context (may not work properly with --tags).
-  -B, --base-import-paths        Whether to use the base path without MD5 hash after KO_DOCKER_REPO (may not work properly with --tags).
-      --debug                    Include Delve debugger into image and wrap around ko-app. This debugger will listen to port 40000.
-      --disable-optimizations    Disable optimizations when building Go code. Useful when you want to interactively debug the created container.
-  -f, --filename strings         Filename, directory, or URL to files to use to create the resource
-  -h, --help                     help for apply
-      --image-label strings      Which labels (key=value) to add to the image.
-      --image-refs string        Path to file where a list of the published image references will be written.
-      --insecure-registry        Whether to skip TLS verification on the registry
-  -j, --jobs int                 The maximum number of concurrent builds (default GOMAXPROCS)
-  -L, --local                    Load into images to local docker daemon.
-      --oci-layout-path string   Path to save the OCI image layout of the built images
-      --platform strings         Which platform to use when pulling a multi-platform base. Format: all | <os>[/<arch>[/<variant>]][,platform]*
-  -P, --preserve-import-paths    Whether to preserve the full import path after KO_DOCKER_REPO.
-      --push                     Push images to KO_DOCKER_REPO (default true)
-  -R, --recursive                Process the directory used in -f, --filename recursively. Useful when you want to manage related manifests organized within the same directory.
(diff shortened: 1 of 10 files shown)


imjasonh (approved):

imjasonh:
This looks good! I slightly wonder if we should have some mode (possibly default) where labels are just copied to annotations, but I don't feel strongly and this is clearly something that folks want, so let's just do it.

Thanks for this contribution!

afrittoli:
Cool, thank you!

Issue #1424, opened 2024-10-13
Allow adding custom CA certificates


kosamson:
To load in custom CA root certificates into a container image, I usually use [this method in a Dockerfile](https://stackoverflow.com/questions/42292444/how-do-i-add-a-ca-root-certificate-inside-a-docker-image). I also recently learned of another possible method using [`SSL_CERT_FILE`](https://go.dev/src/crypto/x509/root_unix.go).

Using Ko, I was able to add my custom CA root certificates by baking them into a custom image and then [overriding the default Ko base image](https://ko.build/configuration/#overriding-base-images) to use the custom image. 

However, given the common use case of adding a custom root CA (ex: accessing sites in a private corporate network with a private corporate CA), I am wondering if loading in custom CA root certificates should have a first-class configuration option within Ko.

kosamson:
I think I found a different workaround by setting [`SSL_CERT_DIR`](https://go.dev/src/crypto/x509/root_unix.go) in combination with Ko's [static assets (kodata)](https://ko.build/features/static-assets/).

[Example implementation](https://github.com/kosamson/ko-private-ca-test)

This avoids needing to make a custom base image, and instead relies on:
1. The CI runner to download the custom CA certificates and store them in the repository under `${REPOSITORY_ROOT}/kodata/`
2. The container runtime / orchestrator to set `SSL_CERT_DIR` equal to `KO_DATA_PATH`

kosamson:
To clarify, the requirement I was looking to satisfy was injecting the custom CA certificates into the container and integrate it with the Go app **without storing the CA certificate(s) in the repository**, which the above workaround satisfies. I think this may work long-term, but I need to do some testing to see if overriding `SSL_CERT_DIR=$KO_DATA_PATH` impacts the default CA certificates.

Based on the Go source, I'm concerned that overriding the [default value](https://github.com/golang/go/blob/master/src/crypto/x509/root_linux.go#L20) will [replace, not append, the certificate directories](https://github.com/golang/go/blob/master/src/crypto/x509/root_unix.go#L53).

kosamson:
I did note in my test that even after overriding `SSL_CERT_DIR`, I was still able to perform a TLS handshake with common public root CAs.

[...]

PR #1428, opened 2024-10-17
add documentation for installing custom root ca certificates

kosamson:
Adds a documentation page under the "Advanced" section outlining how to install custom root certificates into a container built using `ko`.

Discussed initially in #1424 and suggested by @jeffmendoza

### DIFF
--- a/docs/advanced/root-ca-certificates.md
+++ b/docs/advanced/root-ca-certificates.md
@@ -1,0 +1,72 @@
+# Root CA Certificates
+
+To install a [root certificate](https://en.wikipedia.org/wiki/Root_certificate) into your container built using `ko`, you can use one of the following methods.
+
+## Custom Base Image
+
+New root certificates can be [installed into a custom image](https://stackoverflow.com/questions/42292444/how-do-i-add-a-ca-root-certificate-inside-a-docker-image) using standard OS packages. Then, this custom image can be used [to override the base image for `ko`](https://ko.build/configuration/#overriding-base-images). Once the Go application container image is built using `ko` with the custom base image, the root certificates installed on the base image will be trusted by the Go application.
+
+### Example
+
+1. Make a custom container image with your new root certificates
+```dockerfile
+# Dockerfile
+FROM alpine
+
+RUN apk update
+RUN apk add ca-certificates
+
+ADD new-root-ca.crt /usr/local/share/ca-certificates/new-root-ca.crt
+RUN chmod 644 /usr/local/share/ca-certificates/new-root-ca.crt
+RUN update-ca-certificates
+```
+
+2. Build and push the custom container image to a container registry
+```sh
+docker build . -t docker.io/ko-build/image-with-new-root-certs
+docker push docker.io/ko-build/image-with-new-root-certs
+```
+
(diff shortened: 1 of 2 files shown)


docs/advanced/root-ca-certificates.md
```diff
@@ -0,0 +1,72 @@
...
+mkdir -p kodata
+cp $CERT_FILE_DIR/*.crt kodata/
+```
+
+2. Build the Go application container image
+```sh
+KO_DOCKER_REPO=docker.io/ko-build/static-assets-certs ko build .
+```
+
+3. Run the Go application container image with `SSL_CERT_DIR` equal to `/var/run/ko` (the default value for `$KO_DATA_PATH`)
```
kosamson, line 67: **question**: I thought of something I might have misunderstood while writing this PR.

Can `ko` users override the value for `KO_DATA_PATH` and subsequently where `ko` will bundle the static assets in the built image? I tried doing this myself and could not override the default path of `/var/run/ko` on a Linux base image.

After looking at [the code](https://github.com/ko-build/ko/blob/903da574dc54a272a58b3881dee1f61a4a284b1f/pkg/build/gobuild.go#L775) I think I understand that this value is statically set, and is abstracted via an environment variable just for the sake of running `ko` on different operating systems.

Is this correct? If so, I will probably change this example to say 
> equal to `/var/run/ko` (the default value for `$KO_DATA_PATH` on Linux)

Also, if I am correct about the value being statically set per OS, should we clarify that on the [static assets documentation page](https://ko.build/features/static-assets/)?

cpanato (commented):
also can use: https://github.com/chainguard-dev/incert after you build your image with ko

kosamson:
> also can use: https://github.com/chainguard-dev/incert after you build your image with ko

This is really useful, I've never heard about this before. Let me add a section on using this in the new documentation page. This makes building a CI workflow to add the certificates automatically seem really easy.

kosamson:
added a section on using `incert` (thank you for the suggestion @cpanato).

cpanato (dismissed):
lgtm

will wait for @imjasonh for more 👀

kosamson:
rebased my branch to main. i haven't done a github PR in a while so not sure if I messed up the review workflow. if I need to change something pls let me know and I'll fix it.

jeffmendoza (approved):
Awesome!

docs/advanced/root-ca-certificates.md
imjasonh: When running `ko` to build, the directory is by convention `<importpath>/kodata` with no way to override that.

The env var comes into play _inside_ the container when it runs, and points to the path where the contents of `kodata` have been bundled in the image (`/var/run/ko`). The reason the env var is useful is so that you can have code that works inside and outside the container:

```
    http.Handle("/", http.FileServer(http.Dir(os.Getenv("KO_DATA_PATH"))))
```

works in the running container since `ko` sets that env for you to `/var/run/ko`, and outside the container you can do `KO_DATA_PATH=cmd/foo/kodata go run ./cmd/foo` to achieve the same result.

Let me know if there's a clearer way to state that in the docs.

imjasonh (approved):
Thanks for writing this up!

revusf81-cmyk:
Hey I'm Revus F

PR #1447, opened 2024-11-21
fix(formatting): Updated the codeblock for Dockerfile to make it more readable

userbradley:
The Dockerfile example, whilst not functional at all, is not readable with the current formatting applies

This PR updates the formatting on the mkdocs site

image


cpanato (approved):
thanks

PR #1451, opened 2024-11-24
fix(debug): ensure reproducible builds in debug mode

l-qing:
When use `--debug` build flag, a new dlv executable file is built into the image. However, during the build, some build flags are not specified, resulting in the binary being different each time, which ultimately causes the digest of the image to be inconsistent.

This change aims to ensure that, as long as the remote dlv code has not been updated, the digest remains consistent with each build.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -331,6 +331,8 @@ func getDelve(ctx context.Context, platform v1.Platform) (string, error) {
 	// install delve to tmp directory
 	args := []string{
 		"build",
+		"-trimpath",
+		"-ldflags=-s -w",
 		"-o",
 		delveBinaryPath,
 		"./cmd/dlv",


pkg/build/gobuild.go
```diff
@@ -331,6 +331,8 @@ func getDelve(ctx context.Context, platform v1.Platform) (string, error) {
 	// install delve to tmp directory
 	args := []string{
 		"build",
+		"-trimpath",
```
l-qing, line 334: - `trimpath`:
  - [Perfectly Reproducible, Verified Go Toolchains](https://go.dev/blog/rebuild)
    - For Go programs that don’t need cgo, a reproducible build is as simple as compiling with CGO_ENABLED=0 go build -trimpath. Disabling cgo removes the host C toolchain as a relevant input, and `-trimpath removes the current directory`.

- `"-ldflags=-s -w"`
  - [[Go: Reproducible Builds](https://crashonly.com/)]([Go: Reproducible Builds](https://crashonly.com/))
    - `-ldflags: reduce binary size by removing debug tables`
    - `-trimpath: remove paths from stacktraces`

imjasonh (approved):
Thank you!

PR #1464, opened 2024-12-12
fix(): use defaultFlags when flags is empty

### LINKED ISSUE #1450: defaultFlags is ignored
Hello! I'm trying to use `defaultFlags` but it looks like its ignored.

Here the reproducer:

```
mkdir ko-reproducer
cd ko-reproducer

cat <<EOF > main.go
package main

import "fmt"

func main() {
        fmt.Println("Hello World!")
}
EOF
go mod init example.com
go mod tidy
go build -cover .
GOCOVERDIR=hello ./example.com
```

This results in this error:

```
error: coverage meta-data emit failed: output directory "hello" inaccessible (err: stat hello: no such file or directory); no coverage data written
Hello World!
error: coverage counter data emit failed: output directory "hello" inaccessible (err: stat hello: no such file or directory); no coverage data written
```

Now with `ko`:

```
cat <<EOF > .ko.yaml
---
defaultFlags:
  - -cover
EOF

KOCACHE=kocache ko build . --push=false
GOCOVERDIR=hello ./kocache/bin/example.com/linux/amd64/out
```

Does not fail:

```
Hello World!
```

Now using "normal" flags:

```
cat <<EOF > .ko.yaml
---
builds:
- id: main
  main: ./
  flags:
    - -cover
EOF

rm -rf ./kocache
KOCACHE=kocache ko build . --push=false
GOCOVERDIR=hello ./kocache/bin/example.com/linux/amd64/out
```

```
error: coverage meta-data emit failed: output directory "hello" inaccessible (err: stat hello: no such file or directory); no coverage data written
Hello World!
error: coverage counter data emit failed: output directory "hello" inaccessible (err: stat hello: no such file or directory); no coverage data written
```

PD: I tested introducing a random string into `defaultFlags` and it does not fail, so its effectively ignored.
PD2: I can give it a go and try to fix it myself, but I want someone to confirm it first.

rh-hemartin:
This PR changes where defaultFlags get applied. Currently it is conditioned to `config.Flags` to be empty, but that is manipulated by a previous method. That manipulation includes adding flags based on the configuration options. This makes `config.Flags` almost never empty, so `defaultFlags` is almost never used. The PR changes where `defaultFlags` gets applied so it is previous to the manipulation based on command line arguments. This way if the build itself didn't specify flags, defaultFlags is used.

Resolves #1450

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -917,6 +917,11 @@ func createBuildArgs(ctx context.Context, buildCtx buildContext) ([]string, erro
 
 func (g *gobuild) configForImportPath(ip string) Config {
 	config := g.buildConfigs[ip]
+	if len(config.Flags) == 0 {
+		// Use the default, if any.
+		config.Flags = g.defaultFlags
+	}
+
 	if g.trimpath {
 		// The `-trimpath` flag removes file system paths from the resulting binary, to aid reproducibility.
 		// Ref: https://pkg.go.dev/cmd/go#hdr-Compile_packages_and_dependencies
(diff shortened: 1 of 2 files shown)


rh-hemartin:
Closing due to inactivity

PR #1500, opened 2025-04-10
fix: update viper to latest

caarlos0:
it has a few breaking changes, it seems that the only one that affects Ko is the mapstructure library change.

see https://github.com/spf13/viper/blob/master/UPGRADE.md#v120x

PS: doing this as it is preventing our upgrade in @goreleaser (ko is a dep :))


caarlos0:
sure

caarlos0:
@cpanato seems like this was already fixed on main, will close

caarlos0:
Nice! Thanks 🙏

PR #1590, opened 2025-12-13
release: disable SLSA generation

imjasonh:
The SLSA provenance steps in the release seem to be failing: https://github.com/ko-build/ko/actions/runs/20194679392/job/57977256910

These are tricky to test, since we only exercise them during releases, and we've had a history of mysteriously failing release workflows as a result. This latest release wasn't broken by the SLSA steps, but it's unclear whether users would be able to verify our releases.

I don't really have time to debug these, and it's unclear what value they provide to users. If there's a simpler way to achieve this (and, critically, test it without having to cut a release) I'm all ears.

For now I think we should just disable SLSA provenance generation for these releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,20 +3,16 @@ name: goreleaser
 on:
   push:
     tags:
-      - '*'
+      - "*"
 
 permissions: {}
 
 jobs:
   goreleaser:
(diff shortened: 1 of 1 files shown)


cpanato:
let me fix and make sure the next releases will work.

please hold this for a couple of days

PR #1619, opened 2026-03-26
build: prevent kodata symlinks from escaping the kodata root

evilgensec:
## Summary

`walkRecursive` dereferences symlinks in `kodata/` when packing files into container image layers. There was no check that the resolved path remained within the kodata root, so a symlink pointing at an arbitrary host path (e.g. `~/.ssh/id_rsa`, `/etc/passwd`, `/root/.ssh/authorized_keys`) would silently pack the target file into the published container image.

A module maintainer or CI pipeline operator who can place a symlink in `kodata/` — including via a dependency's `kodata/` directory or a compromised build step — could cause credentials and secrets from the build host to be embedded in the image pushed to a registry.

## Changes

**`pkg/build/gobuild.go`**

- `walkRecursive` gains an `absKodataRoot string` parameter: the canonical absolute path of the original kodata directory (resolved with both `filepath.EvalSymlinks` and `filepath.Abs`).
- After `filepath.EvalSymlinks(hostPath)`, the resolved path is checked with `strings.HasPrefix(absEvalPath, absKodataRoot+string(filepath.Separator))`. If the target is outside the kodata root the walk returns an error.
- The recursive call for directory symlinks passes the same `absKodataRoot` so the check is enforced at all nesting levels.
- `tarKoData` computes `absKodataRoot` using `filepath.EvalSymlinks` then `filepath.Abs` (order matters on macOS where `/var` → `/private/var`) and passes it to `walkRecursive`.

**`pkg/build/gobuild_test.go`**

- `TestWalkRecursiveSymlinkTraversal`: symlink escaping kodata root returns an error.
- `TestWalkRecursiveSymlinkWithinKodata`: symlink within kodata root is still followed correctly.

## Test

```
go test ./pkg/build/ -run TestWalkRecursiveSymlink -v
=== RUN   TestWalkRecursiveSymlinkTraversal
--- PASS
=== RUN   TestWalkRecursiveSymlinkWithinKodata
--- PASS
```

Related security report: https://issuetracker.google.com/issues/495623327

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -737,7 +737,10 @@ func writeFileToTar(tw *tar.Writer, name, evalPath string, size int64, modTime t
 // walkRecursive performs a filepath.Walk of the given root directory adding it
 // to the provided tar.Writer with root -> chroot.  All symlinks are dereferenced,
 // which is what leads to recursion when we encounter a directory symlink.
-func walkRecursive(tw *tar.Writer, root, chroot string, creationTime v1.Time, platform *v1.Platform) error {
+// absKodataRoot is the absolute path of the original kodata directory; symlinks
+// that resolve to a path outside of it are rejected to prevent arbitrary host
+// files from being packed into the container image.
+func walkRecursive(tw *tar.Writer, root, chroot, absKodataRoot string, creationTime v1.Time, platform *v1.Platform) error {
 	return filepath.Walk(root, func(hostPath string, info os.FileInfo, err error) error {
 		if hostPath == root {
 			return nil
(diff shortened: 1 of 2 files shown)


evilgensec:
Dear @imjasonh,

I received the following message while reporting a security issue. Is there any further way I can disclose the issue to the team responsibly?

Screenshot 2026-03-26 at 08 33 57

imjasonh (approved):

evankanderson:
FWIW, this seems likely to be breaking for a number of projects that symlink to a file such as LICENSE or other configuration that is _in the project repository_, but outside the command root.

Would a patch to define an "allowed parent directory" (and/or pick one out of git automatically) be accepted?

imjasonh:
Yeah I'd accept that. We can also revert the change if it's a problem.

evilgensec:
@evankanderson @imjasonh good point, and reverting would re-open the exfil hole (a symlink to `~/.ssh/id_rsa` or `/etc/passwd` getting packed into the pushed image). I'd rather fix the breakage than roll it back.

Proposal: widen the allowed boundary from the kodata root to the project's git repo root, so in-repo symlinks (LICENSE, shared config outside the command dir) keep working while out-of-repo targets stay blocked.

- Resolve the allowed parent as the git top-level (`git rev-parse --show-toplevel`) from the importpath dir; fall back to the kodata root when it isn't a git checkout.
- Keep the existing `EvalSymlinks` + `Abs` + `HasPrefix` check, just against the wider root.
- Add an override env (e.g. `KO_DATA_PATH_ALLOWED_ROOT`) for monorepos / non-git builds.

Tests: in-repo-but-outside-kodata symlink is allowed; outside-repo symlink still errors.

The git checkout is already trusted build context, so treating the repo root as the trust boundary keeps the security property while removing the false positive. Happy to send a PR if that direction works for you.

evankanderson:
I'm not going to fight you to get a PR in; I will probably have time to do it Wednesday or so if I haven't seen a PR by then. 

The spec you laid out looks correct for the cases I've thought of.

evilgensec:
Opened #1699 with the git-root allowed-parent approach we discussed (env override + git worktree fallback, ancestor-only so it never narrows below kodata). @evankanderson that should cover the in-repo LICENSE/config symlink case without reverting the escape check. Thanks for the review.

PR #1636, opened 2026-04-22
pkg/build: also block -overlay and -modfile from GOFLAGS and args

### LINKED ISSUE #1620: Security: GOFLAGS env bypasses toolexec check in gobuild.go
## Summary

The toolexec security check at `pkg/build/gobuild.go:925-931` only inspects the `flags`/`ldflags` args slice. It does not inspect the build environment. The `.ko.yaml` `env` field passes environment variables directly to `go build` via `buildEnv()` at lines 570-571 with no filtering.

Setting `GOFLAGS=-toolexec=/path/to/script` in the `env` field bypasses the check because Go's toolchain reads `GOFLAGS` from the environment.

Additionally, `-overlay` and `-modfile` are not blocked in either path (args or env), enabling source file substitution during build.

## Reproduce

`.ko.yaml`:
```yaml
builds:
  - env:
      - GOFLAGS=-toolexec=/tmp/test.sh
```

Where `/tmp/test.sh` is `#!/bin/sh\nexec "$@"`. The script executes on every compiler/linker invocation during `ko build`.

## Impact

A PR modifying `.ko.yaml` to add `GOFLAGS=-toolexec=...` in the env field can achieve code execution on CI runners, bypassing the existing toolexec mitigation.

## Suggested fix

Filter `GOFLAGS` from the build environment in `buildEnv()`, or extend the args check to also inspect env entries for `toolexec`, `overlay`, and `modfile`.

cc @imjasonh @jonjohnsonjr -- I could not find a private disclosure channel for this project. Happy to discuss privately if preferred.

SAY-5:
## Summary

Fixes #1620.

The existing guard in `buildEnv` rejected `-toolexec` via `GOFLAGS`, and the args-level check rejected `-toolexec` passed as a build flag. `-overlay` and `-modfile` have the same supply-chain reach:

- `-overlay=/path/map.json` replaces arbitrary source files on disk during the build with an attacker-controlled mapping.
- `-modfile=/path/other.mod` substitutes the go.mod that drives module resolution, silently redirecting the dependency graph.

Both are accepted by `go build` through either the args or `GOFLAGS`, so an attacker who can influence a `.ko.yaml` entry (via a PR, for example) can reproduce the same supply-chain effect the `-toolexec` blocklist was added to stop.

## Fix

Extract the three-flag check into `isBlockedGoBuildFlag` and run it over every `GOFLAGS` token in `buildEnv` and every element of `args` in the flag assembler. The helper handles both `-flag` / `--flag` single-token forms and `-flag=value` / `--flag=value` forms.

## Tests

Extended `TestBuildEnvRejectsToolexecInGOFLAGS` with six new cases:

- `-overlay`, `--overlay=` via `GOFLAGS` → rejected
- `-modfile` via `GOFLAGS` → rejected
- `-ldflags=-X=pkg/v.overlayID=abc` (string "overlay" as substring of a ldflag value) → allowed

Full `go test ./pkg/build/...` green (102s); `go vet ./pkg/build/...` clean.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -570,16 +570,20 @@ func buildEnv(platform v1.Platform, osEnv, buildEnv []string) ([]string, error) 
 	env = append(env, osEnv...)
 	env = append(env, buildEnv...)
 
-	// Reject GOFLAGS containing -toolexec to prevent bypassing the
-	// toolexec blocklist via environment variables.
+	// Reject GOFLAGS carrying flags that can redirect the build into
+	// attacker-controlled code: -toolexec swaps the compiler/linker
+	// wrapper, -overlay replaces arbitrary source files on disk with an
+	// alternate mapping, and -modfile substitutes the go.mod that drives
+	// module resolution. All three bypass the args-level blocklist when
+	// passed via the environment.
 	for _, e := range env {
 		if strings.HasPrefix(e, "GOFLAGS=") {
 			val := strings.TrimPrefix(e, "GOFLAGS=")
-			// Parse GOFLAGS into individual arguments and only match on flag tokens,
-			// so that values containing "-toolexec" as a substring are allowed.
+			// Parse GOFLAGS into individual arguments and only match on
+			// flag tokens, so that values containing one of these
+			// substrings as text (e.g. a path) are allowed.
 			for _, arg := range strings.Fields(val) {
-				if arg == "-toolexec" || arg == "--toolexec" ||
-					strings.HasPrefix(arg, "-toolexec=") || strings.HasPrefix(arg, "--toolexec=") {
+				if isBlockedGoBuildFlag(arg) {
 					return nil, fmt.Errorf("cannot set %s via GOFLAGS environment variable", arg)
 				}
 			}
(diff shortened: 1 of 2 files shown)


imjasonh (changes requested):
I'm not sure I believe that -modfile and -overlay are necessarily a risky option, the way that -toolexec is. I can imagine some source situation in which a person would want to `go build -modfile ...` etc, which I would want to make sure `ko build` also supports.

I agree it's a bit weird, and could theoretically be implicated in a supply chain attack, but I think I see the value a bit more than I see with `-toolexec`.

SAY-5:
Fair point. `-modfile` and `-overlay` both have legitimate build-time uses (vendored builds pointing at a different go.mod, upstream-source patching via overlay maps), and blocking them by default would be a behaviour regression for `ko build` users who rely on either.

I'll close this — the existing `-toolexec` guard already covers the canonical RCE primitive, and the helper/refactor bits alone aren't worth the diff without the extra blocks. Happy to revisit as an opt-in `--deny-unsafe-flags` style flag if that ever becomes desirable; let me know if that would be welcome and I can put up a smaller PR.

PR #1694, opened 2026-06-29
fix: skip yaml encoding when selector filters all documents

### LINKED ISSUE #1693: ko resolve fails with "expected STREAM-START" when label selector filters all documents in a file
## Bug

  `ko resolve` with a label selector (e.g. `-l app.kubernetes.io/component!=resolvers`)
  fails with:

      Error: error processing import paths in "config/resolvers/200-clusterrole.yaml":
      failed to close yaml encoder: yaml: expected STREAM-START

  when a YAML file has ALL its documents filtered out by the selector.

  ## Root cause

  In `resolveFile`, when the selector excludes every document, `docNodes`
  is empty and the encode loop is never entered. `e.Close()` is then called
  on an encoder that never wrote anything, which causes
  `go.yaml.in/yaml/v4`'s `Encoder.Close()` to return
  `"yaml: expected STREAM-START"` via the emitter.

  This was silently ignored in v0.18.x (`e.Close()` error was discarded).
  v0.19.0 started checking and propagating the error, surfacing the bug.

khrm:
When a label selector filters out all documents in a YAML file, docNodes is empty and the encode loop is never entered. Calling Close() on an encoder that never wrote anything causes go.yaml.in/yaml/v4 to return "yaml: expected STREAM-START", which was silently ignored in v0.18.x (error was discarded) but is now propagated as a fatal error.

Return early with empty bytes when docNodes is empty to avoid invoking the encoder at all.

Fixes: `failed to close yaml encoder: yaml: expected STREAM-START` when using ko resolve with a label selector that excludes all resources in a file.

Fixes #1693

### DIFF
--- a/pkg/commands/resolver.go
+++ b/pkg/commands/resolver.go
@@ -470,6 +470,13 @@ func resolveFile(
 		return nil, fmt.Errorf("error resolving image references: %w", err)
 	}
 
+	// If the selector filtered out all documents, skip encoding entirely.
+	// Calling Close() on an encoder that never wrote anything causes
+	// go.yaml.in/yaml/v4 to return "yaml: expected STREAM-START".
+	if len(docNodes) == 0 {
+		return []byte{}, nil
+	}
+
 	buf := &bytes.Buffer{}
 	e := yaml.NewEncoder(buf)
 	e.SetIndent(2)


khrm (commented):
/assign @imjasonh

khrm:
@imjasonh Can you look into this? Tekton CI is broken due to this.

imjasonh (approved):

PR #1699, opened 2026-07-06
build: allow kodata symlinks within the source tree

evilgensec:
## Summary

Follow-up to #1619. That change rejected any `kodata/` symlink whose target resolved outside the kodata directory. As @evankanderson noted on #1619, this also breaks the legitimate and common case of symlinking a file that lives inside the project repository but outside the command root (a top-level `LICENSE`, shared config, etc.).

This PR keeps the security property (a symlink to `~/.ssh/id_rsa` or `/etc/passwd` still cannot be packed into the image) while removing the false positive, by widening the allowed boundary from the kodata root to the enclosing source tree.

## Changes

**`pkg/build/gobuild.go`**

- `walkRecursive`'s boundary parameter is renamed `absKodataRoot` -> `absAllowedRoot`; the `EvalSymlinks` + `Abs` + `HasPrefix` containment check now runs against this wider root. The recursion for directory symlinks passes it through unchanged, so the check holds at every nesting level.
- New `resolveKodataAllowedRoot` computes the allowed root:
  - `KO_DATA_PATH_ALLOWED_ROOT` when set (for monorepos / non-git builds),
  - otherwise the enclosing git worktree via `git rev-parse --show-toplevel`,
  - falling back to the kodata root when neither is available.
  - The candidate is only used when it is an ancestor of (or equal to) the kodata root, so the boundary can never be narrowed below kodata and normal kodata contents are never affected.
- `tarKoData` computes `absAllowedRoot` and passes it to `walkRecursive`.

**`pkg/build/gobuild_test.go`**

- `TestWalkRecursiveSymlinkWithinAllowedRoot`: a symlink in kodata pointing to an in-repo, out-of-kodata `LICENSE` is followed and packed (the case from #1619).
- `TestResolveKodataAllowedRoot`: ancestor override widens the boundary; a non-ancestor override and a non-git tree both fall back to the kodata root.
- Existing `TestWalkRecursiveSymlinkTraversal` / `TestWalkRecursiveSymlinkWithinKodata` still pass (error message updated to "outside the allowed root").

## Test

```
go test ./pkg/build/ -run 'TestWalkRecursiveSymlink|TestResolveKodataAllowedRoot' -v
=== RUN   TestWalkRecursiveSymlinkTraversal
--- PASS
=== RUN   TestWalkRecursiveSymlinkWithinKodata
--- PASS
=== RUN   TestWalkRecursiveSymlinkWithinAllowedRoot
--- PASS
=== RUN   TestResolveKodataAllowedRoot
--- PASS
```

Full `go test ./pkg/build/` passes.

Closes the compatibility regression raised in #1619 without reverting the symlink-escape fix.

### DIFF
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -874,14 +874,13 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
 	// If the kodata directory does not exist, walkRecursive handles that
 	// gracefully (filepath.Walk skips missing roots), so we fall back to the
 	// raw path as the boundary — no traversal is possible without a root.
-	absKodataRoot := root
 	absAllowedRoot := root
 	if _, statErr := os.Stat(root); statErr == nil {
 		resolvedRoot, err := filepath.EvalSymlinks(root)
 		if err != nil {
 			return nil, fmt.Errorf("filepath.EvalSymlinks(%q): %w", root, err)
 		}
-		absKodataRoot, err = filepath.Abs(resolvedRoot)
+		absKodataRoot, err := filepath.Abs(resolvedRoot)
 		if err != nil {
 			return nil, fmt.Errorf("filepath.Abs(%q): %w", resolvedRoot, err)
 		}
@@ -925,18 +924,15 @@ func resolveKodataAllowedRoot(absKodataRoot string) string {
 }
 
 // withinRoot reports whether p is root itself or lies inside the root
-// directory.  It tolerates a root that already ends in a path separator (a
-// filesystem root such as "/" or a Windows drive root like `C:\`), which would
-// otherwise produce a double-separator prefix and spuriously fail the check.
+// directory.  Using filepath.Rel keeps the comparison robust against trailing
+// separators (a filesystem root such as "/" or a Windows drive root like
+// `C:\`) and platform-specific separators.
 func withinRoot(p, root string) bool {
-	if p == root {
-		return true
+	rel, err := filepath.Rel(root, p)
+	if err != nil {
+		return false
 	}
-	sep := string(filepath.Separator)
-	if !strings.HasSuffix(root, sep) {
-		root += sep
-	}
-	return strings.HasPrefix(p, root)
+	return rel != ".." && !strings.HasPrefix(rel, ".."+string(filepath.Separator))
 }
 
 // kodataAllowedRootCandidate returns the caller-supplied or git-derived
@@ -945,11 +941,23 @@ func kodataAllowedRootCandidate(absKodataRoot string) string {
 	if env := os.Getenv("KO_DATA_PATH_ALLOWED_ROOT"); env != "" {
 		return env
 	}
-	out, err := exec.Command("git", "-C", absKodataRoot, "rev-parse", "--show-toplevel").Output()
-	if err != nil {
-		return ""
+	return enclosingGitRoot(absKodataRoot)
+}
+
+// enclosingGitRoot walks up from dir looking for a .git entry and returns the
+// worktree root, or "" if dir is not inside a git checkout.  It reads the
+// filesystem directly rather than shelling out to git.
+func enclosingGitRoot(dir string) string {
+	for {
+		if _, err := os.Stat(filepath.Join(dir, ".git")); err == nil {
+			return dir
+		}
+		parent := filepath.Dir(dir)
+		if parent == dir {
+			return ""
+		}
+		dir = parent
 	}
-	return strings.TrimSpace(string(out))
 }
 
 func createTemplateData(ctx context.Context, buildCtx buildContext) (map[string]any, error) {
--- a/pkg/build/gobuild.go
+++ b/pkg/build/gobuild.go
@@ -763,10 +763,12 @@ func writeFileToTar(tw *tar.Writer, name, evalPath string, size int64, modTime t
 // walkRecursive performs a filepath.Walk of the given root directory adding it
 // to the provided tar.Writer with root -> chroot.  All symlinks are dereferenced,
 // which is what leads to recursion when we encounter a directory symlink.
-// absKodataRoot is the absolute path of the original kodata directory; symlinks
-// that resolve to a path outside of it are rejected to prevent arbitrary host
-// files from being packed into the container image.
-func walkRecursive(tw *tar.Writer, root, chroot, absKodataRoot string, creationTime v1.Time, platform *v1.Platform) error {
+// absAllowedRoot is the absolute path of the directory tree within which
+// symlinks are permitted to resolve; symlinks that resolve to a path outside of
+// it are rejected to prevent arbitrary host files from being packed into the
+// container image.  It defaults to the kodata root but may be widened to the
+// enclosing source tree (see resolveKodataAllowedRoot).
+func walkRecursive(tw *tar.Writer, root, chroot, absAllowedRoot string, creationTime v1.Time, platform *v1.Platform) error {
 	return filepath.Walk(root, func(hostPath string, info os.FileInfo, err error) error {
 		if hostPath == root {
 			return nil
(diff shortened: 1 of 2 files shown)


pkg/build/gobuild.go
```diff
@@ -798,17 +800,17 @@ func walkRecursive(tw *tar.Writer, root, chroot, absKodataRoot string, creationT
...
 		}
-		absKodataRootWithSep := absKodataRoot + string(filepath.Separator)
-		if absEvalPath != absKodataRoot && !strings.HasPrefix(absEvalPath, absKodataRootWithSep) {
-			return fmt.Errorf("kodata symlink %q resolves to %q which is outside the kodata root %q",
-				hostPath, evalPath, absKodataRoot)
+		absAllowedRootWithSep := absAllowedRoot + string(filepath.Separator)
+		if absEvalPath != absAllowedRoot && !strings.HasPrefix(absEvalPath, absAllowedRootWithSep) {
+			return fmt.Errorf("kodata symlink %q resolves to %q which is outside the allowed root %q",
+				hostPath, evalPath, absAllowedRoot)
 		}
```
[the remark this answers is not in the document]
evilgensec: Good catch. Fixed in 40c8744: extracted a `withinRoot` helper that only appends the separator when the root does not already end in one, used in both `walkRecursive` and `resolveKodataAllowedRoot`, and added `TestWithinRoot` covering the trailing-separator root case.

pkg/build/gobuild.go
```diff
@@ -883,8 +886,56 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
...
+		return absKodataRoot
+	}
+	// Canonicalize so the comparison matches absKodataRoot, which was resolved
+	// with EvalSymlinks (e.g. /var -> /private/var on macOS).
+	if resolved, err := filepath.EvalSymlinks(abs); err == nil {
+		abs = resolved
+	}
+	if abs == absKodataRoot || strings.HasPrefix(absKodataRoot, abs+string(filepath.Separator)) {
+		return abs
+	}
```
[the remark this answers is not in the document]
evilgensec: Good catch. Fixed in 40c8744: extracted a `withinRoot` helper that only appends the separator when the root does not already end in one, used in both `walkRecursive` and `resolveKodataAllowedRoot`, and added `TestWithinRoot` covering the trailing-separator root case.

imjasonh:
I had Copilot throw https://github.com/ko-build/ko/pull/1698/ together -- I haven't reviewed it yet myself but I will probably later today.

pkg/build/gobuild.go
```diff
@@ -883,8 +885,71 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
...
+func withinRoot(p, root string) bool {
+	if p == root {
+		return true
+	}
+	sep := string(filepath.Separator)
+	if !strings.HasSuffix(root, sep) {
+		root += sep
+	}
+	return strings.HasPrefix(p, root)
+}
```
evankanderson, line 936: I'd feel slightly better using `filepath.Rel` here:

```suggestion
func withinRoot(p, root string) bool {
    relPath, err := filepath.Rel(root, p)
    if err != nil || strings.HasPrefix(relPath, "../") {
    	return false
	}
	return true
}
```

pkg/build/gobuild.go
```diff
@@ -883,8 +885,71 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
...
+
+// kodataAllowedRootCandidate returns the caller-supplied or git-derived
+// candidate for the widened symlink boundary, or "" when none is available.
+func kodataAllowedRootCandidate(absKodataRoot string) string {
+	if env := os.Getenv("KO_DATA_PATH_ALLOWED_ROOT"); env != "" {
+		return env
+	}
+	out, err := exec.Command("git", "-C", absKodataRoot, "rev-parse", "--show-toplevel").Output()
```
evankanderson: I'm not sure how we feel about shelling out to `git`.

pkg/build/gobuild.go
```diff
@@ -883,8 +885,71 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
...
+		return absKodataRoot
+	}
+	// Canonicalize so the comparison matches absKodataRoot, which was resolved
+	// with EvalSymlinks (e.g. /var -> /private/var on macOS).
+	if resolved, err := filepath.EvalSymlinks(abs); err == nil {
+		abs = resolved
+	}
+	if withinRoot(absKodataRoot, abs) {
+		return abs
+	}
```
evankanderson, line 922: I'm assuming this is just a sanity check?

pkg/build/gobuild_test.go
```diff
@@ -1821,3 +1821,139 @@ func TestWalkRecursiveSymlinkWithinKodata(t *testing.T) {
...
+// root (e.g. a top-level LICENSE in the project) is followed correctly.
+func TestWalkRecursiveSymlinkWithinAllowedRoot(t *testing.T) {
+	// repoDir is the widened allowed root; kodata lives beneath it.
+	repoDir := t.TempDir()
+	kodataDir := filepath.Join(repoDir, "cmd", "app", "kodata")
+	if err := os.MkdirAll(kodataDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	// A file inside the repo but outside kodata (the case evankanderson raised).
```
evankanderson, line 1836: I believe Tekton has this case, and Knative did for a while.

Minder has a similar case where it packages a default config file that's near the directory root.

pkg/build/gobuild.go
```diff
@@ -874,6 +875,7 @@ func (g *gobuild) tarKoData(ref reference, platform *v1.Platform) (*bytes.Buffer
 	// gracefully (filepath.Walk skips missing roots), so we fall back to the
 	// raw path as the boundary — no traversal is possible without a root.
 	absKodataRoot := root
```
evankanderson: You have a lint failure here.

evankanderson (commented):
A couple comments, but for me it's faster to review a solution than to decide which solution to suggest, so thanks for the PR!

pkg/build/gobuild.go
evilgensec: Done in d455fba — switched to `filepath.Rel`.

pkg/build/gobuild.go
evilgensec: Fair. Dropped the shell-out in d455fba — `enclosingGitRoot` just walks up for a `.git` entry, no subprocess.

pkg/build/gobuild.go
evilgensec: Yep, sanity check — it confirms the candidate is actually an ancestor of the kodata dir, so a stray `KO_DATA_PATH_ALLOWED_ROOT` (or a git root that somehow doesn't contain kodata) can only widen the boundary, never narrow it below kodata.

pkg/build/gobuild_test.go
evilgensec: Thanks, good to know — that's the exact shape TestWalkRecursiveSymlinkWithinAllowedRoot covers (an in-repo file above the command root).

pkg/build/gobuild.go
evilgensec: Fixed in d455fba — it was an ineffectual assignment (absKodataRoot gets overwritten before use), cleaned up.

evilgensec:
Thanks for the review @evankanderson! Pushed d455fba addressing all of it:

- `withinRoot` now uses `filepath.Rel`
- replaced the `git` shell-out with a filesystem `.git` walk-up (`enclosingGitRoot`), no subprocess
- fixed the ineffassign lint failure
- added a test for the git-worktree detection

Let me know if you'd prefer to drop the git auto-detection entirely and lean on `KO_DATA_PATH_ALLOWED_ROOT` only.

imjasonh (approved):

PR #1732, opened 2026-08-17
Fix kind.Write goroutine leak when image import fails

SebTardif:
## What

`kind.Write` streams an image tarball into `ctr images import` over an `io.Pipe`. If import fails, it returned without closing the read end or waiting on the writer goroutine. `tarball.Write` then blocked on a full pipe until process exit.

This is reachable with `KO_DOCKER_REPO=kind.local` when a node rejects the import (missing ctr, cancelled context, bad image). Library callers such as Skaffold can leak one blocked goroutine per failed load.

## Fix

Close the pipe after `cmd.Run()` and always `Wait` on the errgroup so the writer is unblocked on both success and failure.

## Test

`TestWriteClosesPipeWhenImportFailsWithoutReadingStdin` uses a large image and a fake node that fails import without reading stdin. Before the fix, a follow-up Read on the pipe either returned data or blocked. After the fix, the pipe is closed.

`go test ./pkg/publish/kind/ -count=1`

## Origin

Introduced in #180 (2020-09-04), present for 2173 days.

---

> [!NOTE]
> **Low Risk**
> Localized change to kind image publish error handling and tests; no auth, data, or schema assumptions.
> 
> **Overview**
> Fixes a **goroutine leak** in `kind.Write` when `ctr images import` fails or stops reading stdin before the tarball is fully streamed.
> 
> After `cmd.Run()`, the read end of the `io.Pipe` is **closed** and the errgroup is **always waited on**, so the background `tarball.Write` goroutine is unblocked on both success and failure instead of blocking on a full pipe until process exit.
> 
> Adds **`TestWriteClosesPipeWhenImportFailsWithoutReadingStdin`** (large image + fake node that fails without consuming stdin) and extends test fakes with **`skipStdin`** to simulate that failure mode.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1993ca40b3ca887e3e96ab97c2943fa60142a15b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pkg/publish/kind/write.go
+++ b/pkg/publish/kind/write.go
@@ -73,7 +73,7 @@ func Write(ctx context.Context, tag name.Tag, img v1.Image) error {
 		cmd.SetStdout(&buf)
 		cmd.SetStderr(&buf)
 		err := cmd.Run()
-		// Unblock tarball.Write if the importer stopped reading (failed or cancelled).
+		// Unblock tarball.Write if the importer stopped reading (failed or canceled).
 		pr.Close()
 		waitErr := grp.Wait()
 		if err != nil {
--- a/pkg/publish/kind/write.go
+++ b/pkg/publish/kind/write.go
@@ -65,7 +65,10 @@ func Write(ctx context.Context, tag name.Tag, img v1.Image) error {
 
 		grp := errgroup.Group{}
 		grp.Go(func() error {
-			return pw.CloseWithError(tarball.Write(tag, img, pw))
+			// CloseWithError always returns nil; return the Write error so Wait surfaces it.
+			err := tarball.Write(tag, img, pw)
+			pw.CloseWithError(err)
+			return err
 		})
 
 		var buf bytes.Buffer
--- a/pkg/publish/kind/write.go
+++ b/pkg/publish/kind/write.go
@@ -72,12 +72,15 @@ func Write(ctx context.Context, tag name.Tag, img v1.Image) error {
 		cmd := n.CommandContext(ctx, "ctr", "--namespace=k8s.io", "images", "import", "--all-platforms", "-").SetStdin(pr)
 		cmd.SetStdout(&buf)
 		cmd.SetStderr(&buf)
-		if err := cmd.Run(); err != nil {
+		err := cmd.Run()
+		// Unblock tarball.Write if the importer stopped reading (failed or cancelled).
+		pr.Close()
+		waitErr := grp.Wait()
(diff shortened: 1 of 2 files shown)


pkg/publish/kind/write.go
```diff
@@ -72,12 +72,15 @@ func Write(ctx context.Context, tag name.Tag, img v1.Image) error {
...
+		// Unblock tarball.Write if the importer stopped reading (failed or cancelled).
+		pr.Close()
+		waitErr := grp.Wait()
+		if err != nil {
 			return fmt.Errorf("failed to load image to node %q: %w\n%s", n, err, buf.String())
 		}
-
-		if err := grp.Wait(); err != nil {
-			return fmt.Errorf("failed to write intermediate tarball representation: %w", err)
+		if waitErr != nil {
```
imjasonh, line 85:
```suggestion
		if err != nil {
return fmt.Errorf("failed to load image to node %q: %w\n%s", n, err, buf.String())
}
		if waitErr := grp.Wait(); waitErr != nil {
```

SebTardif: @imjasonh

> if err != nil { return ... }
> if waitErr := grp.Wait(); waitErr != nil {

We still want `grp.Wait()` on the import-failure path. `pr.Close()` unblocks `tarball.Write`, and `Wait` then reaps that goroutine instead of leaving it to exit in the background. After Close, Wait should return quickly.

The `waitErr := grp.Wait()` before the `if err != nil` is intentional for that reason. Happy to restructure if you would rather it read differently.

imjasonh: Thanks for that explanation, I'm convinced.

pkg/publish/kind/write.go
```diff
@@ -72,12 +72,15 @@ func Write(ctx context.Context, tag name.Tag, img v1.Image) error {
...
+			return fmt.Errorf("failed to write intermediate tarball representation: %w", waitErr)
```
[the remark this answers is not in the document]
SebTardif: @Copilot

> waitErr := grp.Wait() is expected to surface tarball.Write errors, but the writer goroutine currently returns the result of pw.CloseWithError(...)

Good catch. CloseWithError always returns nil, so Wait never saw a Write failure. The writer now returns the tarball.Write error after CloseWithError, and TestWriteSurfacesTarballWriteError covers it.

Issue #562, opened 2022-01-04
failed to save tarfile but output code is success


Dentrax:
```
$ KO_DOCKER_REPO=registry/bar ko resolve -B --bare --tags latest -f cronjob_template.yaml --tarball ./foo.tar --push=false

2022/01/04 17:23:52 Using base gcr.io/distroless/static:nonroot for foo/bar
2022/01/04 17:23:54 Building foo/bar for linux/amd64
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: test
spec:
  jobTemplate:
    metadata:
      name: test
    spec:
      template:
        spec:
          imagePullSecrets:
            - name: docker-registry
          containers:
            - image: registry/bar@sha256:d9542b623e20a20ddf87d452f5f6dffaa22167c8a77449c232805590a03e6645
              name: test
              imagePullPolicy: Always
          restartPolicy: OnFailure
  schedule: ${SCHEDULE}

---
2022/01/04 17:23:58 Saving ./foo.tar
2022/01/04 17:23:58 failed to save "./foo.tar": Get "https://gcr.io/v2/distroless/static/blobs/sha256:2df365faf0e3007f983fadd7a65ba51d41b488eb2ed8fc70f4bf97043cfea560": context canceled
```

It says `failed to save "./foo.tar": ...: context canceled` but actually overall exit code is `0` and it writes to file successfully:

```
$ echo $?
0

$ stat ./foo.tar
File: foo.tar
...
```

```
$ ko version
0.9.3
```

Any ideas about why it says `failed to save`? 🤔

jonjohnsonjr:
Yeah tarball was broken at some point, fixed in https://github.com/google/ko/pull/545 which hasn't been released. Can you try at HEAD to see if it's still broken?

PR #1734, opened 2026-08-17
Propagate publisher Close and resolved YAML write errors

SebTardif:
## What

Tarball persist happens only in `publisher.Close()`. Every command used `defer publisher.Close()`, so a failed save was logged and then discarded. `ko build --tarball=out.tar` could print a digest and exit 0.

The same `defer` hid recorder file-close errors. `ResolveFilesToWriter` ignored `Write` and `Close` errors, so a broken pipe looked like success.

This matches the user report in #562 (closed stale).

## Fix

- Named-return `defer` calls `closePublisher`, which always Closes and returns the Close error when the command otherwise succeeded.
- Recorder Closes the refs file even if the inner publisher Close fails (`errors.Join`).
- `ResolveFilesToWriter` checks Write and Close. Future channels are buffered so a Write failure cannot deadlock in-flight resolvers.

## Test

- `TestClosePublisher` (prefer work error; surface Close error on success)
- `TestRecorderCloseClosesFileWhenInnerFails`
- `TestResolveFilesToWriterReturnsWriteError`
- `TestResolveFilesToWriterReturnsWriteErrorWithMultipleFiles` (would hang without the buffered channel)

`go test ./pkg/commands/ ./pkg/publish/ -count=1 -run 'TestClosePublisher|TestRecorderClose|TestResolveFilesToWriterReturnsWriteError'`

## Origin

Tarball Close plus the "bad practice" comment landed in #134 (2020-02-19), 2371 days ago. YAML Write has discarded errors since the initial commit (2019-03-14). #540 tried to remove `Close` instead and was closed.

Ref #562

### DIFF
--- a/pkg/commands/build.go
+++ b/pkg/commands/build.go
@@ -55,7 +55,7 @@ func addBuild(topLevel *cobra.Command) {
   #   ko.local/<import path>
   # This always preserves import paths.
   ko build --local github.com/foo/bar/cmd/baz github.com/foo/bar/cmd/blah`,
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			if err := options.Validate(po, bo); err != nil {
 				return fmt.Errorf("validating options: %w", err)
 			}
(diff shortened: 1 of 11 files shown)


pkg/commands/build.go
```diff
@@ -76,7 +76,9 @@ func addBuild(topLevel *cobra.Command) {
 			if err != nil {
 				return fmt.Errorf("error creating publisher: %w", err)
 			}
-			defer publisher.Close()
+			defer func() {
+				err = closePublisher(publisher, err)
+			}()
```
imjasonh, line 81: I'm not sure this is the most readable way to achieve this.

It seems like a pretty rare occurrence for publisher.Close to fail, so I'm not super concerned about handling it if it's going to cause all the callsites to grow this wart.

PR #269, opened 2020-12-16
Introduce KOCACHE

jonjohnsonjr:
Start of https://github.com/google/ko/issues/264

Currently, use this by setting `KOCACHE=/tmp/ko` (or any directory you want). We can default to doing this in the future.

We will cache binaries and metdata under `$KOCACHE/bin/<import path>/<platform>`:

```
$ tree /tmp/ko
/tmp/ko                                                                                                                                                                                                                                                                                                                        
`-- bin                                                                                                                                                                                                                                                                                                                        
    `-- github.com                                                                                                                                                                                                                                                                                                             
        `-- google                                                                                                                                                                                                                                                                                                             
            `-- ko                                                                                                                                                                                                                                                                                                             
                `-- linux                                                                                                                                                                                                                                                                                                      
                    |-- amd64                                                                                                                                                                                                                                                                                                  
                    |   |-- buildid-to-diffid                                                                                                                                                                                                                                                                                  
                    |   |-- diffid-to-descriptor                                                                                                                                                                                                                                                                               
                    |   `-- out                                                                                                                                                                                                                                                                                                
                    |-- arm                                                                                                                                                                                                                                                                                                    
                    |   |-- buildid-to-diffid                                                                                                                                                                                                                                                                                  
                    |   |-- diffid-to-descriptor                                                                                                                                                                                                                                                                               
                    |   `-- out                                                                                                                                                                                                                                                                                                
                    |-- arm64                                                                                                                                                                                                                                                                                                  
                    |   |-- buildid-to-diffid                                                                                                                                                                                                                                                                                  
                    |   |-- diffid-to-descriptor                                                                                                                                                                                                                                                                               
                    |   `-- out                                                                                                                                                                                                                                                                                                
                    |-- ppc64le                                                                                                                                                                                                                                                                                                
                    |   |-- buildid-to-diffid                                                                                                                                                                                                                                                                                  
                    |   |-- diffid-to-descriptor                                                                                                                                                                                                                                                                               
                    |   `-- out                                                                                                                                                                                                                                                                                                
                    `-- s390x                                                                                                                                                                                                                                                                                                  
                        |-- buildid-to-diffid                                                                                                                                                                                                                                                                                  
                        |-- diffid-to-descriptor                                                                                                                                                                                                                                                                               
                        `-- out   
```

In `buildid-to-diffid`, we maintain a map from the binary's buildid to the diffid we produced from it:

```
[...]

PR #1735, opened 2026-08-17
Preserve git and digest errors instead of dropping them

SebTardif:
## What

Three small error-handling bugs:

1. `git run()` replaced the exec error with `errors.New(stderr)`. A cancelled context or missing stderr became an empty error. `GetInfo` then treated cancel as "not a git repository".
2. `go tool buildid` logged `err` in the filename slot, so the message showed `go tool buildid exit status 1` instead of the path.
3. The multi-arch path discarded `baseIndex.Digest()` and wrote a zero hash into `org.opencontainers.image.base.digest`. The single-platform path already returns that error.

## Fix

Wrap git failures as `fmt.Errorf("%w: %s", err, stderr)` (or return `err` when stderr is empty). Propagate `context.Canceled` / `DeadlineExceeded` from `isRepo` and `checkDirty`. Swap the buildid log arguments. Return the Digest() error on the multi-arch path.

## Test

- `TestGetInfoCanceledContext` (`errors.Is(err, context.Canceled)`)
- `TestGetBuildIDLogIncludesFilename` (log contains the path, not `go tool buildid exit status`)
- Existing `TestNotAGitFolder` still expects `ErrNotRepository`
- Existing `TestGoBuildIndex` still passes

`go test ./pkg/internal/git/ ./pkg/build/ -count=1`

## Origin

Git wrapper: #1307 (2024-05-11). Buildid log: #269 (2021-12-08). Index digest discard: #1426 (2024-10-15).

### DIFF
--- a/pkg/build/cache.go
+++ b/pkg/build/cache.go
@@ -208,7 +208,7 @@ func getBuildID(ctx context.Context, file string) (string, error) {
 	cmd.Stdout = &output
 
 	if err := cmd.Run(); err != nil {
-		log.Printf("Unexpected error running \"go tool buildid %s\": %v\n%v", err, file, output.String())
+		log.Printf("Unexpected error running \"go tool buildid %s\": %v\n%v", file, err, output.String())
 		return "", fmt.Errorf("go tool buildid %s: %w", file, err)
 	}
 	return strings.TrimSpace(output.String()), nil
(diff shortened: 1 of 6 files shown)


imjasonh (changes requested):
I'm not sure this change is worth the effort. In practice, git init will very likely not be context-canceled, so returning errors in `isRepo` is overkill.

